PR #10, opened 2020-07-08
Header parsing tests

### LINKED ISSUE #5: Test and document HTTP request headers parsing by net/http
Using #4, add test scenarios showing how an `http.Handler` sees the incoming HTTP request headers.

Take into account:
- tricky encoding
- ordering of headers
- case sensitivity
- how different headers interact with each other according to the HTTP spec
- and many others.

Flag potentially problematic behavior.

Some reading material:
- The Tangled Web by Michał Zalewski (2011)
- https://portswigger.net/web-security/request-smuggling

mattiasgrenfeldt:
Fixes #5 

Includes many different tests.

### DIFF
--- a/tests/requestparsing/README.md
+++ b/tests/requestparsing/README.md
@@ -1,0 +1,6 @@
+
+# Request Parsing tests
+
+This package contains tests for verify the request parsing behavior of `net/http` in Go's standard library.
+
+Note: All tests use verbatim newline characters in their requests instead of using multiline strings to ensure that \r and \n end up in exactly the right places.
--- a/tests/requestparsing/README.md
+++ b/tests/requestparsing/README.md
@@ -1,4 +1,3 @@
-
 # Request Parsing tests
 
 This package contains tests for verify the request parsing behavior of `net/http` in Go's standard library.
--- a/tests/requestparsing/README.md
+++ b/tests/requestparsing/README.md
@@ -1,5 +1,0 @@
-# Request Parsing tests
-
-This package contains tests for verify the request parsing behavior of `net/http` in Go's standard library.
-
-Note: All tests use verbatim newline characters in their requests instead of using multiline strings to ensure that \r and \n end up in exactly the right places.
--- a/tests/requestparsing/basicauth_test.go
+++ b/tests/requestparsing/basicauth_test.go
@@ -101,6 +101,40 @@ func TestBasicAuth(t *testing.T) {
 				password: "aaa",
 			},
 		},
+		{
+			name: "CasingOrdering1",
+			request: []byte("GET / HTTP/1.1\r\n" +
+				"Host: localhost:8080\r\n" +
+				// Base64 encoding of "AAA:aaa".
+				"Authorization: basic QUFBOmFhYQ==\r\n" +
+				// Base64 encoding of "BBB:bbb".
+				"authorization: basic QkJCOmJiYg==\r\n" +
+				"\r\n"),
+			want: basicAuthWant{
+				// Base64 encoding of "AAA:aaa" and then of "BBB:bbb" in that order.
+				headers:  map[string][]string{"Authorization": []string{"basic QUFBOmFhYQ==", "basic QkJCOmJiYg=="}},
+				ok:       true,
+				username: "AAA",
+				password: "aaa",
+			},
+		},
+		{
+			name: "CasingOrdering2",
+			request: []byte("GET / HTTP/1.1\r\n" +
+				"Host: localhost:8080\r\n" +
+				// Base64 encoding of "AAA:aaa".
+				"authorization: basic QUFBOmFhYQ==\r\n" +
+				// Base64 encoding of "BBB:bbb".
+				"Authorization: basic QkJCOmJiYg==\r\n" +
+				"\r\n"),
+			want: basicAuthWant{
+				// Base64 encoding of "AAA:aaa" and then of "BBB:bbb" in that order.
+				headers:  map[string][]string{"Authorization": []string{"basic QUFBOmFhYQ==", "basic QkJCOmJiYg=="}},
+				ok:       true,
+				username: "AAA",
+				password: "aaa",
+			},
+		},
 	}
 
 	for _, tt := range basicAuthTests {
--- a/tests/requestparsing/basicauth_test.go
+++ b/tests/requestparsing/basicauth_test.go
@@ -28,23 +28,23 @@ func TestBasicAuth(t *testing.T) {
 			name: "Basic",
 			request: []byte("GET / HTTP/1.1\r\n" +
 				"Host: localhost:8080\r\n" +
-				// Base64 encoding of "Pelle:Password".
-				"Authorization: Basic UGVsbGU6UGFzc3dvcmQ=\r\n" +
+				// Base64 encoding of "Pizza:Password".
+				"Authorization: Basic UGl6emE6UGFzc3dvcmQ=\r\n" +
 				"\r\n"),
 			wantBasicAuth: basicAuth{
-				username: "Pelle",
+				username: "Pizza",
 				password: "Password",
 				ok:       true,
 			},
 			// Same Base64 as above.
-			wantHeaders: map[string][]string{"Authorization": []string{"Basic UGVsbGU6UGFzc3dvcmQ="}},
+			wantHeaders: map[string][]string{"Authorization": []string{"Basic UGl6emE6UGFzc3dvcmQ="}},
 		},
 		{
 			name: "NoTrailingEquals",
 			request: []byte("GET / HTTP/1.1\r\n" +
 				"Host: localhost:8080\r\n" +
-				// Base64 encoding of "Pelle:Password" without trailing equals.
-				"Authorization: Basic UGVsbGU6UGFzc3dvcmQ\r\n" +
+				// Base64 encoding of "Pizza:Password" without trailing equals.
+				"Authorization: Basic UGl6emE6UGFzc3dvcmQ\r\n" +
 				"\r\n"),
 			wantBasicAuth: basicAuth{
 				username: "",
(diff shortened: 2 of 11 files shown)


tests/requestparsing/basicauth_test.go
```diff
@@ -0,0 +1,117 @@
...
+func TestBasicAuth(t *testing.T) {
+	for _, tt := range basicAuthTests {
+		t.Run(tt.name, func(t *testing.T) {
+			resp, err := requesttesting.MakeRequest(context.Background(), tt.request, func(r *http.Request) {
+				if diff := cmp.Diff(tt.want.headers, map[string][]string(r.Header)); diff != "" {
+					t.Errorf("r.Header mismatch (-want +got):\n%s", diff)
+				}
+
+				username, password, ok := r.BasicAuth()
+				if ok != tt.want.ok || username != tt.want.username || password != tt.want.password {
```
kele, line 103: Split this into three separate checks.

tests/requestparsing/basicauth_test.go
```diff
@@ -0,0 +1,117 @@
...
+					t.Errorf("username, password, ok := r.BasicAuth() got: %q, %q, %v want: %q, %q, %v", username, password, ok, tt.want.username, tt.want.password, tt.want.ok)
+				}
+			})
+			if err != nil {
+				t.Errorf("MakeRequest() got: %v want: nil", err)
```
kele, line 108: got err

tests/requestparsing/basicauth_test.go
```diff
@@ -0,0 +1,117 @@
...
+var basicAuthTests = []struct {
+	name    string
+	request []byte
+	want    basicAuthWant
+}{
+	{
+		name: "Basic",
+		request: []byte("GET / HTTP/1.1\r\n" +
+			"Host: localhost:8080\r\n" +
+			"Authorization: Basic UGVsbGU6UGFzc3dvcmQ=\r\n" +
```
kele, line 30: Please add a comment saying what the base64 encodes. Here and everywhere else.

tests/requestparsing/basicauth_test.go
```diff
@@ -0,0 +1,117 @@
...
+			password: "",
+		},
+	},
+	{
+		name: "Ordering",
+		request: []byte("GET / HTTP/1.1\r\n" +
+			"Host: localhost:8080\r\n" +
+			"Authorization: basic QUFBOmFhYQ==\r\n" +
+			"Authorization: basic QkJCOmJiYg==\r\n" +
+			"\r\n"),
```
kele, line 84: Could you add tests checking whether the ordering is case sensitive? i.e. what if `authorization` is before `Authorization` and the other way around? You can imagine that they're sorted first and then applied ToLower().

I know you made such a test for custom headers, but I wonder if the ones that have some special handling behave differently.

tests/requestparsing/contentlengthandtransferencoding_test.go
```diff
@@ -0,0 +1,446 @@
...
+			t.Errorf("r.Header mismatch (-want +got):\n%s", diff)
+		}
+
+		if r.ContentLength != 5 {
+			t.Errorf("r.ContentLength got: %v want: 5", r.ContentLength)
+		}
+
+		body, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			t.Errorf("ioutil.ReadAll(r.Body) got: %v want: nil", err)
```
kele, line 35: got err

tests/requestparsing/contentlengthandtransferencoding_test.go
```diff
@@ -0,0 +1,446 @@
...
+		}
+
+		if got := string(body); got != "ABCDE" {
+			t.Errorf(`r.Body got: %q want: "ABCDE"`, got)
+		}
+	})
+	if err != nil {
+		t.Errorf("MakeRequest() got: %v want: nil", err)
```
kele, line 43: This looks like a Fatal.

tests/requestparsing/contentlengthandtransferencoding_test.go
```diff
@@ -0,0 +1,446 @@
...
+		if err != nil {
+			t.Errorf("ioutil.ReadAll(r.Body) got: %v want: nil", err)
+		}
+
+		if got := string(body); got != "" {
+			t.Errorf(`r.Body got: %q want: ""`, got)
+		}
+	})
+	if err != nil {
+		t.Errorf("ERROR: %v", err)
```
kele, line 100: Make it consistent.

tests/requestparsing/basicauth_test.go
```diff
@@ -0,0 +1,117 @@
...
+				username, password, ok := r.BasicAuth()
+				if ok != tt.want.ok || username != tt.want.username || password != tt.want.password {
+					t.Errorf("username, password, ok := r.BasicAuth() got: %q, %q, %v want: %q, %q, %v", username, password, ok, tt.want.username, tt.want.password, tt.want.ok)
+				}
+			})
+			if err != nil {
+				t.Errorf("MakeRequest() got: %v want: nil", err)
+			}
+
+			if !bytes.HasPrefix(resp, []byte(statusOK)) {
```
kele, line 111: I'm on the edge with making this a test helper. I'd like to have another reviewer chime in here, so please keep this open.

tests/requestparsing/contentlengthandtransferencoding_test.go
```diff
@@ -0,0 +1,446 @@
...
+	}
+
+	if !bytes.HasPrefix(resp, []byte(statusOK)) {
+		got := string(resp[:bytes.IndexByte(resp, '\n')+1])
+		t.Errorf("status code got: %q want: %q", got, statusOK)
+	}
+}
```
kele, line 412: It seems to me all of this could be a tabletest with:
```
tests := []struct{
  name string
  request string
  verifyHandler func(*http.Request)
  wantStatus string
} {
  // ...
}
```

Right?

tests/requestparsing/header_test.go
```diff
@@ -0,0 +1,197 @@
...
+)
+
+var headerTests = []struct {
+	name    string
+	request []byte
+	want    map[string][]string
+}{
+	{
+		name: "Basic",
+		// All tests use verbatim newline characters instead of using multiline strings to ensure that \r and \n end up in exactly the right places.
```
kele, line 21: This is uncomfortably long, please add some linebreaks. Also, consider adding the comment a few levels higher instead of by the first test case.

tests/requestparsing/basicauth_test.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+type basicAuthWant struct {
+	headers  map[string][]string
+	ok       bool
+	username string
+	password string
+}
+
+var basicAuthTests = []struct {
```
kele, line 21: Define these inside the test function.

mattiasgrenfeldt: Both `basicAuthWant` and `basicAuthTests`? 
Similar for all table tests?

tests/requestparsing/contentlengthandtransferencoding_test.go
mattiasgrenfeldt: Ah, sloppy copy pasting!

tests/requestparsing/basicauth_test.go
kele: Yup.

tests/requestparsing/basicauth_test.go
mattiasgrenfeldt: Fixed in 46df8b2. Does this look good?

tests/requestparsing/contentlengthandtransferencoding_test.go
mattiasgrenfeldt: Yes! But all `verifyHandler`s would look really similar.
You could also convert most of the tests in the file to a table test that looks like this:
```go
tests := []struct{
  name string
  request string
  want wantStruct{
    headers map[string][]string
    contentLength int
    transferEncoding []string
    body string
  }
} {
  // ...
}
```
And put all tests with non-`200 OK` status codes in a separate table test.
What do you think about this solution?

mattiasgrenfeldt: I implemented my solution in 83d3413. What do you think?

tests/requestparsing/README.md
```diff
@@ -0,0 +1,6 @@
+
```
bsiegert, line 1: nit: remove this empty line

tests/requestparsing/basicauth_test.go
bsiegert: I would make it a helper that just extracts the status from the reponse and compare that string.

```
if got, want := extractStatus(resp), statusOK; got != want {
  t.Errorf(...)
}
```

tests/requestparsing/basicauth_test.go
```diff
@@ -0,0 +1,170 @@
...
+				if password != tt.want.password {
+					t.Errorf("_, password, _ := r.BasicAuth() got: %q want: %q", password, tt.want.password)
+				}
+			})
+			if err != nil {
+				t.Fatalf("MakeRequest() got err: %v", err)
+			}
+
+			if !bytes.HasPrefix(resp, []byte(statusOK)) {
+				got := string(resp[:bytes.IndexByte(resp, '\n')+1])
```
bsiegert, line 165: Won't this give you the trailing newline as well? That will make the error message look weird, with a linebreak in between.

tests/requestparsing/contentlengthandtransferencoding_test.go
```diff
@@ -0,0 +1,347 @@
...
+				"\r\n"),
+			want: testWant{
+				headers:          map[string][]string{"Content-Length": []string{"5"}},
+				contentLength:    5,
+				transferEncoding: nil,
+				body:             "ABCDE",
+			},
+		},
+		{
+			name: "ContentButNoContentLength",
```
bsiegert, line 44: General comment: these names do not have to be camel case. I usually see regular strings with spaces.

tests/requestparsing/basicauth_test.go
mattiasgrenfeldt: I'm printing the error messages with `%q` so it will be escaped like this: `"HTTP/1.1 200 OK\r\n"`.
Should I removed the line breaks and the switch to `%s`?

bsiegert: Ah right, %q escapes the newline characters.

I proposed in another comment to have a helper that extracts just the first line, without the \r\n.

tests/requestparsing/basicauth_test.go
kele: +1.

tests/requestparsing/README.md
```diff
@@ -0,0 +1,5 @@
+# Request Parsing tests
```
kele, line 1: So, we've added README.md in the `testing/` directory mostly because there will be no packages there directly, so it might give people a general idea of what to expect there.

That being said, package documentation should live in godoc, so keep it there: https://blog.golang.org/godoc. You can either create a `doc.go` file or just put it on top of one of the package files.

tests/requestparsing/basicauth_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/google/go-safeweb/testing/requesttesting"
+)
+
+func TestBasicAuth(t *testing.T) {
+	type testWant struct {
+		headers  map[string][]string
+		ok       bool
+		username string
+		password string
+	}
```
kele, line 20: I don't generally have an issue with creating a new struct for `want` values, even though I've mostly seen just a `wantFoo` prefix.

In this case, I think it is a bit problematic, because "ok" doesn't mean much. I'd recommend having this instead:

```
type basicAuth {
  username string
  password string
  ok bool
}
var tests = []struct{
  name string
  request []byte
  wantBasicAuth basicAuth
  wantHeaders map[string][]string
}{...}
```

tests/requestparsing/README.md
```diff
@@ -0,0 +1,5 @@
...
+
+This package contains tests for verify the request parsing behavior of `net/http` in Go's standard library.
```
empijei, line 3:
```suggestion
This package contains tests to verify the request parsing behavior of `net/http` in Go's standard library.
```

tests/requestparsing/contentlengthandtransferencoding_test.go
```diff
@@ -0,0 +1,347 @@
...
+					t.Errorf("r.TransferEncoding mismatch (-want +got):\n%s", diff)
+				}
+
+				if r.ContentLength != tt.want.contentLength {
+					t.Errorf("r.ContentLength got: %v want: %v", r.ContentLength, tt.want.contentLength)
+				}
+
+				body, err := ioutil.ReadAll(r.Body)
+				if err != nil {
+					t.Errorf("ioutil.ReadAll(r.Body) got err: %v", err)
```
empijei, line 219: I think this should be a `t.Fatal`. If you got an unespected I/O error it doesn't make much sense to assert anything on the returned body.

kele:
@mattiasgrenfeldt please rebase and resolve the conflicts. I can merge then.

kele:
Great work. :)

PR #22, opened 2020-07-15
Moved test harness to internal

mattiasgrenfeldt:
We talked about moving the test harness to the `internal` folder to avoid having both a `testing` package and a `tests` package. We also didn't want the test harness to be importable when using the `safehttp` package.


mihalimara22:
Nice, thanks for doing this! LGMT

Issue #9, opened 2020-07-08
Add a high-level design document


kele:
CC: @empijei

kele:
this should not have been closed.

kele:
ETA: EOW

kele:
What remains:
- [ ] resolve `safehttp` godoc TODOs
- [ ] developer guide

PR #24, opened 2020-07-16
Added a document about goals.

kele:
Part of #9.

The design parts have been purposefully left out. We'll start adding these as we go.


kele:
Thanks, Benny!

PR #25, opened 2020-07-16
Added test for ordering of http.Header.Values()

mattiasgrenfeldt:
A question came up in #21 whether the `http.Header.Values()` function sorts the resulting string slice somehow. I added a test to verify that the current behavior is to order the slice in the same order as the headers were sent from the client.

### DIFF
--- a/tests/headers/header_test.go
+++ b/tests/headers/header_test.go
@@ -205,3 +205,47 @@ func TestStatusCode(t *testing.T) {
 		})
 	}
 }
+
+func TestValues(t *testing.T) {
+	var tests = []struct {
+		name    string
+		request []byte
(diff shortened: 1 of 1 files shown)


tests/headers/header_test.go
```diff
@@ -205,3 +205,47 @@ func TestStatusCode(t *testing.T) {
 		})
 	}
 }
+
+func TestValues(t *testing.T) {
```
mihalimara22, line 209: can you add a comment about what this test is doing, please? :)

mattiasgrenfeldt:
@kele and @empijei you are the only ones that can accept and merge since only you two have write access :)

PR #28, opened 2020-07-21
Initial structure of the safe form type

### LINKED ISSUE #19: Retrieve form values consistently
In Go, accessing the form values associated to one key can be done using multiple functions, some with different return values. These can cause inconsistencies in the handlers' implementation of a server.  We want to design and implement a way to mitigate this problem.

mihalimara22:
Fixes #19 

Created the initial  structure of safehttp.Form with some preliminary tests

### DIFF
--- a/safehttp/form.go
+++ b/safehttp/form.go
@@ -19,7 +19,7 @@ import (
 	"strconv"
 )
 
-// Form contains parsed data either from URL's query or form parameters, part of
+// Form contains parsed data from form parameters, part of
 // the body of POST, PATCH or PUT requests that are not multipart requests. The
 // form values will only be available after parsing the form, and only through
 // the getter functions.
@@ -28,7 +28,7 @@ type Form struct {
 	err    error
 }
 
-// Int64 checks whether key param maps to any query or form parameter
+// Int64 checks whether key param maps to any form parameter
 // values. In case it does, it will try to convert the first value to a valid
 // int64 and return it. If there are no values associated with param, it
 // will return the defaultValue value. If the first value is not an integer, it will
@@ -46,7 +46,7 @@ func (f *Form) Int64(param string, defaultValue int64) int64 {
 	return paramVal
 }
 
-// Uint64 checks whether key param maps to any query or form parameter
+// Uint64 checks whether key param maps to any form parameter
 // values. In case it does, it will try to convert the first valid
 // uint64 and return it. If there are no values associated with
 // param, it will return the defaultValue value. If the first value is not an
@@ -65,7 +65,7 @@ func (f *Form) Uint64(param string, defaultValue uint64) uint64 {
 	return paramVal
 }
 
-// String checks whether key param maps to any query or form parameter
+// String checks whether key param maps to any form parameter
 // values. In case it does, it will return the first value. If it doesn't, it
 // will return the defaultValue value.
 func (f *Form) String(param string, defaultValue string) string {
@@ -76,7 +76,7 @@ func (f *Form) String(param string, defaultValue string) string {
 	return vals[0]
 }
 
-// Float64 checks whether key param maps to any query or form parameter
+// Float64 checks whether key param maps to any form parameter
 // values. In case it does, it will try to convert the first value to a valid
 // float64 and return it. If there are no values associated with param, it will
 // return the defaultValue value. If the first value is not a float, it will return
@@ -94,7 +94,7 @@ func (f *Form) Float64(param string, defaultValue float64) float64 {
 	return paramVal
 }
 
-// Bool checks whether key param maps to any query or form parameter
+// Bool checks whether key param maps to any form parameter
 // values. In case it does, it will try to convert the first value to a valid
 // bool and return it. If there are no values associated with param, it will
 // return the defaultValue value. If the first value is not a boolean, it will return
@@ -133,7 +133,7 @@ func clearSlice(slicePtr interface{}) error {
 	return nil
 }
 
-// Slice checks whether key param maps to any query or form parameters. If it
+// Slice checks whether key param maps to any form parameters. If it
 // does, it will try to convert them to the type of slice elements slicePtr
 // points to. If there are no values associated with param, it will clear the
 // slice. If type conversion fails at any point, the Form error field will be
--- a/safehttp/form.go
+++ b/safehttp/form.go
@@ -1,0 +1,225 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package safehttp
+
+import (
+	"errors"
+	"mime/multipart"
+	"strconv"
+)
+
+// Form TODO
+type Form struct {
+	values map[string][]string
+	err    error
+	parsed bool
+}
+
+// MultipartForm TODO
+type MultipartForm struct {
+	form Form
+	file map[string][]*multipart.FileHeader
+}
+
+// Int TODO
+func (f *Form) Int(paramName string, defaultValue int) int {
+	const fnInt = "Int"
+	if f.err != nil {
+		return defaultValue
+	}
+	if !f.parsed {
+		panic(fnInt + ": form has not been parsed")
+	}
+	vals, ok := f.values[paramName]
+	if !ok {
+		f.err = errors.New(fnInt + ": no value found for key " + paramName)
+		return defaultValue
+	}
+	paramVal, err := strconv.Atoi(vals[0])
+	if err != nil {
+		f.err = errors.New(fnInt + ": " + err.Error())
+		return defaultValue
+	}
+	return paramVal
+}
+
+// Uint TODO
+func (f *Form) Uint(paramName string, defaultValue uint64) uint64 {
+	const fnUint = "Uint"
+	if f.err != nil {
+		return defaultValue
+	}
+	if !f.parsed {
+		panic(fnUint + ": form has not been parsed")
+	}
+	vals, ok := f.values[paramName]
+	if !ok {
+		f.err = errors.New(fnUint + ": no value found for key " + paramName)
+		return defaultValue
+	}
+	paramVal, err := strconv.ParseUint(vals[0], 10, 0)
+	if err != nil {
+		f.err = errors.New(fnUint + ": " + err.Error())
+		return defaultValue
+	}
+	return paramVal
+}
+func (f *Form) String(paramName string, defaultValue string) string {
+	const fnString = "String"
+	if f.err != nil {
+		return defaultValue
+	}
+	if !f.parsed {
+		panic(fnString + ": form has not been parsed")
+	}
+	vals, ok := f.values[paramName]
+	if !ok {
+		f.err = errors.New(fnString + ": no value found for key " + paramName)
+		return defaultValue
+	}
+	return vals[0]
+}
+
+// Float64 TODO
+func (f *Form) Float64(paramName string, defaultValue float64) float64 {
+	const fnFloat64 = "Float64"
+	if f.err != nil {
+		return defaultValue
+	}
+	if !f.parsed {
+		panic(fnFloat64 + ": form has not been parsed")
+	}
+	vals, ok := f.values[paramName]
+	if !ok {
+		f.err = errors.New(fnFloat64 + "no value found for key " + paramName)
+		return defaultValue
+	}
+	paramVal, err := strconv.ParseFloat(vals[0], 64)
+	if err != nil {
+		f.err = errors.New(fnFloat64 + ": " + err.Error())
+		return defaultValue
+	}
+	return paramVal
+}
+
+// Bool TODO
+func (f *Form) Bool(paramName string, defaultValue bool) bool {
+	const fnBool = "Bool"
+	if f.err != nil {
+		return defaultValue
+	}
+	if !f.parsed {
+		panic(fnBool + ": form has not been parsed")
+	}
+	vals, ok := f.values[paramName]
+	if !ok {
+		f.err = errors.New(fnBool + ": no value found for key " + paramName)
+		return defaultValue
+	}
+	if vals[0] != "true" {
+		if vals[0] != "false" {
+			f.err = errors.New(fnBool + ": values of form parameter " + paramName + " not a boolean")
+		}
+		return false
+	}
+	return true
+}
+
+// Slice TODO
+func (f *Form) Slice(slice interface{}, paramName string) {
+	const fnSlice = "Slice"
+	if f.err != nil {
+		slice = nil
+		return
+	}
+	if !f.parsed {
+		panic(fnSlice + ": form has not been parsed")
+	}
+	vals, ok := f.values[paramName]
+	if !ok {
+		f.err = errors.New(fnSlice + ": no value found for key " + paramName)
+		slice = nil
+	}
+	len := len(vals)
+	switch data := slice.(type) {
+	case *[]string:
+		res := *data
+		res = make([]string, 0, len)
+		for _, x := range vals {
+			res = append(res, x)
+		}
+	case *[]int:
+		res := *data
+		res = make([]int, 0, len)
+		for _, x := range vals {
+			x, err := strconv.Atoi(x)
+			if err != nil {
+				f.err = errors.New(fnSlice + ": " + err.Error())
+				slice = nil
+				return
+			}
+			res = append(res, x)
+		}
+	case *[]uint64:
+		res := *data
+		res = make([]uint64, 0, len)
+		for _, x := range vals {
+			x, err := strconv.ParseUint(x, 10, 0)
+			if err != nil {
+				f.err = errors.New(fnSlice + ": " + err.Error())
+				slice = nil
+				return
+			}
+			res = append(res, x)
+		}
+	case *[]float64:
+		res := *data
+		res = make([]float64, 0, len)
+		for _, x := range vals {
+			x, err := strconv.ParseFloat(x, 64)
+			if err != nil {
+				f.err = errors.New(fnSlice + ": " + err.Error())
+				slice = nil
+				return
+			}
+			res = append(res, x)
+		}
+	case *[]bool:
+		res := *data
+		res = make([]bool, 0, len)
+		for _, x := range vals {
+			if x != "true" {
+				if x != "false" {
+					f.err = errors.New(fnSlice + ": " + ": values of form parameter " + paramName + " not a boolean")
+					slice = nil
+					return
+				}
+				res = append(res, false)
+				continue
+			}
+			res = append(res, true)
+		}
+	default:
+		f.err = errors.New(fnSlice + ": slice type not supported")
+		slice = nil
+	}
+
+	return
+}
+
+// Error TODO
+func (f *Form) Error() error {
+	return f.err
+}
--- a/safehttp/form.go
+++ b/safehttp/form.go
@@ -134,70 +134,100 @@ func (f *Form) Bool(paramName string, defaultValue bool) bool {
 
 // Slice TODO
 func (f *Form) Slice(slice interface{}, paramName string) {
-	if f.err != nil {
-		slice = nil
-		return
-	}
 	if !f.parsed {
 		panic("form has not been parsed")
 	}
-	vals, ok := f.values[paramName]
-	if !ok {
-		f.err = errors.New("no value found for key " + paramName)
-		slice = nil
-	}
-	len := len(vals)
-	switch data := slice.(type) {
+	mapVals, ok := f.values[paramName]
+	switch values := slice.(type) {
 	case *[]string:
-		res := *data
-		res = make([]string, 0, len)
-		for _, x := range vals {
+		if f.err != nil {
+			*values = nil
+			return
+		}
+		if !ok {
+			f.err = errors.New("no value found for key " + paramName)
+			*values = nil
+		}
+		res := make([]string, 0, len(mapVals))
+		for _, x := range mapVals {
 			res = append(res, x)
 		}
+		*values = res
 	case *[]int:
-		res := *data
-		res = make([]int, 0, len)
-		for _, x := range vals {
+		if f.err != nil {
+			*values = nil
+			return
+		}
+		if !ok {
+			f.err = errors.New("no value found for key " + paramName)
+			*values = nil
+		}
+		res := make([]int, 0, len(mapVals))
+		for _, x := range mapVals {
 			x, err := strconv.Atoi(x)
 			if err != nil {
 				f.err = err
-				slice = nil
+				*values = nil
 				return
 			}
 			res = append(res, x)
+			*values = res
 		}
 	case *[]uint64:
-		res := *data
-		res = make([]uint64, 0, len)
-		for _, x := range vals {
+		if f.err != nil {
+			*values = nil
+			return
+		}
+		if !ok {
+			f.err = errors.New("no value found for key " + paramName)
+			*values = nil
+		}
+		res := make([]uint64, 0, len(mapVals))
+		for _, x := range mapVals {
 			x, err := strconv.ParseUint(x, 10, 0)
 			if err != nil {
 				f.err = err
-				slice = nil
+				*values = nil
 				return
 			}
 			res = append(res, x)
 		}
+		*values = res
 	case *[]float64:
-		res := *data
-		res = make([]float64, 0, len)
-		for _, x := range vals {
+		if f.err != nil {
+			*values = nil
+			return
+		}
+		if !ok {
+			f.err = errors.New("no value found for key " + paramName)
+			*values = nil
+		}
+		res := make([]float64, 0, len(mapVals))
+		for _, x := range mapVals {
 			x, err := strconv.ParseFloat(x, 64)
 			if err != nil {
 				f.err = err
-				slice = nil
+				*values = nil
 				return
 			}
 			res = append(res, x)
 		}
+		*values = res
 	case *[]bool:
-		res := *data
-		res = make([]bool, 0, len)
-		for _, x := range vals {
+		if f.err != nil {
+			*values = nil
+			return
+		}
+		if !ok {
+			f.err = errors.New("no value found for key " + paramName)
+			*values = nil
+		}
+		res := make([]bool, 0, len(mapVals))
+		for _, x := range mapVals {
 			if x != "true" {
 				if x != "false" {
 					f.err = errors.New(": values of form parameter " + paramName + " not a boolean")
-					slice = nil
+					*values = nil
 					return
 				}
 				res = append(res, false)
@@ -205,11 +235,10 @@ func (f *Form) Slice(slice interface{}, paramName string) {
 			}
 			res = append(res, true)
 		}
+		*values = res
 	default:
 		f.err = errors.New("slice type not supported")
-		slice = nil
 	}
-
 	return
 }
 
--- a/safehttp/form.go
+++ b/safehttp/form.go
@@ -24,7 +24,6 @@ import (
 type Form struct {
 	values map[string][]string
 	err    error
-	parsed bool
 }
 
 // MultipartForm TODO
@@ -37,9 +36,6 @@ type MultipartForm struct {
 func (f *Form) Int(paramName string, defaultValue int) int {
 	if f.err != nil {
 		return defaultValue
-	}
-	if !f.parsed {
-		panic("form has not been parsed")
 	}
 	vals, ok := f.values[paramName]
 	if !ok {
@@ -59,9 +55,6 @@ func (f *Form) Uint(paramName string, defaultValue uint64) uint64 {
 	if f.err != nil {
 		return defaultValue
 	}
-	if !f.parsed {
-		panic("form has not been parsed")
-	}
 	vals, ok := f.values[paramName]
 	if !ok {
 		f.err = errors.New("no value found for key " + paramName)
@@ -78,9 +71,6 @@ func (f *Form) String(paramName string, defaultValue string) string {
 	if f.err != nil {
 		return defaultValue
 	}
-	if !f.parsed {
-		panic("form has not been parsed")
-	}
 	vals, ok := f.values[paramName]
 	if !ok {
 		f.err = errors.New("no value found for key " + paramName)
@@ -93,9 +83,6 @@ func (f *Form) String(paramName string, defaultValue string) string {
 func (f *Form) Float64(paramName string, defaultValue float64) float64 {
 	if f.err != nil {
 		return defaultValue
-	}
-	if !f.parsed {
-		panic("form has not been parsed")
 	}
 	vals, ok := f.values[paramName]
 	if !ok {
@@ -115,9 +102,6 @@ func (f *Form) Bool(paramName string, defaultValue bool) bool {
 	if f.err != nil {
 		return defaultValue
 	}
-	if !f.parsed {
-		panic("form has not been parsed")
-	}
 	vals, ok := f.values[paramName]
 	if !ok {
 		f.err = errors.New("no value found for key " + paramName)
@@ -134,9 +118,6 @@ func (f *Form) Bool(paramName string, defaultValue bool) bool {
 
 // Slice TODO
 func (f *Form) Slice(slice interface{}, paramName string) {
-	if !f.parsed {
-		panic("form has not been parsed")
-	}
 	mapVals, ok := f.values[paramName]
 	switch values := slice.(type) {
 	case *[]string:
--- a/safehttp/form.go
+++ b/safehttp/form.go
@@ -225,6 +225,7 @@ func (f *Form) Slice(slicePtr interface{}, paramKey string) {
 			}
 		}
 		*values = res
+
 	default:
 		f.err = clearSlice(slicePtr)
 	}
--- a/safehttp/form.go
+++ b/safehttp/form.go
@@ -31,17 +31,17 @@ type Form struct {
 	err    error
 }
 
-// Int checks whether key paramKey maps to any query or form parameter
-// values. In case it does, it will try to convert the first value to an integer
-// and return it. If there are no values associated with paramKey, it will
-// return the default value. If the first value is not an integer, it will
+// Int64 checks whether key paramKey maps to any query or form parameter
+// values. In case it does, it will try to convert the first value to a 64-bit
+// integer and return it. If there are no values associated with paramKey, it
+// will return the default value. If the first value is not an integer, it will
 // return the default value and set the Form error field.
-func (f *Form) Int(paramKey string, defaultValue int) int {
+func (f *Form) Int64(paramKey string, defaultValue int64) int64 {
 	vals, ok := f.values[paramKey]
 	if !ok {
 		return defaultValue
 	}
-	paramVal, err := strconv.Atoi(vals[0])
+	paramVal, err := strconv.ParseInt(vals[0], 10, 64)
 	if err != nil {
 		f.err = err
 		return defaultValue
@@ -49,18 +49,18 @@ func (f *Form) Int(paramKey string, defaultValue int) int {
 	return paramVal
 }
 
-// Uint checks whether key paramKey maps to any query or form parameter
+// Uint64 checks whether key paramKey maps to any query or form parameter
 // values. In case it does, it will try to convert the first value to an
-// unsigned integer and return it. If there are no values associated with
+// 64-bit unsigned integer and return it. If there are no values associated with
 // paramKey, it will return the default value. If the first value is not an
 // unsigned integer, it will return the default value and set the Form
 // error field.
-func (f *Form) Uint(paramKey string, defaultValue uint64) uint64 {
+func (f *Form) Uint64(paramKey string, defaultValue uint64) uint64 {
 	vals, ok := f.values[paramKey]
 	if !ok {
 		return defaultValue
 	}
-	paramVal, err := strconv.ParseUint(vals[0], 10, 0)
+	paramVal, err := strconv.ParseUint(vals[0], 10, 64)
 	if err != nil {
 		f.err = err
 		return defaultValue
@@ -122,7 +122,7 @@ func clearSlice(slicePtr interface{}) error {
 	switch vs := slicePtr.(type) {
 	case *[]string:
 		*vs = nil
-	case *[]int:
+	case *[]int64:
 		*vs = nil
 	case *[]float64:
 		*vs = nil
@@ -154,10 +154,10 @@ func (f *Form) Slice(slicePtr interface{}, paramKey string) {
 			res = append(res, x)
 		}
 		*values = res
-	case *[]int:
-		res := make([]int, 0, len(mapVals))
+	case *[]int64:
+		res := make([]int64, 0, len(mapVals))
 		for _, x := range mapVals {
-			x, err := strconv.Atoi(x)
+			x, err := strconv.ParseInt(x, 10, 64)
 			if err != nil {
 				f.err = err
 				*values = nil
@@ -169,7 +169,7 @@ func (f *Form) Slice(slicePtr interface{}, paramKey string) {
 	case *[]uint64:
 		res := make([]uint64, 0, len(mapVals))
 		for _, x := range mapVals {
-			x, err := strconv.ParseUint(x, 10, 0)
+			x, err := strconv.ParseUint(x, 10, 64)
 			if err != nil {
 				f.err = err
 				*values = nil
--- a/safehttp/form.go
+++ b/safehttp/form.go
@@ -31,27 +31,17 @@ type Form struct {
 	err    error
 }
 
-// MultipartForm extends the Form structure to define a POST, PATCH or PUT
-// request that has Content-Type: multipart/form-data. Its fields are only
-// available after parsing the form, through getter functions that specify the
-// type.
-type MultipartForm struct {
-	Form
-	file map[string][]*multipart.FileHeader
-}
-
 // Int checks whether key paramKey maps to any query or form parameter
-// values. In case it does, it will try to convert the first value to an integer and
-// return it. If the first value is not an integer or there are no values
-// associated with paramKey, it will return the default value and set the Form
-// error field.
+// values. In case it does, it will try to convert the first value to an integer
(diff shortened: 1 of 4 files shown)


safehttp/form.go
```diff
@@ -0,0 +1,225 @@
...
+
+// MultipartForm TODO
+type MultipartForm struct {
+	form Form
+	file map[string][]*multipart.FileHeader
+}
+
+// Int TODO
+func (f *Form) Int(paramName string, defaultValue int) int {
+	const fnInt = "Int"
```
kele, line 38: Drop the `fnInt` convention. There is not much value in adding the prefix. panics include stack traces

mihalimara22: Resolved in [43ad129](url)

safehttp/form.go
```diff
@@ -0,0 +1,219 @@
...
+	values map[string][]string
+	err    error
+	parsed bool
+}
+
+// MultipartForm TODO
+type MultipartForm struct {
+	form Form
+	file map[string][]*multipart.FileHeader
+}
```
kele, line 34: Keep the definition of a struct with it's method's definition together in the file.

safehttp/form.go
```diff
@@ -0,0 +1,219 @@
...
+	"errors"
+	"mime/multipart"
+	"strconv"
+)
+
+// Form TODO
+type Form struct {
+	values map[string][]string
+	err    error
+	parsed bool
```
kele, line 27: Hm... Why do we need this field? Haven't we decided that `req.Form()` and so on will just parse the form, and there is no other way to access the object?

safehttp/form.go
```diff
@@ -0,0 +1,219 @@
...
+func (f *Form) Int(paramName string, defaultValue int) int {
+	if f.err != nil {
+		return defaultValue
+	}
+	if !f.parsed {
+		panic("form has not been parsed")
+	}
+	vals, ok := f.values[paramName]
+	if !ok {
+		f.err = errors.New("no value found for key " + paramName)
```
kele, line 46:
```suggestion
		f.err = fmt.Printf("no value found for key %q", paramName)
```

kele: I'm adding this mostly for `%q` that adds quotes around the string and does escaping as needed.

safehttp/form.go
```diff
@@ -0,0 +1,219 @@
...
+		slice = nil
+		return
+	}
+	if !f.parsed {
+		panic("form has not been parsed")
+	}
+	vals, ok := f.values[paramName]
+	if !ok {
+		f.err = errors.New("no value found for key " + paramName)
+		slice = nil
```
kele, line 147: I'm not sure it does what you think it does.

It makes the **function variable** `slice` point to nil. It doesn't change the underlying slice.

safehttp/form.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	len := len(vals)
```
kele, line 149: Just use `len(vals)` everywhere, especially since you are shadowing the `len` keyword.

safehttp/form.go
```diff
@@ -0,0 +1,219 @@
...
+	switch data := slice.(type) {
```
kele, line 150: `values`? I know that both are very vague, but it seems that `data` is even less concrete.

safehttp/form.go
```diff
@@ -0,0 +1,219 @@
...
+	case *[]string:
+		res := *data
+		res = make([]string, 0, len)
+		for _, x := range vals {
+			res = append(res, x)
```
kele, line 155: You need to write some tests for this, because my hunch is that this code doesn't work as you expected. I'm also fine if you add `Slice` in a separate PR if you want to make this a quick one. :)

safehttp/form.go
mihalimara22: Resolved in [7b2f76b](url)

safehttp/form.go
mihalimara22: Resolved in [7b2f76b](url)

safehttp/form.go
mihalimara22: Partly resolved in [7b2f76b](url), more tests to come

safehttp/form.go
mihalimara22: It does not, thanks for pointing out. Resolved in [7b2f76b](url)

safehttp/form.go
mihalimara22: We don't actually, I was overly careful but the user can't access the form if it hasn't been parsed in the current design. Thanks for spotting it. Resolved in [1f3a543](url)

safehttp/form_test.go
```diff
@@ -0,0 +1,195 @@
...
+	}{
+		{
+			name:    "valid int in GET request",
+			req:     getReq,
+			resp:    status200OK,
+			formVal: 10,
+		},
+		{
+			name:    "valid int in POST non-multipart request",
+			req:     postReq,
```
kele, line 94:
```suggestion
			req:     func() *http.Request{
						req := httptest.NewRequest("POST", "/", strings.NewReader("pizza=10"))
						req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
						return req
			},
```

safehttp/incoming_request.go
```diff
@@ -14,9 +14,67 @@
...
 
 // IncomingRequest TODO
 type IncomingRequest struct {
 	req *http.Request
 }
+
+// QueryForm TODO
+func (r IncomingRequest) QueryForm() (*Form, error) {
+	if r.req.Method != "GET" {
+		return &Form{}, errors.New("got request method " + r.req.Method + ", want GET")
```
kele, line 31: return `nil`

kele, line 31: Use `errors.New` only when the message is not parametrized. Use `fmt.Errorf` instead here.

safehttp/incoming_request.go
```diff
@@ -14,9 +14,67 @@
...
+	}
+	err := r.req.ParseForm()
+	if err != nil {
+		return &Form{}, err
+	}
+	return &Form{values: r.req.Form, err: nil}, nil
```
kele, line 37: You can omit `err: nil`, the zero value of `err` is going to be `nil`.

safehttp/incoming_request.go
```diff
@@ -14,9 +14,67 @@
...
+
+// QueryForm TODO
+func (r IncomingRequest) QueryForm() (*Form, error) {
+	if r.req.Method != "GET" {
+		return &Form{}, errors.New("got request method " + r.req.Method + ", want GET")
+	}
+	err := r.req.ParseForm()
+	if err != nil {
+		return &Form{}, err
+	}
```
kele, line 36:
```suggestion
	if err := r.req.ParseForm(); err != nil {
		return nil, err
	}
```

safehttp/form_test.go
mihalimara22: Thanks! Resolved in [1287192](url)

safehttp/form.go
```diff
@@ -0,0 +1,234 @@
...
+		f.err = errors.New("no value found for key " + paramName)
+		return defaultValue
+	}
+	if vals[0] != "true" {
+		if vals[0] != "false" {
+			f.err = errors.New("values of form parameter " + paramName + " not a boolean")
+		}
+		return false
+	}
+	return true
```
empijei, line 121: `switch`?

safehttp/form.go
```diff
@@ -0,0 +1,234 @@
...
+	switch values := slice.(type) {
+	case *[]string:
+		if f.err != nil {
+			*values = nil
+			return
+		}
+		if !ok {
+			f.err = errors.New("no value found for key " + paramName)
+			*values = nil
+		}
```
empijei, line 136: instead of having this check in every branch of the switch, can we have it before the switch?

safehttp/form.go
```diff
@@ -0,0 +1,234 @@
...
+
+// Slice TODO
+func (f *Form) Slice(slice interface{}, paramName string) {
+	mapVals, ok := f.values[paramName]
+	switch values := slice.(type) {
+	case *[]string:
+		if f.err != nil {
+			*values = nil
+			return
+		}
```
empijei, line 132: see my comment above, I'm not sure this is the best behavior we can have.

safehttp/form.go
```diff
@@ -0,0 +1,234 @@
...
+					return
+				}
+				res = append(res, false)
+				continue
+			}
+			res = append(res, true)
+		}
+		*values = res
+	default:
+		f.err = errors.New("slice type not supported")
```
empijei, line 226: this could not be a slice, just use `fmt.Errorf("type not supported in Slice call: %T", slice)`

safehttp/form.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	return
+}
+
+// Error TODO
+func (f *Form) Error() error {
+	return f.err
+}
```
empijei, line 234: I'm not sure we want Form to implement the `error` interface. Maybe call this `Err`?

safehttp/form.go
```diff
@@ -0,0 +1,234 @@
...
+type Form struct {
+	values map[string][]string
+	err    error
+}
+
+// MultipartForm extends the Form structure to define a POST, PATCH or PUT
+// request that is multipart. Its fields are only available after parsing the
+// form, through getter functions that specify the type.
+type MultipartForm struct {
+	*Form
```
empijei, line 36: why a pointer here? MultipartForm will always be used by ref, so we can embed a value here.

safehttp/incoming_request.go
```diff
@@ -14,9 +14,66 @@
...
+func (r IncomingRequest) PostForm() (*Form, error) {
+	if r.req.Method != "POST" && r.req.Method != "PATCH" && r.req.Method != "PUT" {
+		return &Form{}, errors.New("got request method " + r.req.Method + ", want POST/PATCH/PUT")
+	}
+
+	if ct := r.req.Header.Get("Content-Type"); ct != "application/x-www-form-urlencoded" {
+		return &Form{}, errors.New("invalid method called for Content-Type:" + ct + ", want MultipartForm")
+	}
+	err := r.req.ParseForm()
+	if err != nil {
```
empijei, line 50:
```suggestion
	if err := r.req.ParseForm(); err != nil {
```

safehttp/incoming_request.go
```diff
@@ -14,9 +14,66 @@
...
+	"strings"
+)
 
 // IncomingRequest TODO
 type IncomingRequest struct {
 	req *http.Request
 }
+
+// QueryForm TODO
+func (r IncomingRequest) QueryForm() (*Form, error) {
```
empijei, line 29: maybe we should wrap this and the other getters in a `sync.Once` call or have another memoization for the values, otherwise every time the getter is called a lot of work needs to be done.

safehttp/form.go
kele:
```
x := f.Int("x", 0)
y := f.Int("y", 0)
if err := f.Error(); err != nil {
  return err // here the information about which one errored is lost
}

File: safehttp/form.go (hunk 2 above)
Comment (mihalimara22) [reply safehttp/form.go]: As discussed, I used embedding in[542ad4d](url) , so I left the two types in the same file.

File: safehttp/form.go (hunk 4 above)
Comment (mihalimara22) [reply safehttp/form.go]: Resolved in [2dc01f6](url)

File: safehttp/incoming_request.go (hunk 10 above)
Comment (mihalimara22) [reply safehttp/incoming_request.go]: Resolved in [2dc01f6](url)

Comment (mihalimara22) [reply safehttp/incoming_request.go]: Resolved in [2dc01f6](url)

File: safehttp/incoming_request.go (hunk 12 above)
Comment (mihalimara22) [reply safehttp/incoming_request.go]: Resolved in [2dc01f6](url)

File: safehttp/form.go [hunk 21]
```diff
@@ -0,0 +1,234 @@
...
+	file map[string][]*multipart.FileHeader
+}
+
+// Int TODO
+func (f *Form) Int(paramName string, defaultValue int) int {
+	if f.err != nil {
+		return defaultValue
+	}
```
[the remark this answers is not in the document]
mihalimara22: I followed this article: https://blog.golang.org/errors-are-values in which, when an error occurs, the subsequent operation become no-ops. Should I create a helper function or have a different behaviour?

safehttp/form.go
mihalimara22: Good question... resolved in [d511d47](url)

safehttp/incoming_request.go
mihalimara22: Oh, true, thanks for this! Resolved in [d511d47](url)

safehttp/form.go
mihalimara22: Resolved in [d511d47](url)

safehttp/form.go
mihalimara22: Thanks, resolved in [d511d47](url).

safehttp/form.go
mihalimara22: Implemented a solution in [d511d47](url)

safehttp/form.go
mihalimara22: Yep, sorry. Resolved in [d511d47](url)

safehttp/form.go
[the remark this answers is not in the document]
kele:
```
x := f.Int("x", 1)
y := f.Int("y", 2)
z := f.Err()
```

What should the values of `x`, `y` and `z` be if reading `x` results in a parsing error? How can we make it unsurprising?

safehttp/form.go
```diff
@@ -0,0 +1,252 @@
...
+	switch vs := slicePtr.(type) {
+	case *[]string:
+		*vs = nil
+	case *[]int:
+		*vs = nil
+	case *[]float64:
+		*vs = nil
+	case *[]uint64:
+		*vs = nil
+	case *[]bool:
```
kele, line 161: missed?

safehttp/form.go
```diff
@@ -0,0 +1,252 @@
...
+		return true
+	case "false":
+		return false
+	default:
+		f.err = fmt.Errorf("values of form parameter %q not a boolean", paramKey)
+	}
+	return false
+}
+
+// clearSlice is a helper functions that clears the slice slicePtr points to when an error occured
```
kele, line 150: The comment doesn't reflect the reality (anyone can call it, even if an error didn't occur). Also, this doesn't have to be and shouldn't be a method of Form.

Drop the comment and make it a free function. GoDoc comments are not required for unexported symbols. Also, the name is clear.

mihalimara22: Thanks, resolved in [ 43c86aa](url)

safehttp/form.go
mihalimara22: wups, sorry - resolved in [ 43c86aa](url)

safehttp/form.go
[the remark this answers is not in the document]
mihalimara22: Hmm, I'd expect a developer not to care about the value of x and y if an error occurred in parsing and treat those as garbage since something went wrong and that should be handled. No-oping everything after an error occurred is less surprising than continuing parsing imo. I am leaning towards having this behaviour mainly because we always report the first error that occurred. Otherwise, if we keep parsing, the error might get overwritten in the current implementation. If you prefer the later, I'd say to keep a slice with all the errors that occurred.

safehttp/form.go
```diff
@@ -0,0 +1,253 @@
...
+// Form contains parsed data either from URL's query or from
+// form parameters, part of the body of POST, PATCH or PUT requests that are not
+// multipart requests. The fields are only available after parsing the form,
+// through getter functions that specify the type. If parsing failed, Form will
+// be set to nil. Field err will only be set if
+// an error occurs when the user tries to access a parameter.
+type Form struct {
+	values map[string][]string
+	err    error
+}
```
kele, line 32: Keep the definition of the struct next to the methods (i.e. move `MultipartForm` somewhere else/below).

safehttp/form.go
```diff
@@ -0,0 +1,253 @@
...
+
+// Slice checks whether key paramKey maps to any query or form parameters. If it
+// does, it will try to convert them to the type of slice elements slicePtr
+// points to. If there are no values associated with paramKey or type conversion
+// fails at any point, the Form error field will be set and slicePtr will point
+// to nil.
+func (f *Form) Slice(slicePtr interface{}, paramKey string) {
+	if f.err != nil {
+		f.err = clearSlice(slicePtr)
+	}
```
kele, line 176: Why?

safehttp/form.go
```diff
@@ -0,0 +1,253 @@
...
+	return
+}
+
+// Err returns the value of the Form error field. This will be nil unless an
+// error occured while accessing a parsed form value.
+func (f *Form) Err() error {
+	return f.err
+}
```
kele, line 251: Wasn't the plan to return and clear? So you can easily do:

```
x := f.Int("x", 0)
if f.Err() != nil {
  // Well, x is not *that* important. Let's continue.
}
y := f.Int("y", 0)
if f.Err() != nil {
  panic("y is very important!")
}
```

I actually like to hear back from other folks about this.

kele: http://doc/1OHXUwC-sOK4Uk6-By7_lPW8ioJRydd7AuO4gIMnUjNU#bookmark=id.6ndng7y6lrs8

safehttp/form.go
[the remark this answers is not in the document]
hazcod: To pitch in, i've seen strange bugs when someone forgets to handle an error and continues with (partly) initialized return values.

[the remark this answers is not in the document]
kele: It basically boils down to this: if an error occurs, should it be recoverable or not? Form values are controlled by the client, and if a client passes incorrect form values, I'd expect that the server should not put extra effort to fix this and just assume the whole form is broken.

safehttp/form.go
```diff
@@ -0,0 +1,253 @@
...
+// error field.
+func (f *Form) Int(paramKey string, defaultValue int) int {
+	if f.err != nil {
+		return defaultValue
+	}
+	vals, ok := f.values[paramKey]
+	if !ok {
+		f.err = fmt.Errorf("no value found for key %q", paramKey)
+		return defaultValue
+	}
```
kele, line 56: Hm... Don't we have default return values to make this a non-error?

safehttp/form.go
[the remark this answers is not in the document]
hazcod: Plus, it prevents reflective-type of attacks too. I guess it impacts traditional web apps tough (and not SPAs), but someone can always re-insert those values.

safehttp/form.go
mihalimara22: Seeing that usecase, it would make sense.. will keep refactoring on hold until we make a decision on this

safehttp/form.go
mihalimara22: Resolved in [c5059f7](url)

safehttp/form.go
mihalimara22: Resolved in [c5059f7](url)

safehttp/form.go
```diff
@@ -0,0 +1,249 @@
...
+			*values = res
+		}
+	case *[]uint64:
+		res := make([]uint64, 0, len(mapVals))
+		for _, x := range mapVals {
+			x, err := strconv.ParseUint(x, 10, 0)
+			if err != nil {
+				f.err = err
+				*values = nil
+				slicePtr = values
```
mattiasgrenfeldt, line 195: Should this row be removed? You are already setting `slicePtr` to nil by setting `*values = nil` on the row above

safehttp/form.go
```diff
@@ -0,0 +1,249 @@
...
+				*values = nil
+				return
+			}
+			res = append(res, x)
+			*values = res
+		}
+	case *[]uint64:
+		res := make([]uint64, 0, len(mapVals))
+		for _, x := range mapVals {
+			x, err := strconv.ParseUint(x, 10, 0)
```
mattiasgrenfeldt, line 191: Same comment as in `UInt()`. This should be `strconv.ParseUint(x, 10, 64)`

safehttp/incoming_request.go
```diff
@@ -14,9 +14,80 @@
...
+func (r *IncomingRequest) QueryForm() (f *Form, err error) {
+	r.parseOnce.Do(func() {
+		if r.req.Method != "GET" {
+			f, err = nil, fmt.Errorf("got request method %s, want GET", r.req.Method)
+		}
+		if err := r.req.ParseForm(); err != nil {
+			f = nil
+		}
+	})
+	f, err = &Form{values: r.req.Form}, nil
```
mattiasgrenfeldt, line 41: Don't you want a `if err != nil { return ...` before this row?

safehttp/incoming_request.go
```diff
@@ -14,9 +14,80 @@
...
+		}
+
+		if ct := r.req.Header.Get("Content-Type"); ct != "application/x-www-form-urlencoded" {
+			f, err = nil, fmt.Errorf("invalid method called for Content-Type: %s, want MultipartForm", ct)
+		}
+		if err := r.req.ParseForm(); err != nil {
+			f = &Form{}
+		}
+	})
+	f, err = &Form{values: r.req.PostForm}, nil
```
mattiasgrenfeldt, line 61: Same here as above with `if err != nil { return ...`

safehttp/incoming_request.go
```diff
@@ -14,9 +14,80 @@
...
+	parseOnce sync.Once
+}
+
+// QueryForm parses the query parameters provided in the request. It returns
+// the parsed parameters as a Form object, if no error occured, or the parsing error otherwise.
+func (r *IncomingRequest) QueryForm() (f *Form, err error) {
+	r.parseOnce.Do(func() {
+		if r.req.Method != "GET" {
+			f, err = nil, fmt.Errorf("got request method %s, want GET", r.req.Method)
+		}
```
mattiasgrenfeldt, line 36: Don't you want to add a return in this if-statement? Like so:
```go
f, err = ...
return
```
And similar in all other ifs in `r.parseOnce.Do` inside of `QueryForm`, `PostForm` and `MultipartForm`

safehttp/incoming_request.go
```diff
@@ -14,9 +14,80 @@
...
+	return
+}
+
+// MultipartForm parses the form parameters provided in the body of a POST,
+// PATCH or PUT request that has Content-Type set tomultipart/form-data. It
+// returns a MultipartForm object containing the parsed form parameter and
+// files, if no error occured, or the parsing error otherwise.
+func (r *IncomingRequest) MultipartForm(maxMemory int64) (f *MultipartForm, err error) {
+	r.parseOnce.Do(func() {
+		const defaultMaxMemory = 32 << 20
```
mattiasgrenfeldt, line 71: `32 << 20` seems like a pretty arbitrary limit. 33 megabytes. Is there any reason behind picking this number?
It should probably be mentioned in the comment above that `maxMemory` is capped at this value.

safehttp/form.go
```diff
@@ -0,0 +1,249 @@
...
+// error field.
+func (f *Form) Uint(paramKey string, defaultValue uint64) uint64 {
+	if f.err != nil {
+		return defaultValue
+	}
+	vals, ok := f.values[paramKey]
+	if !ok {
+		return defaultValue
+	}
+	paramVal, err := strconv.ParseUint(vals[0], 10, 0)
```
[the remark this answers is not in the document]
mihalimara22: Indeed, switched to `int64` and `uint64` in [f50431e](url)

safehttp/incoming_request.go
mihalimara22: Thanks, I missed this bug :)

safehttp/incoming_request.go
mihalimara22: That is the maxDefaultMemory in the `net/http` implementation so it's mostly to be backwards compatible, but I guess it would be a good thing  to look more into it.

safehttp/form.go
mihalimara22: good spot ^^

safehttp/form_test.go
```diff
@@ -194,4 +270,787 @@ func TestValidIntSlice(t *testing.T) {
...
+					httptest.NewRequest("GET", "/?pizza=diavola", nil),
+					httptest.NewRequest("GET", "/?pizza=ăȚâȘî", nil),
+					httptest.NewRequest("GET", "/?pizza=\x64\x69\x61\x76\x6f\x6c\x61", nil),
+				}
+				return &reqs
+			}(),
+			formVals: []string{"diavola", "ăȚâȘî", "diavola"},
+		},
+		{
+			name: "Valid int in POST non-multipart request",
```
mattiasgrenfeldt, line 363: Should be `Valid string in POST non-multipart request`

safehttp/form_test.go
```diff
@@ -194,4 +270,787 @@ func TestValidIntSlice(t *testing.T) {
...
+		{
+			name: "Invalid booleans in GET request",
+			reqs: func() *[]*http.Request {
+				reqs := []*http.Request{
+					httptest.NewRequest("GET", "/?pizza=TruE", nil),
+					httptest.NewRequest("GET", "/?pizza=potato", nil),
+				}
+				return &reqs
+			}(),
+			formVals: []bool{false, false},
```
mattiasgrenfeldt, line 673: This doesn't need to be a slice since both values are the same

safehttp/form_test.go
```diff
@@ -194,4 +270,787 @@ func TestValidIntSlice(t *testing.T) {
...
+
+func TestFormValidSlice(t *testing.T) {
+	validSlices := []interface{}{
+		[]int64{-8, 9, -100}, []uint64{8, 9, 10}, []string{"margeritta", "diavola", "calzone"}, []float64{1.3, 8.9, -4.1}, []bool{true, false, true},
+	}
+	// Parsing behaviour of native types that are supported is not included as
+	// it was tested in previous tests.
+	tests := []struct {
+		name string
+		want *[]interface{}
```
mattiasgrenfeldt, line 748: want is not needed. You can just use `validSlices` everywhere instead.

safehttp/incoming_request.go
```diff
@@ -14,9 +14,100 @@
...
+	return
+}
+
+// MultipartForm parses the form parameters provided in the body of a POST,
+// PATCH or PUT request that has Content-Type set to multipart/form-data. It
+// returns a MultipartForm object containing the parsed form parameter and
+// files, if no error occurred, or the parsing error together with a nil
+// MultipartForm otherwise. This method should  only be used when the user expects
+// a POST request with the Content-Type: multipart/form-data header.
+func (r *IncomingRequest) MultipartForm(maxMemory int64) (f *MultipartForm, err error) {
```
stapelberg, line 82: here and above: Make this return “(*MultipartForm, error)”, no need to name these.

safehttp/form_test.go
```diff
@@ -0,0 +1,1056 @@
...
+			form, err := getParsedForm(ir)
+			if err != nil {
+				t.Fatalf(`getParsedForm: got "%v", want nil`, err)
+			}
+			var wantInt int64 = 0
+			gotInt := form.Int64("pizzaInt", 0)
+			if diff := cmp.Diff(wantInt, gotInt); diff != "" {
+				t.Errorf("form.Int64: got %v, want %v, diff (-want +got): \n%s", gotInt, wantInt, diff)
+			}
+			if err = form.Err(); test.errs[0].Error() != err.Error() {
```
stapelberg, line 1028: nit: here and elsewhere: use := where you can, to reduce the scope of variables as much as possible

safehttp/form.go
```diff
@@ -0,0 +1,229 @@
...
+// and return it. If there are no values associated with paramKey, it will
+// return the default value. If the first value is not a boolean, it will return
+// the default value and set the Form error field.
+func (f *Form) Bool(paramKey string, defaultValue bool) bool {
+	vals, ok := f.values[paramKey]
+	if !ok {
+		return defaultValue
+	}
+	switch vals[0] {
+	case "true":
```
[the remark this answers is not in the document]
mihalimara22: I've noticed but chose to opt for the obvious cases for now, as we didn't discussed the other ones yet. I'll open an issue since this PR is pretty big already.

safehttp/form_test.go
mihalimara22: Thanks :)

safehttp/form.go
```diff
@@ -0,0 +1,229 @@
...
+// the body of POST, PATCH or PUT requests that are not  multipart requests. The
+// form values will only be available after parsing the form, and only through
+// the getter functions.
+type Form struct {
+	values map[string][]string
+	err    error
+}
+
+// Int64 checks whether key paramKey maps to any query or form parameter
+// values. In case it does, it will try to convert the first value to a 64-bit
```
kele, line 33: Why "query or form"? This is because you reuse the `Form` type for query params, right? That seems quite confusing. :|

Can we please focus on just `Form` and `MultiForm` in this PR?

Here and the other places.

safehttp/form.go
```diff
@@ -0,0 +1,229 @@
...
+// integer and return it. If there are no values associated with paramKey, it
+// will return the default value. If the first value is not an integer, it will
```
kele, line 35: a valid int64

Here and in other places.

safehttp/form.go
```diff
@@ -0,0 +1,229 @@
...
+// return the default value and set the Form error field.
+func (f *Form) Int64(paramKey string, defaultValue int64) int64 {
```
kele, line 37: Simplify to: "param" and "default". Here and all other functions.

safehttp/form.go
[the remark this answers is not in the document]
kele: Please file an issue to reconsider this. Let's keep just `true` and `false` for now.

safehttp/form.go
```diff
@@ -0,0 +1,229 @@
...
+		f.err = clearSlice(slicePtr)
+		return
+	}
+	switch values := slicePtr.(type) {
+	case *[]string:
+		res := make([]string, 0, len(mapVals))
+		for _, x := range mapVals {
+			res = append(res, x)
+		}
+		*values = res
```
kele, line 154:
```suggestion
		*values = append(res, mapVals...)
```

safehttp/form.go
```diff
@@ -0,0 +1,229 @@
...
+// MultipartForm extends the Form structure to define a POST, PATCH or PUT
+// request that has Content-Type: multipart/form-data. Its fields are only
+// available after parsing the form, through getter functions that specify the
+// type.
+type MultipartForm struct {
+	Form
+	file map[string][]*multipart.FileHeader
+}
+
+// TODO(@mihalimara22): Create getters and tests for the `file` field in MultipartForm
```
kele, line 229: So there's nothing there done so far for this type? Drop it then.

safehttp/form.go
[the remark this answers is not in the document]
mihalimara22: That's true, good spot, converted everything to `int64` and `uint64` as mentioned in the design doc

safehttp/incoming_request.go
mihalimara22: Yep, indeed, resolved in [0f32eb1](url)

safehttp/form.go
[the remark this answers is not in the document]
mihalimara22: Done :)

safehttp/form.go
kele: Yes.

safehttp/form.go
kele: Plasae fix.

safehttp/form.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	return nil
+}
+
+// Slice checks whether key param maps to any query or form parameters. If it
+// does, it will try to convert them to the type of slice elements slicePtr
+// points to. If there are no values associated with param, it will clear the
+// slice. If type conversion fails at any point, the Form error field will be
+// set and the slice will be cleared.
+func (f *Form) Slice(slicePtr interface{}, param string) {
```
kele, line 141: IMHO it looks better to do `f.Slice("names", &s)` rather than `f.Slice(&s, "names")`, so you might consider reversing the order.

safehttp/form_test.go
```diff
@@ -0,0 +1,1051 @@
...
+	if !strings.HasPrefix(r.req.Header.Get("Content-Type"), "multipart/form-data") {
+		f, err := r.PostForm()
+		return f, err
+	}
+	mf, err := r.MultipartForm(32 << 20)
+	return &mf.Form, err
+}
+
+func TestFormValidInt(t *testing.T) {
+	maxIntToString := strconv.FormatInt(math.MaxInt64, 10)
```
kele, line 75: It's not a function or anything like that, so the name `xToY` doesn't really fit.

safehttp/form_test.go
```diff
@@ -0,0 +1,1051 @@
...
+				form, err := getParsedForm(ir)
+				if err != nil {
+					t.Fatalf(`getParsedForm: got "%v", want nil`, err)
+				}
+				want := test.formVal
+				got := form.Int64("pizza", 0)
+				if diff := cmp.Diff(want, got); diff != "" {
+					t.Errorf("form.Int64: got %v, want %v, diff (-want +got): \n%s", got, want, diff)
+				}
+				if err := form.Err(); test.err.Error() != err.Error() {
```
kele, line 196: Do not compare error strings. It is fine to just check that an error is returned, unless you programatically depend on the semantics/structure of the error - which we don't.

safehttp/form_test.go
```diff
@@ -0,0 +1,1051 @@
...
+	mf, err := r.MultipartForm(32 << 20)
+	return &mf.Form, err
+}
+
+func TestFormValidInt(t *testing.T) {
+	maxIntToString := strconv.FormatInt(math.MaxInt64, 10)
+	tests := []struct {
+		name    string
+		req     *http.Request
+		formVal int64
```
kele, line 79: Make the expected values clear by either using a `want` name or a prefix. In this case, just `want` is enough.

safehttp/form_test.go
```diff
@@ -0,0 +1,1051 @@
...
+				multipartReqBody := "--123\r\n" +
+					"Content-Disposition: form-data; name=\"pizza\"\r\n" +
+					"\r\n" +
+					"9223372036854775810\r\n" +
+					"--123--\r\n"
+				multipartReq := httptest.NewRequest("POST", "/", strings.NewReader(multipartReqBody))
+				multipartReq.Header.Set("Content-Type", `multipart/form-data; boundary="123"`)
+				res := []*http.Request{getReq, postReq, multipartReq}
+				return &res
+			}(),
```
kele, line 159:
```suggestion
			reqs: []*http.Request{
				httptest.NewRequest("GET", "/?pizza=9223372036854775810", nil),
				func() *http.Request {
					postReq := httptest.NewRequest("POST", "/", strings.NewReader("pizza=9223372036854775810"))
					postReq.Header.Set("Content-Type", "application/x-www-form-urlencoded")
					return postReq
				}(),
				func() *http.Request {
					multipartReqBody := "--123\r\n" +
						"Content-Disposition: form-data; name=\"pizza\"\r\n" +
						"\r\n" +
						"9223372036854775810\r\n" +
						"--123--\r\n"
					multipartReq := httptest.NewRequest("POST", "/", strings.NewReader(multipartReqBody))
					multipartReq.Header.Set("Content-Type", `multipart/form-data; boundary="123"`)
					return multipartReq
				}(),
			},
```

safehttp/form_test.go
```diff
@@ -0,0 +1,1051 @@
...
+		if respStatus := recorder.Result().Status; respStatus != status200OK {
+			t.Errorf("response status: got %s, want %s", respStatus, status200OK)
+		}
+	}
+}
+
+func TestFormInvalidInt(t *testing.T) {
+	tests := []struct {
+		name    string
+		reqs    *[]*http.Request
```
kele, line 140: It doesn't have to be a pointer to a slice.

safehttp/form_test.go
```diff
@@ -0,0 +1,1051 @@
...
+	case *template.Template:
+		return x.Execute(rw, data)
+	default:
+		panic("not a safe response type")
+	}
+}
+
+// A helper function that returns a parsed Form or an empty Form and any errors
+// that occured during parsing
+func getParsedForm(r *IncomingRequest) (*Form, error) {
```
kele, line 60: This helper should not be used. It should be clear from the test what it does, whether it asks for a PostForm or a MultipartForm.

Introducing extra logic to tests makes them complicated and error prone.

safehttp/form_test.go
```diff
@@ -0,0 +1,1051 @@
...
+	for _, test := range tests {
+		m := NewMachinery(func(rw ResponseWriter, ir *IncomingRequest) Result {
+			form, err := getParsedForm(ir)
+			if err != nil {
+				t.Fatalf(`getParsedForm: got "%v", want nil`, err)
+			}
+			want := test.formVal
+			got := form.Uint64("pizza", 0)
+			if err := form.Err(); err != nil {
+				t.Errorf(`form.Error: got "%v", want nil`, err)
```
kele, line 258: nit: `form.Err()`

kele, line 258: nit: No need for the `"%v"`, just `%v` is enough.

safehttp/incoming_request.go
mattiasgrenfeldt: This is not resolved in 0f32eb1. The godoc comment still doesn't mention the capping of `maxMemory`

mihalimara22: Thanks, sorry! :) [b3cb723](url) resolves it

safehttp/form.go
mihalimara22: `default` is a keyword in go so we are going to stick to `defaultVal`

safehttp/form.go
mihalimara22: You're right, thanks :)

kele: Missed.

safehttp/form_test.go
mihalimara22: Is it a good idea to use type assertion here in order to see whether I'm getting a Form or a MultipartForm?

safehttp/form_test.go
```diff
@@ -0,0 +1,1110 @@
...
+			var form *Form
+			if !strings.HasPrefix(ir.req.Header.Get("Content-Type"), "multipart/form-data") {
+				var err error
+				form, err = ir.PostForm()
+				if err != nil {
+					t.Fatalf(`ir.PostForm: got %v, want nil`, err)
+				}
+			} else {
+				mf, err := ir.MultipartForm(32 << 20)
+				form = &mf.Form
```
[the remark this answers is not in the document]
mihalimara22: That would make more sense, thanks! :)

safehttp/form.go
```diff
@@ -0,0 +1,223 @@
...
+	return paramVal
+}
+
+// Uint64 checks whether key param maps to any form parameter
+// values. In case it does, it will try to convert the first valid
+// uint64 and return it. If there are no values associated with
+// param, it will return the defaultValue value. If the first value is not an
+// unsigned integer, it will return the defaultValue value and set the Form
+// error field.
+func (f *Form) Uint64(param string, defaultValue uint64) uint64 {
```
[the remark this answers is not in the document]
mihalimara22: Thanks :)

safehttp/form_test.go
```diff
@@ -0,0 +1,1110 @@
...
+			}(),
+			want: math.MaxInt64,
+		},
+	}
+
+	for _, test := range tests {
+		m := NewMachinery(func(rw ResponseWriter, ir *IncomingRequest) Result {
+			var form *Form
+			if !strings.HasPrefix(ir.req.Header.Get("Content-Type"), "multipart/form-data") {
+				var err error
```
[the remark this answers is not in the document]
mattiasgrenfeldt: oh, sorry, you are totally right! :)

safehttp/form_test.go
```diff
@@ -0,0 +1,1110 @@
...
+			if diff := cmp.Diff(want, got); diff != "" {
+				t.Errorf("form.Int64: got %v, want %v, diff (-want +got): \n%s", got, want, diff)
+			}
+			return Result{}
+		}, &dispatcher{})
+		recorder := httptest.NewRecorder()
+		m.HandleRequest(recorder, test.req)
+		if respStatus := recorder.Result().Status; respStatus != status200OK {
+			t.Errorf("response status: got %s, want %s", respStatus, status200OK)
+		}
```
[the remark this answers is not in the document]
mihalimara22: I think that's a good suggestion. Tbf, I can remove the constants as they don't bring much value.

safehttp/form_test.go
```diff
@@ -0,0 +1,1110 @@
...
+	}
+}
+
+func TestFormInvalidInt(t *testing.T) {
+	tests := []struct {
+		name string
+		reqs []*http.Request
+		err  error
```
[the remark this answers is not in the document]
mattiasgrenfeldt: True! I don't know what is best here. Your call

safehttp/form_test.go
```diff
@@ -0,0 +1,1110 @@
...
+						"\r\n" +
+						"diavola\r\n" +
+						"--123--\r\n"
+					multipartReq := httptest.NewRequest("POST", "/", strings.NewReader(multipartReqBody))
+					multipartReq.Header.Set("Content-Type", `multipart/form-data; boundary="123"`)
+					return multipartReq
+				}(),
+			},
+			err:  errors.New(`strconv.ParseFloat: parsing "diavola": invalid syntax`),
+			want: 0.0},
```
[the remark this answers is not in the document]
mattiasgrenfeldt: This is unresolved

safehttp/form.go
```diff
@@ -0,0 +1,222 @@
...
+type Form struct {
+	values map[string][]string
+	err    error
+}
+
+// Int64 checks whether key param maps to any form parameter
+// values. In case it does, it will try to convert the first value to a valid
+// int64 and return it. If there are no values associated with param, it
+// will return the defaultValue value. If the first value is not an integer, it will
+// return the defaultValue value and calling Err() will return the parsing error.
```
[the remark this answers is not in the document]
mihalimara22: Thanks :D

safehttp/form.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	return nil
+}
+
+// Slice checks whether key param maps to any form parameters. If it
+// does, it will try to convert them to the type of slice elements slicePtr
+// points to. If there are no values associated with param, it will clear the
+// slice. If type conversion fails at any point, the slice will be cleared and
+// calling Err() will return the parsing error.
+func (f *Form) Slice(param string, slicePtr interface{}) {
```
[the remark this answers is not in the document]
mihalimara22: hmm, the alternative is to call the other functions within `Slice`.. maybe, it's a better option

safehttp/incoming_request.go
kele: I don't get this. Why is this capped to 32MB? Just because it's the default, doesn't mean it should be a hard cap.

kele: I'm going to merge this PR but open an issue for this.,

PR #29, opened 2020-07-27
Clearly document strange header parsing behavior

### LINKED ISSUE #23: Document strange parsing behavior
To document the strange behavior we (@mattiasgrenfeldt and @mihalimara22) have found while testing the current behavior of `net/http` in #5 and #6 we should do the following: For each behavior we have observed that is strange or undesirable we should:

- Create a separate test function that tests this behavior.
- Create one subtest in the function that documents the **current** behavior.
- Create one subtest in the function that documents the **desired** behavior. Use `t.Skip()` so that the tests pass CI.
- Document the behavior using comments and explain why it is strange or undesirable.

(Issue created on @kele 's request.)

mattiasgrenfeldt:
Adresses #23 for the header tests.
All of the tests that show strange or undesirable behavior have now been more clearly documented.

### DIFF
--- a/tests/headers/basicauth_test.go
+++ b/tests/headers/basicauth_test.go
@@ -166,7 +166,7 @@ func TestBasicAuth(t *testing.T) {
 
 func TestBasicAuthOrdering(t *testing.T) {
 	// The documentation of http.Request.BasicAuth() doesn't clearly specify
-	// that only the first Authorization header is used and that the other ones
+	// that only the first Authorization header is used and the other ones
 	// are ignored. This could potentially lead to security issues if two
 	// HTTP servers that look at different headers are chained together.
 	//
--- a/tests/headers/basicauth_test.go
+++ b/tests/headers/basicauth_test.go
@@ -98,23 +98,6 @@ func TestBasicAuth(t *testing.T) {
 			wantHeaders: map[string][]string{"Authorization": []string{"xasic UGl6emE6UGFzc3dvcmQ6UGFzc3dvcmQ="}},
 		},
 		{
-			name: "Ordering",
-			request: []byte("GET / HTTP/1.1\r\n" +
-				"Host: localhost:8080\r\n" +
-				// Base64 encoding of "AAA:aaa".
-				"Authorization: basic QUFBOmFhYQ==\r\n" +
-				// Base64 encoding of "BBB:bbb".
-				"Authorization: basic QkJCOmJiYg==\r\n" +
-				"\r\n"),
-			wantBasicAuth: basicAuth{
-				username: "AAA",
-				password: "aaa",
-				ok:       true,
-			},
-			// Base64 encoding of "AAA:aaa" and then of "BBB:bbb" in that order.
-			wantHeaders: map[string][]string{"Authorization": []string{"basic QUFBOmFhYQ==", "basic QkJCOmJiYg=="}},
-		},
-		{
 			name: "CasingOrdering1",
 			request: []byte("GET / HTTP/1.1\r\n" +
 				"Host: localhost:8080\r\n" +
@@ -179,4 +162,71 @@ func TestBasicAuth(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestBasicAuthOrdering(t *testing.T) {
+	// It is not clear from the documentation of http.Request.BasicAuth()
+	// that only the first Authorization header is used and any other ones
+	// are ignored. This could potentially lead to security issues if two
+	// HTTP servers are chained that look at different headers.
+	//
+	// The desired behavior would instead be for http.Request.BasicAuth() to
+	// return ok as false when there are more than one Authorization header.
+
+	request := []byte("GET / HTTP/1.1\r\n" +
+		"Host: localhost:8080\r\n" +
+		// Base64 encoding of "AAA:aaa".
+		"Authorization: basic QUFBOmFhYQ==\r\n" +
+		// Base64 encoding of "BBB:bbb".
+		"Authorization: basic QkJCOmJiYg==\r\n" +
+		"\r\n")
+
+	t.Run("Current behavior", func(t *testing.T) {
+		resp, err := requesttesting.MakeRequest(context.Background(), request, func(r *http.Request) {
+			// Base64 encoding of "AAA:aaa" and then of "BBB:bbb" in that order.
+			wantHeaders := map[string][]string{"Authorization": []string{"basic QUFBOmFhYQ==", "basic QkJCOmJiYg=="}}
+			if diff := cmp.Diff(wantHeaders, map[string][]string(r.Header)); diff != "" {
+				t.Errorf("r.Header mismatch (-want +got):\n%s", diff)
+			}
+
+			username, password, ok := r.BasicAuth()
+			if want := true; ok != want {
+				t.Errorf("_, _, ok := r.BasicAuth() got: %v want: %v", ok, want)
+			}
+
+			if want := "AAA"; username != want {
+				t.Errorf("username, _, _ := r.BasicAuth() got: %q want: %q", username, want)
+			}
+
+			if want := "aaa"; password != want {
+				t.Errorf("_, password, _ := r.BasicAuth() got: %q want: %q", password, want)
+			}
+		})
+		if err != nil {
+			t.Fatalf("MakeRequest() got err: %v want: nil", err)
+		}
+
+		if got, want := extractStatus(resp), statusOK; got != want {
+			t.Errorf("status code got: %q want: %q", got, want)
+		}
+	})
+
+	t.Run("Desired behavior", func(t *testing.T) {
+		t.Skip()
+		_, err := requesttesting.MakeRequest(context.Background(), request, func(r *http.Request) {
+			// Base64 encoding of "AAA:aaa" and then of "BBB:bbb" in that order.
+			wantHeaders := map[string][]string{"Authorization": []string{"basic QUFBOmFhYQ==", "basic QkJCOmJiYg=="}}
+			if diff := cmp.Diff(wantHeaders, map[string][]string(r.Header)); diff != "" {
+				t.Errorf("r.Header mismatch (-want +got):\n%s", diff)
+			}
+
+			_, _, ok := r.BasicAuth()
+			if want := false; ok != want {
+				t.Errorf("_, _, ok := r.BasicAuth() got: %v want: %v", ok, want)
+			}
+		})
+		if err != nil {
+			t.Fatalf("MakeRequest() got err: %v want: nil", err)
+		}
+	})
 }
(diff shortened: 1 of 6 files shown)


tests/headers/basicauth_test.go
```diff
@@ -180,3 +163,70 @@ func TestBasicAuth(t *testing.T) {
 		})
 	}
 }
+
+func TestBasicAuthOrdering(t *testing.T) {
+	// It is not clear from the documentation of http.Request.BasicAuth()
```
mihalimara22, line 168: Please rephrase a bit e.g:

The documentation of `http.Request.BasicAuth()` doesn't clearly specify whether only the first Authorization header is used and the other ones are ignored. This could potentially lead to security issues if two HTTP servers that look at different headers are chained together.

The desired behavior would be to decline(?) an HTTP request with more than one Authorization header.

or something along those lines

tests/headers/clte_test.go
```diff
@@ -355,3 +300,229 @@ func TestTransferEncodingChunkSizeLength(t *testing.T) {
 		t.Fatalf("MakeRequest() got err: %v", err)
 	}
 }
+
+func TestMultipleTransferEncodingChunkedFirst(t *testing.T) {
+	// When faced with two Transfer-Encoding headers net/http
```
mihalimara22, line 305: comma before headers

tests/headers/clte_test.go
```diff
@@ -355,3 +300,229 @@ func TestTransferEncodingChunkSizeLength(t *testing.T) {
...
+	// currently only looks at the first one and ignores the
+	// rest.
+	//
+	// RFC 7230 doesn’t say anything specifically about
```
mihalimara22, line 309: specific

tests/headers/clte_test.go
```diff
@@ -355,3 +300,229 @@ func TestTransferEncodingChunkSizeLength(t *testing.T) {
...
+	// having multiple TE headers. It only says that in general
```
mihalimara22, line 310: that, in general, ..

tests/headers/clte_test.go
```diff
@@ -355,3 +300,229 @@ func TestTransferEncodingChunkSizeLength(t *testing.T) {
...
+	//   field value in order, separated by a comma.  The order in
+	//   which header fields with the same field name are received
+	//   is therefore significant to the interpretation of the
+	//   combined field value; a proxy MUST NOT change the order
+	//   of these field values when forwarding a message. [...] “
+	// - RFC 7230 Section 3.2.4
+	//
+	// net/http should probably reject requests with two TE headers
+	// just as it does with requests having two CL headers.
+	// I (grenfeldt@) see no legitimate reason for having two TE
```
mihalimara22, line 326: I'd replace  "I .." with "There is..."

tests/headers/clte_test.go
```diff
@@ -355,3 +300,229 @@ func TestTransferEncodingChunkSizeLength(t *testing.T) {
...
+		}
+
+		if got, want := extractStatus(resp), statusNotImplemented; got != want {
+			t.Errorf("status code got: %q want: %q", got, want)
+		}
+	})
+}
+
+func TestTransferEncodingListIdentityFirst(t *testing.T) {
+	// In TestTransferEncodingIdentity we see that the
```
mihalimara22, line 460: I'd avoid pointing the user to the previous test and maybe try to (re)explain the entire issue here.

tests/headers/clte_test.go
```diff
@@ -355,3 +300,229 @@ func TestTransferEncodingChunkSizeLength(t *testing.T) {
...
+	// identity transfer coding should not be supported
+	// at all. But in the current implementation, where
+	// identity is supported, there is a bug. Any transfer
+	// coding that you put after identity in the list of
```
mihalimara22, line 464: Replace "you" with "the user"

tests/headers/host_test.go
```diff
@@ -111,3 +104,53 @@ func TestHostHeaderMultiple(t *testing.T) {
 		t.Errorf("status code got: %q want: %q", got, want)
 	}
 }
+
+func TestAbsoluteFormURLInvalidSchema(t *testing.T) {
+	// (I (grenfeldt@) don't actually know whether this
```
mihalimara22, line 109: I'd leave out this part and maybe open an issue if you think it's something we should investigate :)

tests/headers/referer_test.go
```diff
@@ -105,3 +93,57 @@ func TestReferer(t *testing.T) {
 		})
 	}
 }
+
+func TestRefererOrdering(t *testing.T) {
+	// It is not clear from the documentation of http.Request.Referer()
+	// that only the first Referer header is used and any other ones
+	// are ignored. This could potentially lead to security issues if two
+	// HTTP servers are chained that look at different headers.
+	//
```
mihalimara22, line 102: The documentation of http.Request.Referer() doesn't clearly specify whether only the first Referer header is used and the other ones ignored. This could potentially lead to security issues if two HTTP servers that look at different headers are chained together.

tests/headers/referer_test.go
```diff
@@ -105,3 +93,57 @@ func TestReferer(t *testing.T) {
...
+	// The desired behavior would instead be for http.Request.Referer() to
+	// return an error of some sort or an empty string when there are more
```
mihalimara22, line 104: Can you give an example of some error?

tests/headers/useragent_test.go
```diff
@@ -104,3 +92,57 @@ func TestUserAgent(t *testing.T) {
 		})
 	}
 }
+
+func TestUserAgentOrdering(t *testing.T) {
+	// It is not clear from the documentation of http.Request.UserAgent()
```
mihalimara22, line 97: I'd rephrase similarly to what's above as the phrase is not very clear :)

tests/headers/basicauth_test.go
mattiasgrenfeldt: Thanks, that is more clear! :)
But I don't agree on this phrasing:
`The desired behavior would be to decline(?) an HTTP request with more than one Authorization header.`
We don't want to decline the entire request just because there are more than one Authorization header. We just want the `BasicAuth` function to give an error of some kind (or for the golang team to improve their documentation). Since `BasicAuth` already returns an `ok` bool status I thought we could just set this to `false`.
Or do you think it should be handled differently?

tests/headers/referer_test.go
mattiasgrenfeldt: For `Referer()` and `UserAgent()` they don't currently return an error, only a string, so I thought that to signal that something went wrong we might return an empty string. But I think it would be nicer to add an error return value. The kind of error doesn't really matter as long as it is clear what went wrong. So some examples would be:
```go
errors.New("multiple Referer headers")
errors.New("multiple User-Agent headers")
```

Should this amount of detail really be in the comment though? I think that the specific error message would be up to the person implementing this.

tests/headers/basicauth_test.go
empijei: I think that returning a `Bad Request` status would be acceptable in case of double auth header.
Note that we will likely ban basic auth so the behavior of BasicAuth here is not relevant and [the spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.8) doesn't seem to have an opinion on the matter.

tests/headers/referer_test.go
empijei: I think that here we have to do a risk evaluation: getting the referer wrong could have secuirty implications, getting the UA wrong, not too much.

In any case I don't think we should have too strong opinion on the matter as it is normal practice for cooperating clients to not send duplicate request headers, so for our web framework we'll reject other requests, but for the stdlib I guess it is okay to accept odd requests.

tests/headers/basicauth_test.go
mattiasgrenfeldt: Ok, then I will change the desired behavior to be a `Bad Request`

tests/headers/referer_test.go
mattiasgrenfeldt: I'm gonna go with that the desired behavior should be the behavior we want in our framework. I will therefore change the desired behavior for both tests to be to respond with a `400 Bad Request`.

mattiasgrenfeldt:
Rebased on top of master. Ready to merge.

tests/headers/basicauth_test.go
```diff
@@ -180,3 +163,65 @@ func TestBasicAuth(t *testing.T) {
 		})
 	}
 }
+
+func TestBasicAuthOrdering(t *testing.T) {
+	// The documentation of http.Request.BasicAuth() doesn't clearly specify
+	// that only the first Authorization header is used and that the other ones
```
mihalimara22, line 169: Simplify to "the other ones ignored" so it's easier to read :)

tests/headers/clte_test.go
```diff
@@ -355,3 +300,227 @@ func TestTransferEncodingChunkSizeLength(t *testing.T) {
...
 	}
 }
+
+func TestMultipleTransferEncodingChunkedFirst(t *testing.T) {
+	// When faced with two Transfer-Encoding headers, net/http
+	// currently only looks at the first one and ignores the
+	// rest.
+	//
+	// RFC 7230 doesn’t say anything specific about
+	// having multiple TE headers. It only says that, in general
```
mihalimara22, line 310: missing comma after in general

tests/headers/clte_test.go
```diff
@@ -355,3 +300,227 @@ func TestTransferEncodingChunkSizeLength(t *testing.T) {
...
+		}
+
+		if got, want := extractStatus(resp), statusNotImplemented; got != want {
+			t.Errorf("status code got: %q want: %q", got, want)
+		}
+	})
+}
+
+func TestTransferEncodingListIdentityFirst(t *testing.T) {
+	// There is a bug in the handling of lists of transfer
```
mihalimara22, line 460: "handling of Transfer-Encoding lists"

tests/headers/host_test.go
```diff
@@ -111,3 +104,51 @@ func TestHostHeaderMultiple(t *testing.T) {
 		t.Errorf("status code got: %q want: %q", got, want)
 	}
 }
+
+func TestAbsoluteFormURLInvalidSchema(t *testing.T) {
+	// When sending a request using the absolute
+	// form as the request target, any schema is currently
+	// accepted.
+	//
+	// The desired behavior would instead be that only
```
mihalimara22, line 113: "to only accept http or https as schemas and to respond with a 400 (Bad Request) when the server receives any other schema"

tests/headers/basicauth_test.go
mattiasgrenfeldt: Do you mean this change?
`and that the other ones are ignored.` -> `and that the other ones ignored.`
That feels less grammatically correct. 
I agree that we can remove the `that` though.

kele:
```suggestion
	// that only the first Authorization header is used and the other ones
```

PR #30, opened 2020-07-27
Removed reliance on httptest.ResponseRecorder in test

### LINKED ISSUE #27: Don't assume client is Go in tests.
In `tests/integration/safehtml/safehtml_test.go` and in `tests/integration/header/header_test.go` (not merged yet as of writing this, see #21 ) the `ResponseRecorder` from `net/http/httptest` is used. After calling `ResponseRecorder.Result()` the response is parsed and a `http.Response` object is created. This should be removed since it assumes that the client in our tests is written in Go. Instead simpler comparisons should be used so that we don't rely on the implementation of `ResponseRecorder`.

mattiasgrenfeldt:
Fixes #27 for `tests/integration/safehtml/safehtml_test.go`

### DIFF
--- a/tests/integration/safehtml/safehtml_test.go
+++ b/tests/integration/safehtml/safehtml_test.go
@@ -15,9 +15,10 @@
 package safehtml_test
 
 import (
-	"io/ioutil"
+	"io"
 	"net/http"
 	"net/http/httptest"
+	"strings"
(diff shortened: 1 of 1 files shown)


mattiasgrenfeldt:
Rebased on top of master, ready to merge. @empijei

PR #31, opened 2020-07-27
Remove use of httptest.ResponseRecorder in header tests

mattiasgrenfeldt:
Addresses #27 for tests in `tests/integration/header/header_test.go`.

### DIFF
--- a/tests/integration/header/header_test.go
+++ b/tests/integration/header/header_test.go
@@ -53,7 +53,7 @@ func newResponseWriter(w io.Writer) *responseWriter {
 	return &responseWriter{header: http.Header{}, writer: w, status: http.StatusOK}
 }
 
-func (r responseWriter) Header() http.Header {
+func (r *responseWriter) Header() http.Header {
 	return r.header
 }
 
@@ -61,7 +61,7 @@ func (r *responseWriter) WriteHeader(statusCode int) {
 	r.status = statusCode
 }
 
-func (r responseWriter) Write(data []byte) (int, error) {
+func (r *responseWriter) Write(data []byte) (int, error) {
 	return r.writer.Write(data)
 }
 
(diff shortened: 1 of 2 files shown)


tests/integration/header/header_test.go
```diff
@@ -42,6 +43,28 @@ func (dispatcher) ExecuteTemplate(rw http.ResponseWriter, t safehttp.Template, d
 	return nil
 }
 
+type responseWriter struct {
```
kele, line 46: nit: `fakeResponseWriter`

tests/integration/header/header_test.go
```diff
@@ -42,6 +43,28 @@ func (dispatcher) ExecuteTemplate(rw http.ResponseWriter, t safehttp.Template, d
...
+	header http.Header
+	writer io.Writer
+	status int
+}
+
+func newResponseWriter(w io.Writer) *responseWriter {
+	return &responseWriter{header: http.Header{}, writer: w, status: http.StatusOK}
+}
+
+func (r responseWriter) Header() http.Header {
```
kele, line 56: Keep the receiver type (pointer or not) the same across all methods.

tests/integration/header/header_test.go
kele: Just checked other files: feel free to disregard this for consistency. It's not terribly important.

tests/integration/header/header_test.go
mattiasgrenfeldt: Fixed. But why do you want this? Having just a normal receiver sometimes signals that the function is read-only.

kele: Well, not exactly.

Value receivers are used for passing objects *as values* (copies). Having a value receiver doesn't prevent you from altering objects that `responseWriter` references (e.g. slices, maps, pointers). So it's at best a "shallow" read-only.

mattiasgrenfeldt: Right!

PR #34, opened 2020-07-30
Skeleton for safe multiplexer.

mihalimara22:
`safehttp.ServeMux` is a wrapper around the `http.ServeMux`. Currently, it supports registering based on methods and specifying the list of allowed domains.
Co-authored-by: mattiasgrenfeldt <grenfeldt@google.com>
Co-authored-by: mihalimara22 <maramihali@google.com>

### DIFF
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -30,7 +30,11 @@ const (
 // request against a list of registered patterns and calls the handler for
 // the pattern that most closely matches the URL.
 //
-// Patterns name are fixed, rooted paths, like "/favicon.ico", or rooted
+// When creating the multiplexer, the user needs to specify a list of allowed
+// domains. The server will only serve requests target to those domains and
+// otherwise will reply with HTTP 404 Not Found.
+//
+// Patterns names are fixed, rooted paths, like "/favicon.ico", or rooted
 // subtrees like "/images/" (note the trailing slash). Longer patterns take
 // precedence over shorter ones, so that if there are handlers registered for
 // both "/images/" and "/images/thumbnails/", the latter handler will be called
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -15,7 +15,6 @@
 package safehttp
 
 import (
-	"errors"
 	"net/http"
 )
 
@@ -46,6 +45,7 @@ type ServeMux struct {
 }
 
 // NewServeMux allocates and returns a new ServeMux
+// TODO(@mattias, @mara): make domains a variadic of string **literals**.
 func NewServeMux(dispatcher Dispatcher, domains ...string) *ServeMux {
 	d := map[string]bool{}
 	for _, host := range domains {
@@ -64,15 +64,19 @@ func NewServeMux(dispatcher Dispatcher, domains ...string) *ServeMux {
 func (m *ServeMux) Handle(pattern string, method string, h Handler) {
 	ch, ok := m.handlers[pattern]
 	if !ok {
-		ch := combinedHandler{h: make(map[string]Handler)}
+		ch := combinedHandler{
+			h: make(map[string]Handler),
+			domains: m.domains,
+			d: m.dispatcher,
+		}
 		ch.h[method] = h
 
 		m.handlers[pattern] = ch
-		m.mux.Handle(pattern, ch.createHandler(m.domains, m.dispatcher))
+		m.mux.Handle(pattern, ch)
 		return
 	}
 
-	if _, err := ch.lookup(method); err == nil {
+	if _, ok := ch.h[method]; ok {
 		panic("method already registered")
 	}
 	ch.h[method] = h
@@ -88,87 +92,43 @@ func (m *ServeMux) HandleFunc(pattern string, method string, h HandleFunc) {
 	m.Handle(pattern, method, HandlerFunc(h))
 }
 
-// NotFound registers a handler which will be called when an unhandled
-// path is visited. If another handler has already been registered for
-// this purpose, NotFound panics.
+// NotFound is not yet implemented.
 func (*ServeMux) NotFound(h Handler) {
 	panic("not implemented")
 }
 
-// NotFoundFunc registers a handler which will be called when an unhandled
-// path is visited. If another handler has already been registered for
-// this purpose, NotFound panics.
+// NotFoundFunc is not yet implemented.
 func (*ServeMux) NotFoundFunc(h HandleFunc) {
 	panic("not implemented")
 }
 
-// Handler returns the handler to use for the incoming request and the pattern.
-func (m *ServeMux) Handler(r *IncomingRequest) (Handler, string) {
-	h, pattern := m.mux.Handler(r.req)
-
-	if pattern == "" {
-		// If the pattern is empty, no handler was registered and the handler is
-		// an http.NotFoundHandler
-		// TODO: replace http.NotFoundHandler to a safehttp.NotFoundHandler
-		return safeHandler(h), pattern
-	}
-
-	// See if it is redirect or combined
-	c, ok := m.handlers[pattern]
-	if !ok {
-		// We got a http.RedirectHandler from m.mux.Handler()
-		return safeHandler(h), pattern
-	}
-
-	safeH, err := c.lookup(r.req.Method)
-	if err != nil {
-		// err is nil unless no HTTP method was registered for this pattern
-		// TODO: replace http.NotFoundHandler to a safehttp.MethodNotAllowedHandler
-		return safeHandler(http.NotFoundHandler()), pattern
-	}
-
-	return safeH, pattern
-}
-
-func (m *ServeMux) serveHTTP(w http.ResponseWriter, r *http.Request) {
+// ServeHTTP dispatches the request to the handler whose method matches the
+// incoming request and whose pattern most closely matches the request URL.
+func (m *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	m.mux.ServeHTTP(w, r)
 }
 
 // combinedHandler is collection of handlers based on the request method.
 type combinedHandler struct {
 	// Maps an HTTP method to its handler
-	h map[string]Handler
+	h       map[string]Handler
+	domains map[string]bool
+	d       Dispatcher
 }
 
-// lookup returns the handler associated with the HTTP method provided as an
-// argument and a nil error, unless no handler was registered for the HTTP
-// method.
-func (c *combinedHandler) lookup(httpMethod string) (Handler, error) {
-	h, ok := c.h[httpMethod]
-	if !ok {
-		return nil, errors.New("method not registered")
+// ServeHTTP dispatches the request to the handler associated with
+// the incoming request's method.
+func (c combinedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if !c.domains[r.Host] {
+		http.Error(w, http.StatusText(http.StatusNotFound), http.StatusNotFound)
+		return
 	}
-	return h, nil
-}
 
-// createHandler creates a combined handler to be registered with
-// http.ServeMux.// Handle which calls the correct safe handler based on the
-// request method.
-func (c combinedHandler) createHandler(domains map[string]bool, d Dispatcher) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if !domains[r.Host] {
-			http.Error(w, http.StatusText(http.StatusNotFound), http.StatusNotFound)
-			return
-		}
+	h, ok := c.h[r.Method]
+	if !ok {
+		http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
+		return
+	}
 
-		h, ok := c.h[r.Method]
-		if !ok {
-			http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
-			return
-		}
-
-		rw := newResponseWriter(d, w)
-		ir := newIncomingRequest(r)
-		h.ServeHTTP(rw, &ir)
-	})
+	h.ServeHTTP(newResponseWriter(c.d, w), newIncomingRequest(r))
 }
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -77,16 +77,16 @@ type ServeMux struct {
 }
 
 // NewServeMux allocates and returns a new ServeMux
-func NewServeMux(dispatcher Dispatcher, domains ...string) *ServeMux {
+func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
 	// TODO(@mattiasgrenfeldt, @mihalimara22): make domains a variadic of string **literals**.
-	d := map[string]bool{}
+	dm := map[string]bool{}
 	for _, host := range domains {
-		d[host] = true
+		dm[host] = true
 	}
 	return &ServeMux{
 		mux:      http.NewServeMux(),
-		domains:  d,
-		disp:     dispatcher,
+		domains:  dm,
+		disp:     d,
 		handlers: map[string]methodHandler{},
 	}
 }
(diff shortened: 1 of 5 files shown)


safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+
+// TODO: add the missing methods
+const (
+	// HTTP GET request
+	MethodGet = "GET"
+	// HTTP Post request
+	MethodPost = "POST"
+)
+
+// ServeMux TODO
```
empijei, line 30: TODO?

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// NewServeMux returns a safe multiplex containing the list of allowed domains
+// and dispatcher provided by the user.
+func NewServeMux(dispatcher Dispatcher, domains ...string) *ServeMux {
+	d := map[string]bool{}
+	for _, host := range domains {
+		d[host] = true
+	}
+	return &ServeMux{mux: http.NewServeMux(), domains: d, dispatcher: dispatcher, handlers: map[string]combinedHandler{}}
```
empijei, line 48: Nit: break in multiple lines.

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// Handle registers a handler for the given pattern and method. The options
```
empijei, line 51: please improve doc.

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+// are applied to the middleware. If another handler is already registered
+// for the same pattern and method, Handle panics.
+func (m *ServeMux) Handle(pattern string, h map[string]Handler) {
```
empijei, line 54: I would not use a map here, I would allow to call this multiple times with many methods.

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	return h, nil
+}
+
+// combine creates a combined handler to be registered with http.ServeMux.Handle
+// which calls the correct safe handler based on the request method.
+func (c combinedHandler) combine(domains map[string]bool, d Dispatcher) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if !domains[r.Host] {
+			w.WriteHeader(http.StatusForbidden)
```
empijei, line 142: NotFound

empijei, line 142: Use http.Error instead.

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+
+import (
+	"errors"
+	"net/http"
+)
+
+// TODO: add the missing methods
+const (
+	// HTTP GET request
+	MethodGet = "GET"
```
kele, line 25: Not an initialism or an acronym, but feels it should have the same treatment as them: https://github.com/golang/go/wiki/CodeReviewComments#initialisms. `MethodGET`.

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+	dispatcher Dispatcher
+
+	// Maps user-provided patterns to combined handlers which encapsulate
+	// multiple handlers, each one associated with an HTTP method.
+	handlers map[string]combinedHandler
+}
+
+// NewServeMux returns a safe multiplex containing the list of allowed domains
+// and dispatcher provided by the user.
+func NewServeMux(dispatcher Dispatcher, domains ...string) *ServeMux {
```
kele, line 49: Add a TODO to make `domains` a variadic of string **literals**.

safehttp/mux.go
kele: Line breaks are not strictly necessary in Go code, but this seems like a case where it'd benefit readability.

safehttp/mux.go
kele: :+1:

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+	newH := make(map[string]Handler)
+	for key, val := range h {
+		newH[key] = HandlerFunc(val)
+	}
+	m.Handle(pattern, newH)
+}
+
+// NotFound registers a handler which will be called when an unhandled
+// path is visited. If another handler has already been registered for
+// this purpose, NotFound panics.
```
kele, line 93:
```suggestion
// NotFound is not yet implemented.
```

kele: Same below.

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+
+	safeH, err := c.lookup(r.req.Method)
+	if err != nil {
+		// err is nil unless no HTTP method was registered for this pattern
+		// TODO: replace http.NotFoundHandler to a safehttp.MethodNotAllowedHandler
+		return safeHandler(http.NotFoundHandler()), pattern
+	}
+
+	return safeH, pattern
+}
```
kele, line 131: Can we drop this for now? We don't really need it as it is, and I don't really get the purpose of `safeHandler` yet, and this is the only usage.

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+
+func (m *ServeMux) serveHTTP(w http.ResponseWriter, r *http.Request) {
```
kele, line 133: Hm... What's the point of this unexported method?

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+	m.mux.ServeHTTP(w, r)
+}
+
+// combinedHandler is collection of handlers based on the request method.
+type combinedHandler struct {
+	// Maps an HTTP method to its handler
+	h map[string]Handler
+}
+
+// lookup returns the handler associated with the HTTP method provided as an
```
kele, line 143: You don't need a GoDoc comment here as the symbol is unexported. You can keep it, as it seems like a good opportunity to improve on writing.

Avoid phraes like "returns a nil error, unless...". A nil error indicates full success, and there's usually many ways in which a function can fail. I'd recommend completely omitting the docs about error here, and even return a `bool` instead of `err`, similarly to a `x, ok := mapVar`.

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+// argument and a nil error, unless no handler was registered for the HTTP
+// method.
+func (c *combinedHandler) lookup(httpMethod string) (Handler, error) {
```
kele, line 146: `lookup` is a very generic name. Please come up with something more specific.

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+	h, ok := c.h[httpMethod]
+	if !ok {
+		return nil, errors.New("method not registered")
+	}
+	return h, nil
+}
+
+// combine creates a combined handler to be registered with http.ServeMux.Handle
+// which calls the correct safe handler based on the request method.
+func (c combinedHandler) combine(domains map[string]bool, d Dispatcher) http.Handler {
```
kele: Keep the receivers either values or pointers for all the methods.

safehttp/mux.go
```diff
@@ -0,0 +1,156 @@
...
+
+		h, ok := c.h[r.Method]
+		if !ok {
+			w.WriteHeader(http.StatusMethodNotAllowed)
+			return
+		}
+
+		rw := newResponseWriter(d, w)
+		ir := newIncomingRequest(r)
+		h.ServeHTTP(rw, &ir)
```
kele, line 172:
```suggestion
		h.ServeHTTP(newResponseWriter(d, w), newIncomingRequest(r))
```

safehttp/mux_test.go
```diff
@@ -0,0 +1,196 @@
...
+func TestValidHost(t *testing.T) {
+	mux := NewServeMux(fakeDispatcher{}, "foo.com")
+
+	h := HandlerFunc(func(w ResponseWriter, r *IncomingRequest) Result {
+		return w.Write(safehtml.HTMLEscaped("<h1>Hello World!</h1>"))
+	})
+	mux.Handle("/", map[string]Handler{MethodGet: h})
+
+	req := httptest.NewRequest("GET", "/", nil)
+	req.Host = "foo.com"
```
kele, line 79: You can specify the host in the call to `NewRequest`: http://godoc/pkg/net/http/httptest#NewRequest.

safehttp/mux_test.go
```diff
@@ -0,0 +1,196 @@
...
+	b := &strings.Builder{}
+	rw := newFakeResponseWriter(b)
```
kele, line 81: I'd recommend checking the full response. Here and other tests.

safehttp/mux.go
mattiasgrenfeldt: `net/http` uses `MethodGet` https://pkg.go.dev/net/http?tab=doc#pkg-constants

safehttp/mux.go
mattiasgrenfeldt: For usage in tests

safehttp/mux.go
kele: I guess these were defined before the code review comments. Makes sense for them to stay then.

safehttp/mux.go
kele: And how is the ServeMux used outside of tests then?

mattiasgrenfeldt: Ah, sorry, this should be an exported function!

safehttp/mux.go
mihalimara22: After removing the `Handler` function, `lookup` became quite unnecessary so we removed it

safehttp/mux.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// ServeHTTP dispatches the request to the handler whose method matches the
+// incoming request and whose pattern most closely matches the request URL.
+func (m *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	m.mux.ServeHTTP(w, r)
+}
+
+// combinedHandler is collection of handlers based on the request method.
+type combinedHandler struct {
```
empijei, line 92: can we change this name to be methodHandler or something more evocative of what it does instead of what it is?

safehttp/mux.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+// Handle registers a handler for the given pattern and method. If another
+// handler is already registered for the same pattern and method, Handle panics.
+func (m *ServeMux) Handle(pattern string, method string, h Handler) {
+	ch, ok := m.handlers[pattern]
+	if !ok {
+		ch := combinedHandler{
+			h:       make(map[string]Handler),
```
empijei, line 68: Note, optional change here:

```suggestion
			h:       map[string]Handler{method: h},
```

And remove line 72

safehttp/mux_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+	mux.ServeHTTP(rw, req)
+
+	if want := 404; rw.status != want {
+		t.Errorf("rw.status: got %v want %v", rw.status, want)
+	}
+
+	wantedHeaders := map[string][]string{
+		"Content-Type": {"text/plain; charset=utf-8"}, "X-Content-Type-Options": {"nosniff"},
```
empijei, line 117: Where do you set the X-C-T header? Is this done automatically somewhere?

safehttp/mux_test.go
```diff
@@ -0,0 +1,208 @@
...
+func (fakeDispatcher) ExecuteTemplate(rw http.ResponseWriter, t Template, data interface{}) error {
+	switch x := t.(type) {
+	case *template.Template:
+		return x.Execute(rw, data)
+	default:
+		panic("not a safe response type")
+	}
+}
+
+type fakeResponseWriter struct {
```
empijei, line 49: I would say this is a stubResponseWriter or a responseRecorder

safehttp/mux_test.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/google/safehtml"
+	"github.com/google/safehtml/template"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+type fakeDispatcher struct{}
```
empijei, line 28: this is not a fake I would say. You can call this a simpleDispatcher or testDispatcher.

safehttp/mux_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+func (r *fakeResponseWriter) WriteHeader(statusCode int) {
+	r.status = statusCode
+}
+
+func (r *fakeResponseWriter) Write(data []byte) (int, error) {
+	return r.writer.Write(data)
+}
+
+func TestValidHost(t *testing.T) {
```
empijei, line 71: I think that this and the following one-handler tests might benefit from having a table test that has

```go
tests := []struct{
  request  *http.Request
  wantStatus int
  wantHeader map[string][]string
  wantBody string
} {...}
```
Otherwise it is very hard to see the difference between the various tests and what they are there for.

safehttp/mux_test.go
mattiasgrenfeldt: It's done automatically by `http.Error` https://github.com/golang/go/blob/go1.14.6/src/net/http/server.go#L2050

safehttp/mux.go
```diff
@@ -0,0 +1,113 @@
...
+// ServeHTTP dispatches the request to the handler whose method matches the
+// incoming request and whose pattern most closely matches the request URL.
+func (m *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	m.mux.ServeHTTP(w, r)
+}
+
+// methodHandler is collection of handlers based on the request method.
+type methodHandler struct {
+	// Maps an HTTP method to its handler
+	h       map[string]Handler
```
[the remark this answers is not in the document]
kele: Or a different name. But 'h' is way too short.

safehttp/mux_test.go
```diff
@@ -0,0 +1,204 @@
...
+		t.Errorf("rw.status: got %v want %v", rw.status, want)
+	}
+
+	if diff := cmp.Diff(map[string][]string{}, map[string][]string(rw.header)); diff != "" {
+		t.Errorf("rw.header mismatch (-want +got):\n%s", diff)
+	}
+
+	if got, want := b.String(), "&lt;h1&gt;Hello World! GET&lt;/h1&gt;"; got != want {
+		t.Errorf("response body: got %q want %q", got, want)
+	}
```
kele, line 185: Why not a tabletest?

mihalimara22: Refactored to tabletest in [3b8ef82](url)

safehttp/mux.go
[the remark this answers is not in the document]
mihalimara22: `methodMap` ?

safehttp/mux.go
```diff
@@ -0,0 +1,113 @@
...
+	// HTTP GET request
+	MethodGet = "GET"
+	// HTTP Post request
+	MethodPost = "POST"
+)
+
+// ServeMux is a safe HTTP request multiplexer that wraps http.ServeMux.
+// It matches the URL of each incoming request against a list of registered
+// patterns and calls the handler for the pattern that most closely matches the
+// URL.
```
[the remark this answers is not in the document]
mihalimara22: We used the `net/http` documentation (should we cite it?) and added a last paragraph about combined handlers (tried not to go into implementation details) in [7a93883](url)

safehttp/mux.go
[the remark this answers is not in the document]
mattiasgrenfeldt: I think `methods` or `handlers` is better. Include the `Map` suffix feels unnecessary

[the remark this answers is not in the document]
mihalimara22: I'd rather opt for `handlers` then because the values are handlers not methods

safehttp/mux.go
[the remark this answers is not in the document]
mattiasgrenfeldt: I think this part of the paragraph about multiple methods is unnecessary:
```
These will be combined into one handler per pattern. The framework will then match each incoming request to its underlying handler according to its HTTP method.
```
I think we could remove it.

safehttp/mux.go
```diff
@@ -0,0 +1,144 @@
...
+// "http://www.google.com/".
+//
+// ServeMux also takes care of sanitizing the URL request path and the Host
+// header, stripping the port number and redirecting any request containing . or
+// .. elements or repeated slashes to an equivalent, cleaner URL.
+//
+// Multiple HTTP methods can be served for one pattern and the user is expected
+// to register a handler for each method supported. These will be combined
+// into one handler per pattern. The framework will then match each
+// incoming request to its underlying handler according to its HTTP method.
```
kele, line 67:
```suggestion
// Multiple handlers can be registered for a single pattern, as long as they
// handle different HTTP methods.
```

kele: This is enough.

The latter part of "These will be combined (...)" is something that doesn't help the reader. As an example: this would've been very different if "combining" the handlers would've provided some extra features, i.e. combined handlers would somehow be able to communicate with each other via framework-provided mechanisms. This is not the case, though, and it simply leaks an implementation detail.

safehttp/mux.go
```diff
@@ -0,0 +1,144 @@
...
+type ServeMux struct {
+	mux     *http.ServeMux
+	domains map[string]bool
+	disp    Dispatcher
+
+	// Maps user-provided patterns to combined handlers which encapsulate
+	// multiple handlers, each one associated with an HTTP method.
```
kele, line 74:
```suggestion
	// Maps patterns to handlers supporting multiple HTTP methods.
```

kele: A few reasons for this change:

1. It's more concise.
2. "user-provided" is too detailed (and has the risk of being untrue for some scenarios).

safehttp/mux.go
```diff
@@ -0,0 +1,144 @@
...
+		domains:  d,
+		disp:     dispatcher,
+		handlers: map[string]methodHandler{},
+	}
+}
+
+// Handle registers a handler for the given pattern and method. If another
+// handler is already registered for the same pattern and method, Handle panics.
+func (m *ServeMux) Handle(pattern string, method string, h Handler) {
+	ch, ok := m.handlers[pattern]
```
kele, line 96: nit: please go through this PR and make sure the short names still make sense. `ch` is confusing since it holds a `methodHandler` now.

mattiasgrenfeldt:

PR #36, opened 2020-07-31
HSTS plugin skeleton

mattiasgrenfeldt:
A plugin which automatically applies HSTS to incoming requests. The `Before` function must be run on incoming requests before they are passed to a handler.

### DIFF
--- a/safehttp/plugins/hsts/hsts.go
+++ b/safehttp/plugins/hsts/hsts.go
@@ -18,15 +18,16 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/google/go-safeweb/safehttp"
 )
 
 // Plugin implements automatic HSTS functionality.
 type Plugin struct {
-	// The time, in seconds, that the browser should remember
+	// The time that the browser should remember
 	// that a site is only to be accessed using HTTPS.
-	MaxAge uint64
+	MaxAge time.Duration
 
 	// This field controls the includeSubDomains directive.
 	// When DisableIncludeSubDomains is false, all subdomains
@@ -50,7 +51,7 @@ type Plugin struct {
 
 // NewPlugin creates a new HSTS plugin with safe defaults.
 func NewPlugin() Plugin {
-	return Plugin{MaxAge: 63072000} // two years in seconds
+	return Plugin{MaxAge: 63072000 * time.Second} // two years in seconds
 }
 
 // Before should be executed before the request is sent to the handler.
@@ -58,6 +59,10 @@ func NewPlugin() Plugin {
 // is received the Strict-Transport-Security header is applied to the
 // response.
 func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+	if p.MaxAge < 0 {
+		return w.ServerError(safehttp.StatusInternalServerError, "Internal Server Error")
+	}
+
 	if !p.BehindProxy && r.TLS == nil {
 		u, err := url.Parse(r.URL.String())
 		if err != nil {
@@ -69,7 +74,7 @@ func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) 
 
 	var value strings.Builder
 	value.WriteString("max-age=")
-	value.WriteString(strconv.FormatUint(p.MaxAge, 10))
+	value.WriteString(strconv.FormatInt(int64(p.MaxAge.Seconds()), 10))
 	if !p.DisableIncludeSubDomains {
 		value.WriteString("; includeSubDomains")
 	}
--- a/safehttp/plugins/hsts/hsts.go
+++ b/safehttp/plugins/hsts/hsts.go
@@ -61,6 +61,8 @@ func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) 
 		// TODO(@mattiasgrenfeldt): Replace the response with an actual saferesponse somehow.
 		return w.ServerError(safehttp.StatusInternalServerError, "Internal Server Error")
 	}
+	// TODO: Implement header claiming.
+	h.MarkImmutable("Strict-Transport-Security")
 	return safehttp.Result{}
 }
 
--- a/safehttp/plugins/hsts/hsts.go
+++ b/safehttp/plugins/hsts/hsts.go
@@ -24,6 +24,7 @@ import (
 )
 
 // Plugin implements automatic HSTS functionality.
+// See https://tools.ietf.org/html/rfc6797 for more info.
 type Plugin struct {
 	// MaxAge is the duration that the browser should remember
 	// that a site is only to be accessed using HTTPS. MaxAge
@@ -52,6 +53,10 @@ type Plugin struct {
 }
 
 // NewPlugin creates a new HSTS plugin with safe defaults.
+// These safe defaults are:
+//  - max-age set to two years.
+//  - includeSubDomains is enabled.
+//  - preload is disabled.
 func NewPlugin() Plugin {
 	return Plugin{MaxAge: 63072000 * time.Second} // two years in seconds
 }
--- a/safehttp/plugins/hsts/hsts.go
+++ b/safehttp/plugins/hsts/hsts.go
@@ -31,7 +31,7 @@ type Plugin struct {
 // NewPlugin creates a new HSTS plugin with safe defaults.
 func NewPlugin() Plugin {
 	return Plugin{
-		maxAge:            63072000, // Two years in seconds.
+		maxAge:            63072000, // two years in seconds
 		includeSubDomains: true,
 		preload:           false,
 	}
--- a/safehttp/plugins/hsts/hsts.go
+++ b/safehttp/plugins/hsts/hsts.go
@@ -39,6 +39,12 @@ type Plugin struct {
 	// by all major browsers. See https://hstspreload.org/ for
 	// more info.
 	Preload bool
+
+	// If this server is behind a proxy that terminates HTTPS
+	// traffic then this should be enabled. If this is enabled
+	// then the plugin will always send the Strict-Transport-Security
+	// header and will not redirect HTTP traffic to HTTPS traffic.
+	BehindProxy bool
 }
 
 // NewPlugin creates a new HSTS plugin with safe defaults.
@@ -51,7 +57,7 @@ func NewPlugin() Plugin {
 // is received the Strict-Transport-Security header is applied to the
 // response.
 func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
-	if r.TLS == nil {
+	if !p.BehindProxy && r.TLS == nil {
 		r.URL.Scheme = "https"
 		return w.Redirect(r, r.URL.String(), safehttp.StatusMovedPermanently)
 	}
--- a/safehttp/plugins/hsts/hsts.go
+++ b/safehttp/plugins/hsts/hsts.go
@@ -25,24 +25,26 @@ import (
 
 // Plugin implements automatic HSTS functionality.
 type Plugin struct {
-	// The time that the browser should remember
-	// that a site is only to be accessed using HTTPS.
+	// MaxAge is the duration that the browser should remember
+	// that a site is only to be accessed using HTTPS. MaxAge
+	// must be positive.
 	MaxAge time.Duration
 
-	// This field controls the includeSubDomains directive.
+	// DisableIncludeSubDomains disables the includeSubDomains directive.
 	// When DisableIncludeSubDomains is false, all subdomains
 	// of the domain where this service is hosted will also be added
 	// to the browsers HSTS list.
 	DisableIncludeSubDomains bool
 
-	// This field controls the preload directive.
+	// Preload enables the preload directive.
 	// This should only be enabled if this site should be
 	// added to the browser HSTS preload list, which is supported
 	// by all major browsers. See https://hstspreload.org/ for
 	// more info.
 	Preload bool
 
-	// If this server is behind a proxy that terminates HTTPS
+	// BehindProxy controls how the plugin should behave with regards
+	// to HTTPS. If this server is behind a proxy that terminates HTTPS
 	// traffic then this should be enabled. If this is enabled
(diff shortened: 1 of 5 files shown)


safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,76 @@
...
+	includeSubDomains bool
+	preload           bool
+}
+
+// NewPlugin creates a new HSTS plugin with safe defaults.
+func NewPlugin() Plugin {
+	return Plugin{
+		maxAge:            63072000, // Two years in seconds.
+		includeSubDomains: true,
+		preload:           false,
```
mihalimara22, line 36: Why?

safehttp/response_writer.go
```diff
@@ -58,6 +58,13 @@ func (w *ResponseWriter) ServerError(code StatusCode, resp Response) Result {
 	return Result{}
 }
 
+// Redirect responds with a redirect to the given url.
+// The type of redirect is specified via the status code.
```
mihalimara22, line 62: "whose type is specified by the status code"

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	h := w.Header()
+	if err := h.Set("Strict-Transport-Security", value.String()); err != nil {
+		// TODO(@mattiasgrenfeldt): Replace the response with an actual saferesponse somehow.
+		return w.ServerError(500, "Internal Server Error")
+	}
+	return safehttp.Result{}
+}
+
+// EnablePreload enables the preload directive.
```
mihalimara22, line 66: Please add comments about when the user should call or not call this functions. Here and below.

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,76 @@
...
+type Plugin struct {
+	maxAge            uint64
+	includeSubDomains bool
+	preload           bool
+}
+
+// NewPlugin creates a new HSTS plugin with safe defaults.
+func NewPlugin() Plugin {
+	return Plugin{
+		maxAge:            63072000, // Two years in seconds.
```
mihalimara22, line 34: No need to capitalise the comment, I think

safehttp/plugins/hsts/hsts.go
mattiasgrenfeldt: If `preload` is sent from the server it can be registered in the browser preload list (https://hstspreload.org/). This causes all(/most) browsers to only ever send HTTPS requests to your server. It can take a long time to be removed from this list and if you don't want it or your server isn't prepared for it then that is bad. Therefore it is disabled by default. It is good to transition to having preload enabled eventually, but this can't be a default setting.

safehttp/plugins/hsts/hsts.go
mattiasgrenfeldt: I don't see how having `// two years in seconds.` is any better. I'll keep it as `// Two years in seconds.` to respect capitalization in beginning of sentences.

mihalimara22: I was looking at this: https://golang.org/src/go/doc/example.go and it seems like the standard is no capitalisation or full stop for inline comments :) It's not a big thing anyway

mattiasgrenfeldt: ok, then I will change it :)

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// Before should be executed before the request is sent to the handler.
+// The function redirects HTTP requests to HTTPS. When HTTPS traffic
+// is received the Strict-Transport-Security header is applied to the
+// response.
+func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+	if r.TLS == nil {
+		r.URL.Scheme = "https"
+		return w.Redirect(r, r.URL.String(), 301)
```
empijei, line 47:
```suggestion
		return w.Redirect(r, r.URL.String(), safehttp.StatusPermanentRedirect)
```

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	value := strings.Builder{}
```
empijei, line 50: Nit: when initializing zero values it is preferred to use var.
```suggestion
	var value strings.Builder
```

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,81 @@
...
+	value.WriteString("max-age=" + strconv.FormatUint(p.maxAge, 10))
```
empijei, line 51: concatenating strings onto a WriteString call on a strings.Builder does not really make good use of the builder.

Here you can either use a text/template to format this entire header value onto a strings.Builder or just call WriteString once per string.

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,81 @@
...
+	if p.includeSubDomains {
+		value.WriteString("; includeSubDomains")
+	}
+	if p.preload {
+		value.WriteString("; preload")
+	}
+	h := w.Header()
+	if err := h.Set("Strict-Transport-Security", value.String()); err != nil {
+		// TODO(@mattiasgrenfeldt): Replace the response with an actual saferesponse somehow.
+		return w.ServerError(500, "Internal Server Error")
```
empijei, line 61: please use safehttp.StatusInternalServerError.

I think we might need to clone all the statuses inside the safehttp package. You can do that in a follow-up pr.

I also think that maybe the ServerError method might be ok if it only accepts a status, without a string.

@kele WDYT?

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/google/go-safeweb/safehttp"
+)
+
+// Plugin implements automatic HSTS functionality.
+type Plugin struct {
+	maxAge            uint64
+	includeSubDomains bool
+	preload           bool
+}
```
empijei, line 29: setters and getters do nothing on this type, so does the constructor, I'd suggest creating somth with a valid zero-value. E.g:

```suggestion
type Plugin struct {
	maxAge            uint64
	disableIncludeSubDomains bool
	preload           bool
}
```
And when the plugin runs if maxage is 0 you use a safe default.

Then you can remove constructor, getters and setters and put the opportune comments on the fields of this struct.

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,81 @@
...
+	value := strings.Builder{}
+	value.WriteString("max-age=" + strconv.FormatUint(p.maxAge, 10))
+	if p.includeSubDomains {
+		value.WriteString("; includeSubDomains")
+	}
+	if p.preload {
+		value.WriteString("; preload")
+	}
+	h := w.Header()
+	if err := h.Set("Strict-Transport-Security", value.String()); err != nil {
```
empijei, line 59: We still need to implement the header-claiming technique, so maybe leave a TODO here.

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,81 @@
...
+		// TODO(@mattiasgrenfeldt): Replace the response with an actual saferesponse somehow.
```
empijei, line 60: What is this TODO for?

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,81 @@
...
+		includeSubDomains: true,
+		preload:           false,
+	}
+}
+
+// Before should be executed before the request is sent to the handler.
+// The function redirects HTTP requests to HTTPS. When HTTPS traffic
+// is received the Strict-Transport-Security header is applied to the
+// response.
+func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
```
empijei, line 44: @kele do we want plugins to return `safehttp.Result`?

safehttp/plugins/hsts/hsts.go
mattiasgrenfeldt: Nice, I will do this!

But we might want `maxAge=0` as a legitimate value since it is used to expire and disable HSTS:
```
Should it be necessary to disable Strict Transport Security, setting the max-age to 0 (over a https connection) will immediately expire the Strict-Transport-Security header, allowing access via http.
```
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security

So we can't replace a 0 with a default safe value. How could we deal with this without having a constructor which sets `maxAge`?

safehttp/plugins/hsts/hsts.go
mattiasgrenfeldt: `ServerError` takes a safe response as an argument:
```go
func (w *ResponseWriter) ServerError(code StatusCode, resp Response) Result
```
This TODO was to remind me that the string I'm currently passing will not be a valid safe response and needs to be replaced.

Instead of making a TODO I should probably have made a comment here in the code review to discuss this with you and kele. It feels strange to send a safe response to `ServerError` since the plugins can't create them.

safehttp/plugins/hsts/hsts.go
kele: Right now, yeah.

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,82 @@
...
+	return Plugin{MaxAge: 63072000} // two years in seconds
+}
+
+// Before should be executed before the request is sent to the handler.
+// The function redirects HTTP requests to HTTPS. When HTTPS traffic
+// is received the Strict-Transport-Security header is applied to the
+// response.
+func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+	if !p.BehindProxy && r.TLS == nil {
+		r.URL.Scheme = "https"
```
kele, line 61: I'd not do this. Imagine a scenario:

You've installed two plugins, `foo.Plugin` and `hsts.Plugin`. `foo` runs before `hsts`.

- `foo.Before` runs, then `hsts.Before`
- `hsts.After` runs (since `it has run `Before`)
- `foo.After` also runs, because it has already run it's `Before`
- `foo.After` reads the `IncomingRequest.URL`, because it logs all URLs visited when non-200 status was produced as a response

safehttp/response_writer.go
```diff
@@ -58,6 +58,13 @@ func (w *ResponseWriter) ServerError(code StatusCode, resp Response) Result {
 	return Result{}
 }
 
+// Redirect responds with a redirect, whose type is
+// specified by the status code, to the given url.
```
kele, line 62:
```suggestion
// Redirect responds with a redirect to a given url, using code as the status code.
```

safehttp/plugins/hsts/hsts.go
mattiasgrenfeldt: Ah, I thought that if a plugin interrupts during the `Before` phase then the `After` phase is not run.

mattiasgrenfeldt:
Tests are here

safehttp/plugins/hsts/hsts_test.go
```diff
@@ -0,0 +1,232 @@
...
+		wantHeaders map[string][]string
+		wantBody    string
+	}{
+		{
+			name:       "HTTPS",
+			plugin:     hsts.NewPlugin(),
+			req:        httptest.NewRequest("GET", "https://localhost/", nil),
+			wantStatus: 200,
+			wantHeaders: map[string][]string{
+				"Strict-Transport-Security": {"max-age=63072000; includeSubDomains"},
```
kele, line 89: Leave a comment what the number means.

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/google/go-safeweb/safehttp"
+)
+
+// Plugin implements automatic HSTS functionality.
+type Plugin struct {
+	// The time, in seconds, that the browser should remember
+	// that a site is only to be accessed using HTTPS.
+	MaxAge uint64
+
+	// This field controls the includeSubDomains directive.
```
empijei, line 31: Doc should start with the name of the field it is documenting: "DisableInclude... disables ..."

same for all other fields. It is odd to see all godoc starting with "This field".

safehttp/plugins/hsts/hsts_test.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+}
+
+type responseRecorder struct {
+	header http.Header
+	writer io.Writer
+	status int
+}
+
+func newResponseRecorder(w io.Writer) *responseRecorder {
```
empijei, line 58: Is there any reason why this is not using [`httptest.ResponseRecorder`](https://golang.org/pkg/net/http/httptest/#ResponseRecorder)?

mattiasgrenfeldt: Yes, see #27

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,87 @@
...
+	// If this server is behind a proxy that terminates HTTPS
+	// traffic then this should be enabled. If this is enabled
+	// then the plugin will always send the Strict-Transport-Security
+	// header and will not redirect HTTP traffic to HTTPS traffic.
+	BehindProxy bool
+}
+
+// NewPlugin creates a new HSTS plugin with safe defaults.
+func NewPlugin() Plugin {
+	return Plugin{MaxAge: 63072000} // two years in seconds
```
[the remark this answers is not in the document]
mattiasgrenfeldt: MaxAge is now a `time.Duration`

safehttp/plugins/hsts/hsts.go
```diff
@@ -0,0 +1,92 @@
...
+	Preload bool
+
+	// If this server is behind a proxy that terminates HTTPS
+	// traffic then this should be enabled. If this is enabled
+	// then the plugin will always send the Strict-Transport-Security
+	// header and will not redirect HTTP traffic to HTTPS traffic.
+	BehindProxy bool
+}
+
+// NewPlugin creates a new HSTS plugin with safe defaults.
```
mihalimara22, line 54: You could give more details about what the safe defaults are. :)

mattiasgrenfeldt:

PR #37, opened 2020-08-01
Fixes the gofmt GitHub action according to Michael's tweet ;)

kele:
Apparently, gofmt doesn't set an exit status.

https://twitter.com/zekjur/status/1289463636331646977

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -22,10 +22,6 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
-    - name: Ensure all files were formatted as per gofmt
-      run: |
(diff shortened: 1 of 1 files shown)


kele:
Michael, could you take a look at the whole action file? I don't have that much experience with it, so I've mostly copied stuff from your repos and simplified where possible (likely sacrificing some other features).

stapelberg:
> @stapelberg , could you take a look at the whole action file? I don't have that much experience with it, so I've mostly copied stuff from your repos and simplified where possible (likely sacrificing some other features).

The file looks good to me.

The only thing I’d consider changing is pinning to the Go major version: you currently pin to 1.13. Not sure what your policy is for this repo, but always tracking the current major version (without an explicit pin) seems like a good policy to me to start with.

PR #41, opened 2020-08-04
Changed to use strconv.ParseBool

### LINKED ISSUE #35: Handling boolean form values
Currently, we consider valid booleans to be only  the strings `"true"` and `"false"`, unlike `strconv.ParseBool()` which correctly converts `"1", "t", "T", "true", "TRUE", "True", "0", "f", "F", "false", "FALSE", "False"` into `bool`. We should discuss the implications of this and decide whether we want to use `strconv.ParseBool()` in the `safehttp.Form` implementation.

mattiasgrenfeldt:
Fixes #35

### DIFF
--- a/safehttp/form.go
+++ b/safehttp/form.go
@@ -93,12 +93,12 @@ func (f *Form) Bool(param string, defaultValue bool) bool {
 	if !ok {
 		return defaultValue
 	}
-	b, err := strconv.ParseBool(vals[0])
+	paramVal, err := strconv.ParseBool(vals[0])
 	if err != nil {
 		f.err = err
 		return defaultValue
 	}
-	return b
+	return paramVal
 }
 
 func clearSlice(slicePtr interface{}) error {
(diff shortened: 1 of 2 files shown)


safehttp/form.go
```diff
@@ -93,15 +93,12 @@ func (f *Form) Bool(param string, defaultValue bool) bool {
...
 		return defaultValue
 	}
-	switch vals[0] {
-	case "true":
-		return true
-	case "false":
-		return false
-	default:
-		f.err = fmt.Errorf("values of form parameter %q not a boolean", param)
+	b, err := strconv.ParseBool(vals[0])
```
mihalimara22, line 96: The other functions use `paramVal` here. Idk which is more readable, but I think it makes the code more uniform to have the same name in all the functions.

empijei: let's always use the same way to access params.

kele:
@mihalimara22 @empijei please re-review since you've requested changes. I'd like to start doing clean merges. :)

mattiasgrenfeldt:

PR #45, opened 2020-08-04
Header Claiming

### LINKED ISSUE #44: Support header claiming as a means of expressing ownership of (security-sensitive) headers
Plugins might need to mark a header as immutable and then change it at a later stage. A method called `Claim` will therefore be implemented that sets the header as immutable and returns something that can later change the header. When `Claim` is implemented there will be no need for `MarkImmutable` (for now) and therefore it will be removed.

mattiasgrenfeldt:
Fixes #44

Plugins might need to mark a header as immutable and then change it at a later stage. A method called Claim will therefore be implemented that sets the header as immutable and returns something that can later change the header. When Claim is implemented there will be no need for MarkImmutable (for now) and therefore it will be removed.

### DIFF
--- a/safehttp/header.go
+++ b/safehttp/header.go
@@ -24,31 +24,37 @@ import (
 // The keys will be in canonical form, as returned by
 // textproto.CanonicalMIMEHeaderKey.
 type Header struct {
-	wrapped   http.Header
-	immutable map[string]bool
+	wrapped http.Header
+	claimed map[string]bool
 }
 
 func newHeader(h http.Header) Header {
-	return Header{wrapped: h, immutable: map[string]bool{}}
+	return Header{wrapped: h, claimed: map[string]bool{}}
 }
 
-// MarkImmutable marks the header with the given name as immutable.
-// The name is first canonicalized using textproto.CanonicalMIMEHeaderKey.
-// This header is now read-only. If the header previously was
-// undefined, then it will forever be undefined. Other methods in
+// Claim claims the header with the given name and returns a function
+// which can be used to set the header. The name is first canonicalized
+// using textproto.CanonicalMIMEHeaderKey. Other methods in
 // the struct can't write to, change or delete the header with this
-// name. These methods will instead fail when applied on an immutable
-// header.
-func (h Header) MarkImmutable(name string) {
+// name. These methods will instead fail when applied on an claimed
+// header. The only way to modify the header is to use the returned
+// function. The Set-Cookie header can't be claimed.
+func (h Header) Claim(name string) (set func([]string), err error) {
 	name = textproto.CanonicalMIMEHeaderKey(name)
-	h.immutable[name] = true
+	if err := h.writableHeader(name); err != nil {
+		return nil, err
+	}
+	h.claimed[name] = true
+	return func(v []string) {
+		h.wrapped[name] = v
+	}, nil
 }
 
 // Set sets the header with the given name to the given value.
 // The name is first canonicalized using textproto.CanonicalMIMEHeaderKey.
 // This method first removes all other values associated with this
 // header before setting the new value. Returns an error when
-// applied on immutable headers or on the Set-Cookie header.
+// applied on claimed headers or on the Set-Cookie header.
 func (h Header) Set(name, value string) error {
 	name = textproto.CanonicalMIMEHeaderKey(name)
 	if err := h.writableHeader(name); err != nil {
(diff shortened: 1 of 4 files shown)


safehttp/header_test.go
```diff
@@ -354,3 +368,49 @@ func TestValuesEmptyHeader(t *testing.T) {
...
+
+func TestClaim(t *testing.T) {
+	h := newHeader(http.Header{})
+	set, err := h.Claim("Foo-Key")
+	if err != nil {
+		t.Fatalf(`_, err := h.Claim("Foo-Key") got: %v want: nil`, err)
+	}
+	set([]string{"Bar-Value", "Pizza-Value"})
+	if diff := cmp.Diff([]string{"Bar-Value", "Pizza-Value"}, h.Values("Foo-Key")); diff != "" {
+		t.Errorf("h.Values(\"Foo-Key\") mismatch (-want +got):\n%s", diff)
```
mihalimara22, line 380: You could use backticks (```) so you don't need to escape quotes

safehttp/header.go
```diff
@@ -24,31 +24,37 @@ import (
...
 type Header struct {
-	wrapped   http.Header
-	immutable map[string]bool
+	wrapped http.Header
+	claimed map[string]bool
 }
 
 func newHeader(h http.Header) Header {
-	return Header{wrapped: h, immutable: map[string]bool{}}
+	return Header{wrapped: h, claimed: map[string]bool{}}
```
mihalimara22, line 32: nit: You could reformat this so each field is on one line

safehttp/header_test.go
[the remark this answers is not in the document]
mattiasgrenfeldt: Not in this case since I also need to escape the `\n`. I think it looks better to have the string on one line instead of two so I'll keep the `"`

safehttp/header_test.go
```diff
@@ -354,3 +368,49 @@ func TestValuesEmptyHeader(t *testing.T) {
...
+	set([]string{"Bar-Value", "Pizza-Value"})
+	if diff := cmp.Diff([]string{"Bar-Value", "Pizza-Value"}, h.Values("fOo-kEy")); diff != "" {
+		t.Errorf("h.Values(\"fOo-kEy\") mismatch (-want +got):\n%s", diff)
+	}
+}
+
+func TestClaimSetCookie(t *testing.T) {
+	h := newHeader(http.Header{})
+	if _, err := h.Claim("Set-Cookie"); err == nil {
+		t.Error(`_, err := h.Claim("Set-Cookie") got: nil want: error`)
```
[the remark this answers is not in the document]
mattiasgrenfeldt: How should I replace it? Like this: ```t.Error(`h.Claim("Set-Cookie") ...`)``` ?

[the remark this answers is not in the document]
mattiasgrenfeldt: Ah, I see, I'd already done like this ```t.Error(`h.Claim("Set-Cookie") got: nil want: error`)``` before. Sorry! Fixed!

mattiasgrenfeldt:
Apparently the CI is based on what is in master and not only what is in this branch. It failed to build because the HSTS plugin had been merged to master and it uses `MarkImmutable`, which has been removed in this branch. But why did the CI mix in master now? Very strange.

safehttp/header_test.go
```diff
@@ -81,9 +83,11 @@ func TestSetImmutable(t *testing.T) {
...
 
-func TestSetEmptyImmutable(t *testing.T) {
+func TestSetEmptyClaimed(t *testing.T) {
 	h := newHeader(http.Header{})
-	h.MarkImmutable("Foo-Key")
+	if _, err := h.Claim("Foo-Key"); err != nil {
+		t.Errorf(`h.Claim("Foo-Key") got: %v want: nil`, err)
+	}
 	if err := h.Set("Foo-Key", "Bar-Value"); err == nil {
 		t.Error(`h.Set("Foo-Key", "Bar-Value") got: nil want: error`)
```
mihalimara22, line 92: Idk which is the better option, but I wondering if it wouldn't be more helpful to output more info about what sort of error occurred.

mattiasgrenfeldt: When this line runs, then no error has occurred. I could include more information about which error should have occurred, but if we in the future change the error message that is returned somehow then you would have to update this line as well and that is difficult to remember.

PR #48, opened 2020-08-05
 XSRF plugin skeleton

### LINKED ISSUE #47: XSRF Plugin
One of the plugins that are part of the `go-safeweb` will need to provide XSRF protection. This will also include a Fetch Metadata policy in case the browser used has Fetch Metadata support.

mihalimara22:
Fixes #47 

Created the preliminary version of the XSRF Plugin that generates and validates tokens.
Brainstorming draft doc: [url](https://docs.google.com/document/d/1Il7lvT8QZbcQaFUZnv8idziQ46htwByrLHrxbP1b3AA/edit#heading=h.mkyw33dr2av0).


safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	token := f.String(TokenKey, "")
+	if f.Err() != nil || token == "" {
+		return safehttp.Status403Forbidden, false
+	}
+	ok := xsrftoken.Valid(token, p.appKey, userID, ir.Header.Get("Host"))
+	if !ok {
+		return safehttp.Status403Forbidden, false
+	}
+	return 0, true
```
mattiasgrenfeldt, line 116: Shouldn't this be `return safehttp.StatusOK, true`?
Same in `applyPolicy`

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,172 @@
...
+// their own policy
+func (p *Plugin) applyPolicy(ir *safehttp.IncomingRequest) (safehttp.StatusCode, bool) {
+	switch ir.Header.Get("Sec-Fetch-Site") {
+	case "":
+		// Fetch Metadata not supported by the browser
+		return 0, true
+	case "same-origin", "same-site", "none":
+		return 0, true
+	case "cross-origin":
+		if p.allowedCORS[ir.Header.Get("Host")] {
```
mattiasgrenfeldt, line 134: `ir.Header.Get("Host")` doesn't contain the Host header. They are promoted to the `http.Request.Host` field and are removed from the `http.Request.Header` map. You therefore need to implement a getter like you did for `GetMethod` or promote the field to `safehttp.IncomingRequest.Host` when creating an `IncomingRequest` from a `http.Request`

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,172 @@
...
+			return 0, true
+		}
+		return safehttp.Status403Forbidden, false
+	}
+
+	switch ir.Header.Get("Sec-Fetch-Mode") {
+	case "navigate":
```
mattiasgrenfeldt, line 141: This should be an if-statement and not a switch statement

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,172 @@
...
+	f, err := ir.PostForm()
+	if err != nil {
+		return safehttp.Status400BadRequest, false
+	}
+	token := f.String(TokenKey, "")
+	if f.Err() != nil || token == "" {
+		return safehttp.Status403Forbidden, false
+	}
+	ok := xsrftoken.Valid(token, p.appKey, userID, ir.Header.Get("Host"))
+	if !ok {
```
mattiasgrenfeldt, line 113: These lines can be combined:
```go
if ok := xsrftoken.Valid(...); !ok {
```

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,172 @@
...
+		if p.allowedCORS[ir.Header.Get("Host")] {
+			return 0, true
+		}
+		return safehttp.Status403Forbidden, false
+	}
+
+	switch ir.Header.Get("Sec-Fetch-Mode") {
+	case "navigate":
+		dest := ir.Header.Get("Sec-Fetch-Dest")
+		if dest == "object" || dest == "embed" {
```
mattiasgrenfeldt, line 143: These two lines can be combined:
```go
if dest := ...; dest == ... || dest == ... { 
```
There are also some `if`s in `Before` where this can be applied.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,172 @@
...
+			return safehttp.Status403Forbidden, false
+		}
+		if ir.GetMethod() == "GET" || ir.GetMethod() == "HEAD" {
```
mattiasgrenfeldt, line 146: You could call `GetMethod` only once:
```go
if m := ir.GetMethod(); m == ... || m == ... {
```

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// Plugin implements XSRF protection.
+type Plugin struct {
+	// TODO(@mihalimara22): decide whether  we want to allow the user to set a
+	// Fetch Metadata policy and how the report mode is going to be handled
+	appKey       string
+	s            StorageService
+	fmPolicyMode string
+	allowedCORS  map[string]bool
```
mattiasgrenfeldt, line 53: I think this should be left out until we implement the CORS plugin

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return "", fmt.Errorf("token generation failed: %v", err)
+	}
+	token := xsrftoken.Generate(p.appKey, userID, host)
+	return token, nil
+}
+
+// validateToken validates the XSRF token. This should be present in all
+// requests as the value of form parameter xsrf-token.
+func (p *Plugin) validateToken(ir *safehttp.IncomingRequest) (safehttp.StatusCode, bool) {
```
mattiasgrenfeldt, line 98: This could be simplified to only return a status code. 200 if all is good. The same goes for `applyPolicy`

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+	}
+	return safehttp.Status403Forbidden, false
+}
+
+// Before should be executed before directing the request to the handler. The
+// function applies checks to the Incoming Request to ensure this is not part
+// of a Cross-Site Request Forgery. These ensure the request complies to the
+// Fetch Metadata policy provided and contain a XSRF token.
+func (p *Plugin) Before(rw safehttp.ResponseWriter, ir *safehttp.IncomingRequest) safehttp.Result {
```
mattiasgrenfeldt, line 157: nit: rename `rw` with only `w` and `ir` with `r`.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// GenerateToken generates a cryptographically safe XSRF token using the user ID
+// and the path.
+func (p *Plugin) GenerateToken(host string) (string, error) {
+	userID, err := p.s.GetUserID()
+	if err != nil {
+		return "", fmt.Errorf("token generation failed: %v", err)
+	}
+	token := xsrftoken.Generate(p.appKey, userID, host)
```
mattiasgrenfeldt, line 92: I think that the `actionID` should be more specific than only the host. I think it should be a combination of host + path + method or at least host + path. That feels more secure.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,172 @@
...
+	TokenKey = "xsrf-token"
+)
+
+// StorageService is an interface the framework users need to implement. This
+// will contain information about users of the web application, including their
+// IDs, needed in generating the XSRF token.
+type StorageService interface {
+	// TODO(@mihalimara22): add the parameters that the storage service needs in
+	// order to determine the user ID
+	GetUserID() (string, error)
```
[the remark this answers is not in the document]
mihalimara22: I think the storage service should have access to the underlying http request, or at least some of it. It's not clear to me what we should actually pass as an argument right now

safehttp/plugins/xsrf/xsrf.go
mihalimara22: I also considered that option but wasn't sure. The thing is, if validation succeeds we will move on to the handler or another plugin so there is no way of knowing the status code is going to be 200. Therefore, I thought it's a better option (for now) not to return a status code unless something wrong occurs and we have to return immediately. But idk what is best.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,172 @@
...
+		s:            s,
+		fmPolicyMode: FMEnforceMode,
+		allowedCORS:  ad,
+	}
+}
+
+// SetFetchMetadataPolicyMode allows setting the Fetch Metadata policy, enforced
+// by default to one of the other options. Setting it to report will allow
+// requests that violate the policy to pass but will log violation. Setting it
+// to disable
```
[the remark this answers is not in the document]
mihalimara22: Will resolve this comments since Fetch Metadata functionality will not be supported for now.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,106 @@
...
+	// order to determine the user ID
+	GetUserID() (string, error)
+}
+
+// Plugin implements XSRF protection.
+//
+// TODO(@mihalimara22): Add Fetch Metadata support
+type Plugin struct {
+	//appKey uniquely identifies each registered server and should have high
+	//entropy as it will be used in the xsrf token generation algorithm.
```
mattiasgrenfeldt, line 45: Move this comment to the godoc comment of `NewPlugin`. Here it won't be visible to the user.

safehttp/incoming_request.go
```diff
@@ -41,6 +41,21 @@ func newIncomingRequest(req *http.Request) *IncomingRequest {
 	}
 }
 
+// GetMethod TODO
+func (r *IncomingRequest) GetMethod() string {
+	return r.req.Method
+}
```
mattiasgrenfeldt, line 47: This is not used anymore so it can be removed.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	return 0, true
+}
+
+// Before should be executed before directing the request to the handler. The
+// function applies checks to the Incoming Request to ensure this is not part
+// of a Cross-Site Request Forgery.
+func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+	if status, ok := p.validateToken(r); !ok {
+		log.Printf("here %d", status)
```
mattiasgrenfeldt, line 102: Remove debug printing

safehttp/plugins/xsrf/xsrf.go
mattiasgrenfeldt: I see your point about not returning `200`. I would suggest that `0` is returned on success and that the `bool` is removed, it feels unnecessary.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,106 @@
...
+		s:      s,
+	}
+}
+
+// GenerateToken generates a cryptographically safe XSRF token per user, using
+// their ID and the request host and path.
+func (p *Plugin) GenerateToken(host string, path string) (string, error) {
+	userID, err := p.s.GetUserID()
+	if err != nil {
+		return "", fmt.Errorf("token generation failed: %v", err)
```
mattiasgrenfeldt, line 65: Just return the `err` as it is.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	token := xsrftoken.Generate(p.appKey, userID, host+path)
+	return token, nil
```
mattiasgrenfeldt, line 68: Nit: `return xsrftoken.Generate(p.appKey, userID, host+path), nil`
Not sure which is more readable though

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,106 @@
...
+// validateToken validates the XSRF token. This should be present in all
+// requests as the value of form parameter xsrf-token.
+func (p *Plugin) validateToken(r *safehttp.IncomingRequest) (safehttp.StatusCode, bool) {
+	userID, err := p.s.GetUserID()
+	if err != nil {
+		return safehttp.StatusUnauthorized, false
+	}
+	f, err := r.PostForm()
+	if err != nil {
+		mf, err := r.MultipartForm(0)
```
mattiasgrenfeldt, line 80: Add TODO about changing the 0 to something better in the future.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,106 @@
...
+			return safehttp.StatusBadRequest, false
+		}
+		f = &mf.Form
+	}
+	tok := f.String(TokenKey, "")
+	if f.Err() != nil || tok == "" {
+		return safehttp.StatusForbidden, false
+	}
+	ok := xsrftoken.Valid(tok, p.appKey, userID, r.GetHost()+r.GetPath())
+	if !ok {
```
mattiasgrenfeldt, line 91: Don't forget to combine these lines:
```go
if ok := xsrftoken.Valid(...); !ok {
```

safehttp/plugins/xsrf_test/xsrf_test.go
```diff
@@ -0,0 +1,129 @@
...
+	r.status = statusCode
+}
+
+func (r *responseRecorder) Write(data []byte) (int, error) {
+	return r.writer.Write(data)
+}
+
+type testStorageService struct{}
+
+func (s testStorageService) GetUserID() (string, error) {
```
mattiasgrenfeldt, line 78: Could be `func (testStorageService) GetUserID() (string, error)`

safehttp/plugins/xsrf_test/xsrf_test.go
```diff
@@ -0,0 +1,129 @@
...
+		b := strings.NewReader(xsrf.TokenKey + "=" + tok)
+		req := httptest.NewRequest("POST", test.target, b)
+		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+		m := safehttp.NewMachinery(p.Before, &testDispatcher{})
+		rec := newResponseRecorder(&strings.Builder{})
+		m.HandleRequest(rec, req)
+		if want := test.wantStatus; rec.status != want {
+			t.Errorf("response status: got %v, want %v", rec.status, want)
+		}
+
```
mattiasgrenfeldt, line 127: Remove empty line

safehttp/plugins/xsrf_test/xsrf_test.go
```diff
@@ -0,0 +1,129 @@
...
+		if err != nil {
+			t.Fatalf("p.GenerateToken: got %v, want nil", err)
+		}
+		b := strings.NewReader(xsrf.TokenKey + "=" + tok)
+		req := httptest.NewRequest("POST", test.target, b)
+		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+		m := safehttp.NewMachinery(p.Before, &testDispatcher{})
+		rec := newResponseRecorder(&strings.Builder{})
+		m.HandleRequest(rec, req)
+		if want := test.wantStatus; rec.status != want {
```
mattiasgrenfeldt, line 124: `want` is unnecessary here. Just use `test.wantStatus`

safehttp/plugins/xsrf_test/xsrf_test.go
```diff
@@ -0,0 +1,129 @@
...
+			t.Errorf("response status: got %v, want %v", rec.status, want)
+		}
+
+	}
+}
```
mattiasgrenfeldt, line 129: Add test: `Invalid userid in token generation`
Also add tests using multipart requests.

safehttp/incoming_request.go
```diff
@@ -41,6 +41,21 @@ func newIncomingRequest(req *http.Request) *IncomingRequest {
 	}
 }
 
+// GetMethod TODO
+func (r *IncomingRequest) GetMethod() string {
```
kele, line 45: Drop the "Get" prefix, here and everywhere else: https://golang.org/doc/effective_go.html#Getters.

safehttp/incoming_request.go
```diff
@@ -41,6 +41,21 @@ func newIncomingRequest(req *http.Request) *IncomingRequest {
 	}
 }
 
+// GetMethod TODO
```
kele, line 44: Add the comments to functions you are implementing in this PR. Otherwise I don't know what I'm reviewing.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,106 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package xsrf
+
+import (
+	"fmt"
+	"github.com/google/go-safeweb/safehttp"
+	"log"
+	// TODO(@empijei, @kele, @mattiasgrenfeldt, @mihalimara22): decide whether we want to depend on this package or reimplement thefunctionality
```
kele, line 21: nit: please break the line

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,106 @@
...
+	"golang.org/x/net/xsrftoken"
+)
+
+const (
+	// TokenKey is the form key used when sending the token as part of POST
+	// request
```
kele, line 27: nit: dot at the end

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,106 @@
...
+	// TODO(@mihalimara22): add the parameters that the storage service needs in
+	// order to determine the user ID
+	GetUserID() (string, error)
+}
+
+// Plugin implements XSRF protection.
+//
+// TODO(@mihalimara22): Add Fetch Metadata support
+type Plugin struct {
+	//appKey uniquely identifies each registered server and should have high
```
kele, line 44: "service", not "server", right?

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,106 @@
...
+	//entropy as it will be used in the xsrf token generation algorithm.
+	appKey string
+	s      StorageService
```
kele, line 47: `storage`

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,106 @@
...
+	"golang.org/x/net/xsrftoken"
+)
+
+const (
+	// TokenKey is the form key used when sending the token as part of POST
+	// request
+	TokenKey = "xsrf-token"
+)
+
+// StorageService is an interface the framework users need to implement. This
```
kele, line 31: Drop the "is an interface the framework users need to implement" -- it doesn't add much value to the comment.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,106 @@
...
+// will contain information about users of the web application, including their
+// IDs, needed in generating the XSRF token.
+type StorageService interface {
+	// TODO(@mihalimara22): add the parameters that the storage service needs in
+	// order to determine the user ID
+	GetUserID() (string, error)
```
kele, line 37: Add a godoc comment for `GetUserID` - how else users will know how to implement it?

safehttp/plugins/xsrf/xsrf.go
kele, line 65: When your function `Foo()` returns an error, it should include relevant information that the caller of `Foo()` doesn't have.

For instance, if I am calling `GenerateToken()` and I get an error `token generation failed: %v`, the only bit I didn't know as a caller is the `%v`. Instead, I would've way preferred to know which part failed (as a caller, I can't know that), i.e.: `couldn't retrieve user ID'.

safehttp/plugins/xsrf_test/xsrf_test.go
mihalimara22: I added a TODO for this, will do it when we commit the testing API as I need the `safehttp.newIncomingRequest` to be exported.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -40,19 +40,19 @@ type StorageService interface {
...
 	storage StorageService
 }
 
-// NewPlugin creates a new XSRF plugin.
+// NewPlugin creates a new XSRF plugin.It requires an application key and a
+// storage service. The appKey uniquely identifies each registered service and
+// should have high entropy. The storage service supports retrieving ID's of the
+// application's users. Both the appKey and user ID are used in the XSRF
+// token generation algorithm.
+//
```
mattiasgrenfeldt: Nit: add a space after "plugin." `... a new XSRF plugin._It requires ...`
Nit: remove the empty comment line

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,105 @@
...
+	// request.
+	TokenKey = "xsrf-token"
+)
+
+// StorageService  contains information about the users of the web application,
+// including their IDs, needed in generating the XSRF token.
+type StorageService interface {
+	// GetUserID returns the ID of the user making the request based on the
+	// incoming request. If an error occurs, it returns it together with an
+	// empty string.
```
kele, line 37:
```suggestion
	// GetUserID returns the ID of the user making the request.
```

kele: This would be enough. The rest is clear.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,105 @@
...
+
+const (
+	// TokenKey is the form key used when sending the token as part of POST
+	// request.
+	TokenKey = "xsrf-token"
+)
+
+// StorageService  contains information about the users of the web application,
+// including their IDs, needed in generating the XSRF token.
+type StorageService interface {
```
kele, line 34: Why is this called "StorageService" if it's purpose is to map requests to user IDs?

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		mf, err := r.MultipartForm(32 << 20)
+		if err != nil {
+			return safehttp.StatusBadRequest
+		}
+		f = &mf.Form
+	}
+	tok := f.String(TokenKey, "")
+	if f.Err() != nil || tok == "" {
+		return safehttp.StatusForbidden
```
kele, line 89: I'd do StatusUnauthorized here too.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	if ok := xsrftoken.Valid(tok, p.appKey, userID, r.Host()+r.Path()); !ok {
+		return safehttp.StatusForbidden
+	}
+	return 0
```
kele, line 94: What does this 0 mean here?

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			return safehttp.StatusBadRequest
+		}
+		f = &mf.Form
+	}
+	tok := f.String(TokenKey, "")
+	if f.Err() != nil || tok == "" {
+		return safehttp.StatusForbidden
+	}
+	if ok := xsrftoken.Valid(tok, p.appKey, userID, r.Host()+r.Path()); !ok {
```
kele, line 91: There are two places where you create a (appKey, userID, host+path) tuple. Could you unify it somewhere?

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,105 @@
...
+
+// StorageService  contains information about the users of the web application,
+// including their IDs, needed in generating the XSRF token.
+type StorageService interface {
+	// GetUserID returns the ID of the user making the request based on the
+	// incoming request. If an error occurs, it returns it together with an
+	// empty string.
+	// TODO(@mihalimara22): add a *safehttp.IncomingRequest as a parameter to
+	// this function once the method for this is exported.
+	GetUserID() (string, error)
```
kele, line 40: I don't get this TODO. What's missing?

safehttp/plugins/xsrf/xsrf.go
mihalimara22: Isn't 403 more descriptive for the situation here as we know the identity of the client and refuse to give them access (because token validation failed)? 401 is returned when the client identity is unknown, but in this case it is (otherwise `GetUserID` would've failed).

safehttp/plugins/xsrf/xsrf.go
mihalimara22: We need to pass a parameter to the GetUserID function to determine the user. Probably the solution would be to pass a `safehttp.IncomingRequest`, WDYT?

If we choose to pass a `sfehttp.IncomingRequest`, I can't test it until we submit the PR where `safehttp.NewIncomingRequest` is exported because I can't create safe requests. That's why I left a TODO here and in the test file.

safehttp/plugins/xsrf/xsrf.go
mihalimara22: Here, I chose to only return a valid `safehttp.StatusCode` if an error occurs and 0 (instead of 200, otherwise). This is because, if validation succeeds, we will move on to some handler or another plugin so there is no way of knowing whether the status code is actually going to be 200. WDYT?

safehttp/plugins/xsrf/xsrf.go
mihalimara22: Renamed to `UserIDStorage`.

safehttp/plugins/xsrf/xsrf.go
mihalimara22: but one is executed by the user and one internally, I don't see a way of doing so :-?

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		mf, err := r.MultipartForm(32 << 20)
+		if err != nil {
+			return safehttp.StatusBadRequest
+		}
+		f = &mf.Form
+	}
+	tok := f.String(TokenKey, "")
+	if f.Err() != nil || tok == "" {
+		return safehttp.StatusUnauthorized
```
empijei, line 87: maybe log this situation.

safehttp/plugins/xsrf/xsrf.go
empijei: Honestly I would inline this function inside Before and not return if stuff goes right. This function is only called there anyways and that is a 4 lines func.

safehttp/plugins/xsrf/xsrf.go
empijei: +1 on passing the request as parameter

safehttp/plugins/xsrf/xsrf.go
mihalimara22: That would be a way better option actually, thanks :)

safehttp/plugins/xsrf/xsrf.go
kele: Wait, why do you need `safehttp.NewIncomingRequest`?

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,103 @@
...
+// storage service. The appKey uniquely identifies each registered service and
+// should have high entropy. The storage service supports retrieving ID's of the
+// application's users. Both the appKey and user ID are used in the XSRF
+// token generation algorithm.
+func NewPlugin(appKey string, s UserIDStorage) *Plugin {
+	return &Plugin{
+		appKey:  appKey,
+		storage: s,
+	}
+}
```
kele, line 58: Get rid of the constructor and make the AppKey and Storage public.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,103 @@
...
+
+// GenerateToken generates a cryptographically safe XSRF token per user, using
+// their ID and the request host and path.
+func (p *Plugin) GenerateToken(host string, path string) (string, error) {
+	userID, err := p.storage.GetUserID()
```
kele, line 63: As it is, this doesn't make much sense since `GetUserID` has no parameters.

safehttp/plugins/xsrf/xsrf.go
kele: WDYM: one is executed by the user?

safehttp/plugins/xsrf/xsrf.go
kele: I'm not sure I get what the code actually does and how does this resolve the thread.

safehttp/plugins/xsrf/xsrf.go
mihalimara22: The `userIDStorage` needs some information to determine what `userID` to return.

safehttp/plugins/xsrf/xsrf.go
mihalimara22: That was my misunderstanding, sorry. Anyway, we won't have this problem as we will only validate XSRF tokens in this PR

safehttp/plugins/xsrf/xsrf.go
mihalimara22: We will probably pass a `*safehttp.IncomingRequest`.

safehttp/plugins/xsrf/xsrf.go
mihalimara22: with `log` or have a logging service passed by the user?

safehttp/plugins/xsrf/xsrf.go
mihalimara22: I applied the change, but since the `AppKey` is used in the token generation algorithm, wouldn't it be a better idea to make sure the plugin is created properly? Or should I check the AppKey field was set in the `Before` function and panic if that's not the case?

safehttp/plugins/xsrf/xsrf.go
mihalimara22: It checks a `userID` exists for the incoming request, checks a `xsrf-token` is present in the request body and validates it.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,72 @@
...
+	// TokenKey is the form key used when sending the token as part of POST
+	// request.
+	TokenKey = "xsrf-token"
+)
+
+// UserIDStorage stores the web application users' IDs,
+// needed in generating the XSRF token.
+type UserIDStorage interface {
+	GetUserID(*safehttp.IncomingRequest) (string, error)
+}
```
kele, line 32: `UserIDStorage` -> `UserIdentifier`. There doesn't have to be any storage, right?

`GetUserID` -> `UserID(*safehttp.IncomingRequest) (string, error)` -- in Go we avoid words like `Get` -- they don't add much value to the name

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Plugin implements XSRF protection. It requires an application key and a
+// storage service. The appKey uniquely identifies each registered service and
+// should have high entropy. The storage service supports retrieving ID's of the
+// application's users. Both the appKey and user ID are used in the XSRF
+// token generation algorithm.
+type Plugin struct {
```
kele, line 39: This is an interceptor in the updated model, right?

PR #50, opened 2020-08-06
Removed Response from ServerError

### LINKED ISSUE #40: Remove the Response argument from ResponseWriter.ServerError until we come up with how to support default error responses
`safehttp.ResponseWriter.ServerError` currently has the following signature:
```go
func (w *ResponseWriter) ServerError(code StatusCode, resp Response) Result
```
It takes a `safehttp.Response` as an argument, but there might be places where we want to respond with a server error without having the ability to create safe responses. For example if a plugin needs to interrupt the handling of an incoming request. We therefore need to reconsider the signature of `ServerError` so that it doesn't take a `Response` or so that it is possible to optionally call without a `Response`.

mattiasgrenfeldt:
Fixes #40 

`safehttp.ResponseWriter.ServerError` currently has the following signature:
```go
func (w *ResponseWriter) ServerError(code StatusCode, resp Response) Result
```
It takes a `safehttp.Response` as an argument, but there might be places where we want to respond with a server error without having the ability to create safe responses. For example if a plugin needs to interrupt the handling of an incoming request. We have therefore decided that **for now** `Response` should be removed from the `ServerError` signature.

### DIFF
--- a/safehttp/plugins/hsts/hsts.go
+++ b/safehttp/plugins/hsts/hsts.go
@@ -67,13 +67,13 @@ func NewPlugin() Plugin {
 // response.
 func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
 	if p.MaxAge < 0 {
-		return w.ServerError(safehttp.StatusInternalServerError, "Internal Server Error")
+		return w.ServerError(safehttp.StatusInternalServerError)
 	}
 
 	if !p.BehindProxy && r.TLS == nil {
 		u, err := url.Parse(r.URL.String())
 		if err != nil {
-			return w.ServerError(safehttp.StatusInternalServerError, "Internal Server Error")
+			return w.ServerError(safehttp.StatusInternalServerError)
 		}
 		u.Scheme = "https"
 		return w.Redirect(r, u.String(), safehttp.StatusMovedPermanently)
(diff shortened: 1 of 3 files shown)


safehttp/plugins/hsts/hsts.go
```diff
@@ -91,7 +91,7 @@ func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest)
 	h := w.Header()
 	if err := h.Set("Strict-Transport-Security", value.String()); err != nil {
 		// TODO(@mattiasgrenfeldt): Replace the response with an actual saferesponse somehow.
```
mihalimara22, line 93: nit: Remove the TODO

safehttp/response_writer.go
```diff
@@ -54,7 +54,11 @@ func (w *ResponseWriter) WriteTemplate(t Template, data interface{}) Result {
 }
 
 // ServerError TODO
-func (w *ResponseWriter) ServerError(code StatusCode, resp Response) Result {
+func (w *ResponseWriter) ServerError(code StatusCode) Result {
+	if code < 500 || code >= 600 {
```
mihalimara22, line 58: Should we return an empty safe Result here? The user of the framework basically called the wrong method.

safehttp/plugins/hsts/hsts.go
mihalimara22, line 93: Isn't this done?

safehttp/response_writer.go
mattiasgrenfeldt: I agree we should somehow signal to the user that they used the wrong function. I guess the options are:
- Return an error
- Embed error in Result somehow
- panic
What do you think? @mihalimara22 @kele @empijei

safehttp/plugins/hsts/hsts.go
mattiasgrenfeldt: Missed that one, thanks! :)

safehttp/response_writer.go
kele: Unclear. Add a TODO for now.

mattiasgrenfeldt:

PR #55, opened 2020-08-07
Introduced the safehttptest to aid end-to-end testing

### LINKED ISSUE #46: Make plugins testing easier by creating a helper package (similar to net/http/httptest)
Provide a package similar to `net/http/httptest` with helper functions that create data structures we'll use when testing the plugins.

mattiasgrenfeldt:
Fixes #46 

We implemented a test helper package to use when testing plugins. It uses the `httptest` package internally.

### DIFF
--- a/safehttp/incoming_request.go
+++ b/safehttp/incoming_request.go
@@ -32,7 +32,9 @@ type IncomingRequest struct {
 	URL       *url.URL
 }
 
-func newIncomingRequest(req *http.Request) *IncomingRequest {
+// NewIncomingRequest creates an safehttp.IncomingRequest
+// from an http.Request.
+func NewIncomingRequest(req *http.Request) *IncomingRequest {
 	return &IncomingRequest{
 		req:    req,
 		Header: newHeader(req.Header),
--- a/safehttp/incoming_request.go
+++ b/safehttp/incoming_request.go
@@ -32,7 +32,7 @@ type IncomingRequest struct {
 	URL       *url.URL
 }
 
-// NewIncomingRequest creates an safehttp.IncomingRequest
+// NewIncomingRequest creates an IncomingRequest
 // from an http.Request.
 func NewIncomingRequest(req *http.Request) *IncomingRequest {
 	return &IncomingRequest{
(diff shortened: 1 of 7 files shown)


safehttp/incoming_request.go
```diff
@@ -32,7 +32,9 @@ type IncomingRequest struct {
 	URL       *url.URL
 }
 
-func newIncomingRequest(req *http.Request) *IncomingRequest {
+// NewIncomingRequest creates an safehttp.IncomingRequest
```
bsiegert, line 35: nit: creates an IncomingRequest

Don't prefix with the package name if the symbol is in the same package.

safehttp/response_writer.go
```diff
@@ -29,7 +29,9 @@ type ResponseWriter struct {
 	header Header
 }
 
-func newResponseWriter(d Dispatcher, rw http.ResponseWriter) ResponseWriter {
+// NewResponseWriter creates a safehttp.ResponseWriter from
+// a safehttp.Dispatcher and a http.ResponseWriter.
```
bsiegert, line 33: nit: same as above

safehttp/safehttptest/recorder.go
```diff
@@ -0,0 +1,119 @@
...
+		panic("not a safe response type")
+	}
+}
+
+// ResponseRecorder encapsulates a safehttp.ResponseWriter that records
+// mutations for later inspection in tests. The safehttp.ResponseWriter
+// should be passed as part of the handler function in tests.
+type ResponseRecorder struct {
+	safehttp.ResponseWriter
+	rw *responseRecorder
```
bsiegert, line 53: May I ask why there is both a ResponseRecorder and a lowercase responseRecorder? Can't you just merge the latter into the former?

mihalimara22: The lowercase responseRecorder is an implementation of the `http.ResponseWriter` interface that we do not surface to the user. Uppercase ResponseRecorder is an implementation of the `safehttp.ResponseWriter` that is going to be used for testing. I renamed responseRecorder to responseWriter and added a comment.

mattiasgrenfeldt:
Rebased on top of master, ready to merge @kele

PR #58, opened 2020-08-07
Implementing safehttp.URL type and a way of accessing URL query parameters

### LINKED ISSUE #33: Add support for Query parameters
The current implementation of `safehttp.Form` type also supports parsing query parameters as a `Form` object and returning any parsing errors that occurred. This follows the implementation of `net/http` and prevents users from calling `URL.Query()`, which silently discards parsing errors. We might want to consider separating the URL from the form type as future work.

mattiasgrenfeldt:
Fixes #33

A new `safehttp.URL` type is implemented. It is very similar to the `net/url.URL` type. 

You can now access URL query parameters by using the `safehttp.URL.Query` method which returns a `safehttp.Form` containing the values.

`net/url.URL` contains a lot more functionality than `safehttp.URL`. The rest of the functionality can be implemented later as needed.

### DIFF
--- a/safehttp/incoming_request.go
+++ b/safehttp/incoming_request.go
@@ -18,7 +18,6 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
-	"net/url"
 	"strings"
 	"sync"
 )
@@ -29,7 +28,7 @@ type IncomingRequest struct {
 	Header    Header
 	parseOnce sync.Once
 	TLS       *tls.ConnectionState
-	URL       *url.URL
+	URL       URL
 }
 
 func newIncomingRequest(req *http.Request) *IncomingRequest {
@@ -37,7 +36,7 @@ func newIncomingRequest(req *http.Request) *IncomingRequest {
 		req:    req,
 		Header: newHeader(req.Header),
 		TLS:    req.TLS,
-		URL:    req.URL,
+		URL:    URL{url: req.URL},
 	}
 }
 
--- a/safehttp/incoming_request.go
+++ b/safehttp/incoming_request.go
@@ -28,7 +28,7 @@ type IncomingRequest struct {
 	Header    Header
 	parseOnce sync.Once
 	TLS       *tls.ConnectionState
-	URL       URL
+	URL       *URL
 }
 
 func newIncomingRequest(req *http.Request) *IncomingRequest {
@@ -36,7 +36,7 @@ func newIncomingRequest(req *http.Request) *IncomingRequest {
 		req:    req,
 		Header: newHeader(req.Header),
 		TLS:    req.TLS,
-		URL:    URL{url: req.URL},
+		URL:    &URL{url: req.URL},
 	}
 }
 
--- a/safehttp/url.go
+++ b/safehttp/url.go
@@ -37,6 +37,7 @@ func (u URL) Query() (Form, error) {
 // The method uses the net/url.EscapedPath method to obtain the path.
 // See the net/url.EscapedPath method for more details.
 func (u URL) String() string {
+	// The escaping is perfomed by u.url.String()
 	return u.url.String()
 }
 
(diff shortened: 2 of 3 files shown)


kele:
After Mara reviews this one, please send it to Goa folks.

safehttp/url.go
```diff
@@ -0,0 +1,63 @@
...
+package safehttp
+
+import "net/url"
+
+// URL represents a parsed URL (technically, a URI reference).
+type URL struct {
+	url *url.URL
+}
+
+// Query parses the query string in the URL and returns a form
```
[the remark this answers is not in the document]
mattiasgrenfeldt: There is not a lot of the godoc of `url.ParseQuery` that is relevant for `Query` since they take different arguments and return different values. I have included the parts I found relevant.

safehttp/url.go
```diff
@@ -0,0 +1,63 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package safehttp
+
+import "net/url"
+
+// URL represents a parsed URL (technically, a URI reference).
```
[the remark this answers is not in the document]
mattiasgrenfeldt: The part about how urls are structured doesn't seem relevant to the `safehttp.URL` type to me since it doesn't give access to the same parts of the url as `net/url.URL` does.

I Included the parts about Path and String under their respective godoc comments.

safehttp/incoming_request.go
```diff
@@ -29,15 +28,15 @@ type IncomingRequest struct {
 	Header    Header
 	parseOnce sync.Once
 	TLS       *tls.ConnectionState
-	URL       *url.URL
+	URL       URL
```
stapelberg, line 31: Why are we not using a pointer here? This will be a choice we’re stuck with.

mattiasgrenfeldt: I changed it to be a pointer. Why is it better if it's a pointer?

stapelberg: Because then we have the option to declare methods with a pointer receiver later on, allowing us to modify fields of our URL struct, if we ever need to.

If we don’t use a pointer here, we don’t have that option.

safehttp/url.go
```diff
@@ -0,0 +1,72 @@
...
+type URL struct {
+	url *url.URL
+}
+
+// Query parses the query string in the URL and returns a form
+// containing its values. The returned error describes the first
+// decoding error encountered, if any.
+func (u URL) Query() (Form, error) {
```
kele, line 27: Please use a different type.

If the implementation is virtually the same, you can do `type QueryParams Form`.

safehttp/url.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	return Form{values: map[string][]string(v)}, nil
+}
+
+// String reassembles the URL into a valid URL string.
+//
+// The method uses the net/url.EscapedPath method to obtain the path.
+// See the net/url.EscapedPath method for more details.
+func (u URL) String() string {
+	return u.url.String()
```
kele, line 40: Add a comment inside the function that this is where the escaping really happens.

safehttp/url_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+import (
+	"net/url"
+	"testing"
+)
+
+func TestURLToString(t *testing.T) {
+	want := "http://www.example.com/asdf?fruit=apple"
+
+	netURL, err := url.Parse(want)
```
kele, line 25: Make it a Fatal.

Generally: prefer making test setup errors fatal ones. If the test setup fails, then it doesn't matter whether the test passes or not - we need to rerun it again as the context changes.

safehttp/url.go
mattiasgrenfeldt: Why use a different type? I don't see the point when it will be identical. I think I should instead update the documentation of Form.

PR #61, opened 2020-08-10
Make it necessary to provide Content-Type to get a http.ResponseWriter in the dispatcher

### LINKED ISSUE #60: The dispatcher can write a response without providing a Content-Type header.
Currently the `safehttp.Dispatcher` can write a response without providing the `Content-Type` header. This could potentially lead to security issues. We therefore want to ensure that a `Content-Type` header is always written on each response.

This could be accomplished by creating a wrapper around the `http.ResponseWriter` such that the dispatcher needs to first specify a `Content-Type` before getting access to the responsewriter.

mattiasgrenfeldt:
Addresses #60 

To ensure that a Content-Type is written to each response a wrapper called ResponseWriterHolder which wraps a http.ResponseWriter is introduced in the safehttp.Dispatcher interface. For the Dispatcher to get access to the responsewriter it first needs to call ResponseWriterHolder.Release with a specific status code and a content type. The responsewriter is then returned and can be used as normal.

### DIFF
--- a/safehttp/response_writer.go
+++ b/safehttp/response_writer.go
@@ -93,8 +93,8 @@ type ResponseWriterContainer struct {
 // Release releases the held http.ResponseWriter given that a status code
 // and a content type are provided. The status code and the content type are
 // written to the response before returning the ResponseWriter.
-func (c ResponseWriterContainer) Release(statusCode int, contentType string) http.ResponseWriter {
+func (c ResponseWriterContainer) Release(statusCode StatusCode, contentType string) http.ResponseWriter {
 	c.w.Header().Set("Content-Type", contentType)
-	c.w.WriteHeader(statusCode)
+	c.w.WriteHeader(int(statusCode))
 	return c.w
 }
--- a/safehttp/response_writer.go
+++ b/safehttp/response_writer.go
@@ -39,7 +39,7 @@ type Result struct{}
 
 // Write TODO
 func (w *ResponseWriter) Write(resp Response) Result {
-	if err := w.d.Write(ResponseWriterHolder{w: w.rw}, resp); err != nil {
+	if err := w.d.Write(ResponseWriterContainer{w: w.rw}, resp); err != nil {
 		panic("error")
 	}
 	return Result{}
@@ -47,7 +47,7 @@ func (w *ResponseWriter) Write(resp Response) Result {
 
 // WriteTemplate TODO
 func (w *ResponseWriter) WriteTemplate(t Template, data interface{}) Result {
-	if err := w.d.ExecuteTemplate(ResponseWriterHolder{w: w.rw}, t, data); err != nil {
+	if err := w.d.ExecuteTemplate(ResponseWriterContainer{w: w.rw}, t, data); err != nil {
 		panic("error")
 	}
 	return Result{}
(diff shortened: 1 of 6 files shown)


safehttp/response_writer.go
```diff
@@ -80,6 +80,21 @@ func (w ResponseWriter) Header() Header {
...
+
+// ResponseWriterHolder holds an http.ResponseWriter until
+// a status code and a content type to be written is provided.
+type ResponseWriterHolder struct {
+	w http.ResponseWriter
+}
+
+// Release releases the held http.ResponseWriter given that a status code
+// and a content type is provided. The status code and the content type are
+// written to the response before returning the responsewriter.
```
mihalimara22, line 95: nit: ResponseWriter

safehttp/response_writer.go
```diff
@@ -80,6 +80,21 @@ func (w ResponseWriter) Header() Header {
...
 type Dispatcher interface {
-	Write(rw http.ResponseWriter, resp Response) error
-	ExecuteTemplate(rw http.ResponseWriter, t Template, data interface{}) error
+	Write(h ResponseWriterHolder, resp Response) error
+	ExecuteTemplate(h ResponseWriterHolder, t Template, data interface{}) error
+}
+
+// ResponseWriterHolder holds an http.ResponseWriter until
+// a status code and a content type to be written is provided.
+type ResponseWriterHolder struct {
```
mihalimara22, line 89: Maybe `ResponseWriterContainer` instead of `Holder`?

safehttp/response_writer.go
```diff
@@ -80,6 +80,21 @@ func (w ResponseWriter) Header() Header {
...
+	w http.ResponseWriter
+}
+
+// Release releases the held http.ResponseWriter given that a status code
+// and a content type is provided. The status code and the content type are
```
mihalimara22, line 94: are provided

safehttp/response_writer.go
```diff
@@ -80,6 +80,21 @@ func (w ResponseWriter) Header() Header {
...
 // Dispatcher TODO
 type Dispatcher interface {
-	Write(rw http.ResponseWriter, resp Response) error
-	ExecuteTemplate(rw http.ResponseWriter, t Template, data interface{}) error
+	Write(h ResponseWriterHolder, resp Response) error
+	ExecuteTemplate(h ResponseWriterHolder, t Template, data interface{}) error
+}
+
+// ResponseWriterHolder holds an http.ResponseWriter until
+// a status code and a content type to be written is provided.
```
mihalimara22, line 88: are provided

safehttp/response_writer.go
```diff
@@ -80,6 +80,21 @@ func (w ResponseWriter) Header() Header {
...
+type ResponseWriterHolder struct {
+	w http.ResponseWriter
+}
+
+// Release releases the held http.ResponseWriter given that a status code
+// and a content type is provided. The status code and the content type are
+// written to the response before returning the responsewriter.
+func (h ResponseWriterHolder) Release(statusCode int, contentType string) http.ResponseWriter {
```
mihalimara22, line 96: Why not a `safehttp.StatusCode`?

mattiasgrenfeldt: I feel like this statusCode is in a gray area between `net/http` and `safehttp` since it is used in the dispatcher. But it's probably still better to have it as a `safehttp.StatusCode` as you say. :+1:

safehttp/response_writer.go
```diff
@@ -80,6 +80,21 @@ func (w ResponseWriter) Header() Header {
...
 type Dispatcher interface {
-	Write(rw http.ResponseWriter, resp Response) error
-	ExecuteTemplate(rw http.ResponseWriter, t Template, data interface{}) error
+	Write(c ResponseWriterContainer, resp Response) error
+	ExecuteTemplate(c ResponseWriterContainer, t Template, data interface{}) error
+}
+
+// ResponseWriterContainer holds an http.ResponseWriter until
+// a status code and a content type to be written are provided.
+type ResponseWriterContainer struct {
```
empijei, line 89: We should probably discuss the name for this.

@kele

safehttp/response_writer.go
```diff
@@ -80,6 +80,21 @@ func (w ResponseWriter) Header() Header {
...
+	w http.ResponseWriter
+}
+
+// Release releases the held http.ResponseWriter given that a status code
+// and a content type are provided. The status code and the content type are
+// written to the response before returning the ResponseWriter.
+func (c ResponseWriterContainer) Release(statusCode StatusCode, contentType string) http.ResponseWriter {
```
empijei, line 96: Release is not a good name for this.

mattiasgrenfeldt:
Let's just put this entire PR on ice until this has been discussed more.

kele:
Closing in favor of `Dispatcher.ContentType()` method: https://github.com/google/go-safeweb/pull/147.

The main reason here is: adding an API that "guards" getting the `http.ResponseWriter` seems clunky and doesn't really provide that much security guarantees (i.e. the `Dispatcher` can ask for `HTMLWriter` and still write plain text). That was one of the ideas we had and were leaning towards, but we've changed our minds.

Instead, we keep the guarantee of having **a** content type written, as now the framework will call `Dispatcher.ContentType()` and set the header (instead of the `Dispatcher` doing so). 

This also seems like a simpler alternative over using interceptors to set the content type (which we've considered but couldn't find a good way of implementing).

PR #62, opened 2020-08-10
Add a plugin to automatically nonce templates.

empijei:


safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -0,0 +1,193 @@
...
+			}
+			if match {
+				triggeredRules = append(triggeredRules, r)
+			}
+		}
+	}
+
+	// Emit the rewritten HTML
+	{
+		// Write the "<" symbol and the tag name, e.g. "<script"
```
bsiegert, line 167: I am slightly concerned about the possibility that this may go wrong, e.g. if the input string has whitespace between the `<` and the tag name (is that allowed?). Maybe use `len(tagname) + bytes.Index(raw, tagname)`?

empijei: that is not valid HTML and would be rejected by any html template. Also, that is not a token that is emitted by the tokenizer.

safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -0,0 +1,193 @@
...
+// XSRFTokens constructs a Config to add the given string as a child node to forms.
+func XSRFTokens(inputTag string) Config {
+	return Config{Rule{
+		Name:     "XSRFTokens on forms",
+		OnTag:    "form",
+		AddNodes: []string{inputTag}}}
+}
+
+// Transform rewrites the given template according to the given configs.
+// If the passed io.Rewriter has a `Size() int64` method it will be used to pre-allocate buffers.
```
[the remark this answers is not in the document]
empijei: I would expect this to mostly be implicitly implemented (e.g. strings.Reader has it). The idea is that if the info is available it's used, otherwise meh. I don't think this is interesting enough to expose it, I was even considering to remove the comment from the doc.

I would not export an interface just to allow people to micro optimize code.

safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -0,0 +1,194 @@
...
+	// OnTag is the tag to be used to trigger the rule.
+	OnTag string
+	// WithAttributes is a filter applied on tags to decide whether to run the Rule:
+	// only tags with the given attributes key:value will be matched.
+	WithAttributes map[string]string
+	// AddAttributes is a list of strings to add to the HTML in place of attributes.
+	// All the given strings will be appended verbatim after the matched tag so they
+	// should be prefixed with a space.
+	AddAttributes []string
+	// AddNodes is a list of nodes to append immediately after the openin tag that matched.
```
mattiasgrenfeldt, line 26: opening*

safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -0,0 +1,194 @@
...
+			OnTag:          "link",
+			WithAttributes: map[string]string{"rel": "preload", "as": "script"},
+			AddAttributes:  []string{nonceAttr},
+		},
+	}
+}
+
+// XSRFTokensDefault is the default config to add hidden inputs to forms to provide
+// an anti-XSRF token. The rewritten template expects the XSRFToken Func to be available
+// in the template to provide tokens and sets the name for the sent value to be "xsrftoken".
```
mattiasgrenfeldt, line 62: Should be `"xsrf-token"`

safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -0,0 +1,194 @@
...
+// Rule is a directive to instruct Transform on how to rewrite the given template.
+type Rule struct {
+	// Name is used for debug purposes in case rewriting fails.
+	Name string
+	// OnTag is the tag to be used to trigger the rule.
+	OnTag string
+	// WithAttributes is a filter applied on tags to decide whether to run the Rule:
+	// only tags with the given attributes key:value will be matched.
+	WithAttributes map[string]string
+	// AddAttributes is a list of strings to add to the HTML in place of attributes.
```
kele, line 22:
```suggestion
	// AddAttributes is a list of strings to add to the HTML as attributes.
```

safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -0,0 +1,194 @@
...
+func XSRFTokens(inputTag string) Config {
+	return Config{Rule{
+		Name:     "XSRFTokens on forms",
+		OnTag:    "form",
+		AddNodes: []string{inputTag}}}
+}
+
+// Transform rewrites the given template according to the given configs.
+// If the passed io.Rewriter has a `Size() int64` method it will be used to pre-allocate buffers.
+func Transform(src io.Reader, cfg ...Config) (tpl string, _ error) {
```
kele, line 75: Why named args?

safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -0,0 +1,194 @@
...
+	if sizer, ok := src.(interface{ Size() int64 }); ok {
+		rw.out.Grow(int(sizer.Size()))
+	}
+	for _, c := range cfg {
+		for _, r := range c {
+			rw.rules[r.OnTag] = append(rw.rules[r.OnTag], r)
+		}
+	}
+	if err := rw.rewrite(); err != nil {
+		return "", fmt.Errorf("transforming template: %v", err)
```
kele, line 90: The "transforming template" is obvious from the function call. If the caller needs to be reminded of what they're calling, they can add it themselves.

safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -0,0 +1,194 @@
...
+			key, val, more := r.tokenizer.TagAttr()
+			hasAttr = more
+			attributes[string(key)] = string(val)
+		}
+	}
+
+	// Filter rules by attributes
+	var triggeredRules []Rule
+	{
+		rules := r.rules[tagname]
```
kele, line 151: nit: Do you need the variable?

safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -0,0 +1,194 @@
...
+		for _, r := range rules {
+			match := true
+			for k, v := range r.WithAttributes {
+				if attributes[k] != v {
+					match = false
```
kele, line 156: `break`?

PR #67, opened 2020-08-13
Implemented a safehttp.Cookie type, made it readable from IncomingRequest and settable from ResponseWriter.

### LINKED ISSUE #56: Come up with a better way of setting cookies.
Currently, to set cookies on a response you have to use the `safehttp.Header.SetCookie` method which takes a `http.Cookie`. All other ways to modify header values block the modification of the `Set-Cookie` header. 

I suggest that we remove the `safehttp.Header.SetCookie` method, implement a `safehttp.Cookie` type with safe defaults and add a `SetCookie` method to `safehttp.ResponseWriter`.

mattiasgrenfeldt:
Fixes #56 

This PR implements a cookie type with safe default values.

It also implements cookie reading methods on `IncomingRequest` and cookie setting methods on `ResponseWriter`. This mimics the setters and getters available for cookies in `net/http`.

The `safehttp.Header.SetCookie` method has been removed. Now, cookie interactions only happen with the `ResponseWriter` and the `IncomingRequest`  and never with the `Header` directly.

### DIFF
--- a/safehttp/cookie.go
+++ b/safehttp/cookie.go
@@ -75,9 +75,9 @@ func (c *Cookie) SetSameSite(s SameSite) {
 
 // SetMaxAge sets the MaxAge attribute.
 //
-// MaxAge=0 means no 'Max-Age' attribute specified.
-// MaxAge<0 means delete cookie now, equivalently 'Max-Age: 0'
-// MaxAge>0 means Max-Age attribute present and given in seconds
+// MaxAge = 0 means no 'Max-Age' attribute specified.
+// MaxAge < 0 means delete cookie now, equivalently 'Max-Age: 0'
+// MaxAge > 0 means Max-Age attribute present and given in seconds
 func (c *Cookie) SetMaxAge(maxAge int) {
 	c.wrapped.MaxAge = maxAge
 }
--- a/safehttp/cookie.go
+++ b/safehttp/cookie.go
@@ -64,6 +64,8 @@ const (
 // SetSameSite sets the SameSite attribute.
 func (c *Cookie) SetSameSite(s SameSite) {
 	switch s {
+	case SameSiteLaxMode:
+		c.wrapped.SameSite = http.SameSiteLaxMode
 	case SameSiteStrictMode:
 		c.wrapped.SameSite = http.SameSiteStrictMode
 	case SameSiteNoneMode:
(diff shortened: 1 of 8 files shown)


safehttp/cookie.go
```diff
@@ -0,0 +1,111 @@
...
+
+// SetSameSite sets the SameSite attribute.
+func (c *Cookie) SetSameSite(s SameSite) {
+	switch s {
+	case SameSiteStrictMode:
+		c.wrapped.SameSite = http.SameSiteStrictMode
+	case SameSiteNoneMode:
+		c.wrapped.SameSite = http.SameSiteNoneMode
+	}
+}
```
kele, line 72: What about SameSiteLaxMode?

safehttp/header.go
```diff
@@ -114,15 +114,18 @@ func (h Header) Values(name string) []string {
...
-func (h Header) SetCookie(cookie *http.Cookie) {
-	if v := cookie.String(); v != "" {
-		h.wrapped.Add("Set-Cookie", v)
+// addCookie adds the cookie provided as a Set-Cookie header in the header
+// collection. If the cookie is nil or cookie.Name() is invalid, no header is
+// added and an error is returned. This is the only method that can modify the
+// Set-Cookie header. If other methods try to modify the header they will return
+// errors.
+func (h Header) addCookie(c *Cookie) error {
+	v := c.wrapped.String()
```
kele, line 123: Maybe `safehttp.Cookie` should also implement `String`? Also useful for debugging.

safehttp/incoming_request.go
```diff
@@ -108,3 +108,24 @@ func (r *IncomingRequest) MultipartForm(maxMemory int64) (*MultipartForm, error)
 			}},
 		nil
 }
+
+// Cookie returns the named cookie provided in the request or ErrNoCookie if not
```
kele, line 112: net/http.ErrNoCookie

safehttp/cookie.go
mattiasgrenfeldt: Well, `SameSiteLaxMode` is the default. But I can add the option to set to lax mode, so that you can do:
```
c ;= NewCookie(...)
c.SetSameSite(X)
c.SetSameSite(SameSiteLaxMode)
```

mattiasgrenfeldt:

PR #69, opened 2020-08-13
Added integration test which installs the HSTS plugin on a ServeMux and serves a request.

### LINKED ISSUE #68: Create integration tests between ServeMux and HSTS plugin now that plugin infrastructure exists.
Now that there is interceptor/plugin infrastructure in the `ServeMux` there should be integration tests making sure that the `HSTS` plugin works with the `ServeMux`.

mattiasgrenfeldt:
Fixes #68 

Now that there is plugin infrastructure in the `ServeMux` there should be integration tests with the HSTS plugin.

### DIFF
--- a/tests/integration/hsts/hsts_test.go
+++ b/tests/integration/hsts/hsts_test.go
@@ -73,8 +73,7 @@ func TestHSTSServeMuxInstall(t *testing.T) {
 	})
 	mux.Handle("/asdf", safehttp.MethodGet, handler)
 
-	p := hsts.NewPlugin()
-	mux.Install("hsts", &p)
+	mux.Install("hsts", hsts.NewPlugin())
 
 	b := strings.Builder{}
 	rr := newResponseRecorder(&b)
--- a/tests/integration/hsts/hsts_test.go
+++ b/tests/integration/hsts/hsts_test.go
@@ -1,0 +1,100 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


tests/integration/hsts/hsts_test.go
```diff
@@ -0,0 +1,100 @@
...
+	return r.header
+}
+
+func (r *responseRecorder) WriteHeader(statusCode int) {
+	r.status = statusCode
+}
+
+func (r *responseRecorder) Write(data []byte) (int, error) {
+	return r.writer.Write(data)
+}
```
kele, line 66: Isn't all of this now available in some kind of a test package?

tests/integration/hsts/hsts_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+func TestHSTSServeMuxInstall(t *testing.T) {
+	mux := safehttp.NewServeMux(&dispatcher{}, "foo.com")
+
+	handler := safehttp.HandlerFunc(func(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+		return w.Write(safehtml.HTMLEscaped("<h1>Hello World!</h1>"))
+	})
+	mux.Handle("/asdf", safehttp.MethodGet, handler)
+
+	p := hsts.NewPlugin()
```
kele, line 76: This should return something ready to pass to `mux.Install` without any kind of `&` or `*` magic.

tests/integration/hsts/hsts_test.go
mattiasgrenfeldt: Right, the `safehttptest` package. That package can only produce `safehttp.ResponseWriter`s and `safehttp.IncomingRequest`s, but to test the `ServeMux` we need `http.ResponseWriter`s and `http.Request`s since its `ServeHTTP` method has the `net/http` signature.

mattiasgrenfeldt:

PR #91, opened 2020-08-17
Fetch Metadata Plugin Implementation

### LINKED ISSUE #78: Add Fetch Metadata protection
The plugin should validate incoming requests against a FM policy and potentially decide to reject them.

Two policies should be supported: Resource Isolation and Navigation Isolation. It should be possible to selectively disable this plugin on CORS endpoints.

mihalimara22:
Fixes #78 

Implemented the initial version of Fetch Metadata plugin. It enforces a Resource Isolation Policy by default and supports enabling and disabling a Navigation Isolation Policy. The user can provide a custom policy instead of the default policy if needed. It also supports switching to a reporting mode that allows all requests to pass and only logs violation.

### DIFF
--- a/safehttp/incoming_request.go
+++ b/safehttp/incoming_request.go
@@ -44,16 +44,21 @@ func NewIncomingRequest(req *http.Request) *IncomingRequest {
 	}
 }
 
+// Method specifies the HTTP method of an IncomingRequest.
+func (r *IncomingRequest) Method() string {
+	return r.req.Method
+}
+
 // Host returns the host the request is targeted to.
-// TODO(@mihalimara22): Remove this after the safehttp.URL type has been
-// implemented.
+// Method specifies the HTTP method of an IncomingRequest.
+// TODO(@mihalimara22): Remove this after the safehttp.URL type has been    implemented
 func (r *IncomingRequest) Host() string {
 	return r.req.Host
 }
 
 // Path returns the relative path of the URL in decoded format (e.g. %47%6f%2f
 // becomes /Go/).
-// TODO(@mihalimara22): Remove this after the safehttp.URL type has been     //
+// TODO(@mihalimara22): Remove this after the safehttp.URL type has been
 // implemented.
 func (r *IncomingRequest) Path() string {
 	return r.req.URL.Path
--- a/safehttp/incoming_request.go
+++ b/safehttp/incoming_request.go
@@ -50,8 +50,7 @@ func (r *IncomingRequest) Method() string {
 }
 
 // Host returns the host the request is targeted to.
-// Method specifies the HTTP method of an IncomingRequest.
-// TODO(@mihalimara22): Remove this after the safehttp.URL type has been    implemented
+// TODO(@mihalimara22): Remove this after the safehttp.URL type has been implemented
 func (r *IncomingRequest) Host() string {
 	return r.req.Host
 }
--- a/safehttp/plugins/fetch_metadata/fetch_metadata.go
+++ b/safehttp/plugins/fetch_metadata/fetch_metadata.go
@@ -133,7 +133,7 @@ func (p *Plugin) SetEnforce() {
 // the  violation is reported. If a redirectURL was provided and the Navigation
 // Isolation Policy is enabled and fails, the IncomingRequest will be
 // redirected to redirectURL.
-func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+func (p *Plugin) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
 	// TODO(mihalimara22): Remove and disable using configurations when those
 	// have been implemented
 	if p.corsProtected[r.URL.Path()] {
--- a/safehttp/plugins/fetch_metadata/fetch_metadata.go
+++ b/safehttp/plugins/fetch_metadata/fetch_metadata.go
@@ -16,7 +16,6 @@ package fetchmetadata
 
 import (
 	"github.com/google/go-safeweb/safehttp"
-	"log"
 	"net/url"
 )
 
@@ -32,28 +31,12 @@ var (
 		"nested-navigate": true,
 	}
 	navigationalDest = map[string]bool{
-		"audio":           true,
-		"audioworklet":    true,
 		"document":        true,
-		"empty":           true,
-		"font":            true,
-		"image":           true,
-		"manifest":        true,
-		"paintworklet":    true,
-		"report":          true,
-		"script":          true,
-		"serviceworker":   true,
-		"sharedworker":    true,
-		"style":           true,
-		"track":           true,
-		"video":           true,
-		"worker":          true,
-		"xslt":            true,
 		"nested-document": true,
 	}
 	statePreservingMethods = map[string]bool{
-		"GET":  true,
-		"HEAD": true,
+		safehttp.MethodGet:  true,
+		safehttp.MethodHead: true,
 	}
 )
 
@@ -88,9 +71,7 @@ func NewPlugin(endpoints ...string) *Plugin {
 	for _, e := range endpoints {
 		m[e] = true
 	}
-	return &Plugin{
-		corsProtected: m,
-	}
+	return &Plugin{corsProtected: m}
 }
 
 func (p *Plugin) resourceIsolationPolicy(r *safehttp.IncomingRequest) bool {
@@ -99,7 +80,6 @@ func (p *Plugin) resourceIsolationPolicy(r *safehttp.IncomingRequest) bool {
 	site := h.Get("Sec-Fetch-Site")
 	mode := h.Get("Sec-Fetch-Mode")
 	dest := h.Get("Sec-Fetch-Dest")
-	log.Print(method, site, mode, dest)
 	if site != "cross-site" {
 		// The request is allowed to pass because one of the following applies:
 		// - Fetch Metadata is not supported by the browser
@@ -120,7 +100,6 @@ func (p *Plugin) resourceIsolationPolicy(r *safehttp.IncomingRequest) bool {
 		// safe destination so we  allow it to pass .
 		return true
 	}
-	log.Print("here?")
 	// The request is cross-site and not a simple navigation or from an unsafe
 	// destination so it is rejected.
 	return false
@@ -128,8 +107,7 @@ func (p *Plugin) resourceIsolationPolicy(r *safehttp.IncomingRequest) bool {
 
 // SetReportOnly sets the Fetch Metadata policy mode to "report". This will
 // allow requests that violate the policy to pass, but will report the violation
-// using the RequestLogger. The method will panic if no RequestLogger is
-// provided.
+// using the RequestLogger. The method will panic if RequestLogger is nil.
 func (p *Plugin) SetReportOnly(logger RequestLogger) {
 	if logger == nil {
 		panic("logging service required for Fetch Metadata report mode")
@@ -138,9 +116,8 @@ func (p *Plugin) SetReportOnly(logger RequestLogger) {
 	p.reportOnly = true
 }
 
-// SetEnforce sets the Fetch Metadata policy mode to "enforce" and sets the
-// RequestLogger to nil. This will reject any requests that violates the policy
-// provided by the plugin.
+// SetEnforce sets the Fetch Metadata policy mode to "enforce". This will reject
+// any requests that violates the policy provided by the plugin.
 func (p *Plugin) SetEnforce() {
 	p.reportOnly = false
 	p.logger = nil
@@ -156,12 +133,14 @@ func (p *Plugin) SetEnforce() {
 func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
 	reject := false
 	endpoint := r.URL.String()
+
 	if _, ok := p.corsProtected[endpoint]; ok {
 		// The request is targeted to an endpoint on which Fetch Metadata
 		// policies are disabled because it is CORS-protected so we don't apply
 		// the policies.
 		return safehttp.Result{}
 	}
+
 	if p.NavIsolation {
 		h := r.Header
 		if site, mode := h.Get("Sec-Fetch-Site"), h.Get("Sec-Fetch-Mode"); site == "cross-site" && navigationalModes[mode] {
@@ -175,7 +154,7 @@ func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) 
 			reject = true
 		}
 	}
-	log.Print("here")
+
 	if reject || !p.resourceIsolationPolicy(r) {
 		if !p.reportOnly {
 			return w.ClientError(safehttp.StatusForbidden)
--- a/safehttp/plugins/fetch_metadata/fetch_metadata.go
+++ b/safehttp/plugins/fetch_metadata/fetch_metadata.go
@@ -1,0 +1,121 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fetchmetadata
+
+import (
+	"github.com/google/go-safeweb/safehttp"
+)
+
+const (
+	enforceMode = "enforce"
+	reportMode  = "report"
+)
+
+// LoggingService TODO
+type LoggingService interface {
+	Log(ir *safehttp.IncomingRequest)
+}
+
+// Plugin implements Fetch Metadata functionality.
+// See https://www.w3.org/TR/fetch-metadata/ for more details.
+type Plugin struct {
+	mode   string
+	policy func(*safehttp.IncomingRequest) bool
+	logger LoggingService
+}
+
+// NewPlugin creates a new Fetch Metadata plugin using the defaultPolicy and
+// sets the mode to enforce.
+func NewPlugin() *Plugin {
+	return &Plugin{
+		mode:   enforceMode,
+		policy: defaultPolicy,
+	}
+}
+
+// defaultPolicy applies a Resource Isolation Policy to the
+// safehttp.IncomingRequest, allowing it to pass only if it conforms to it.
+//
+// See https://web.dev/fetch-metadata/ for more information.
+func defaultPolicy(r *safehttp.IncomingRequest) bool {
+	switch r.Header.Get("Sec-Fetch-Site") {
+	case "":
+		// Fetch Metadata is not supported by the browser so we allow the
+		// request to pass.
+		return true
+	case "same-origin", "same-site", "none":
+		// The request originated from a site served by your own server
+		// ("same-origin"), a subdomain of your site ("same-site", e.g.
+		// bar.foo.com made a request to foo.com) or was caused by the user
+		// explicitly interacting with the user-agent ("none"). Therefore it is
+		// allowed to pass.
+		return true
+	}
+	if m := r.Method(); r.Header.Get("Sec-Fetch-Mode") == "navigate" && (m == "GET" || m == "HEAD") {
+		if dest := r.Header.Get("Sec-Fetch-Dest"); dest == "object" || dest == "embed" {
+			// The request is cross-site and originates from <object> or <embed>
+			// so it is rejected.
+			return false
+		}
+		// The request is cross-site, but a simple top-level navigation so we
+		// allow it to pass.
+		return true
+	}
+	// The request is cross-site and not navigational so it is rejected.
+	return false
+}
+
+// SetPolicy allows changing the default Fetch Metadata policy
+// to a user-provided policy.
+func (p *Plugin) SetPolicy(policy func(*safehttp.IncomingRequest) bool) {
+	p.policy = policy
+}
+
+// SetReportMode sets the Fetch Metadata policy mode to "report". This will
+// allow requests that violate the policy to pass, but will report the violation
+// using the LoggingService. The method will panic if no LoggingService is
+// provided.
+func (p *Plugin) SetReportMode(logger LoggingService) {
+	if logger == nil {
+		panic("logging service required for Fetch Metadata report mode")
+	}
+	p.logger = logger
+	p.mode = reportMode
+}
+
+// SetEnforceMode sets the Fetch Metadata policy mode to "enforce". This will
+// reject any requests that violate the policy provided by the plugin.
+func (p *Plugin) SetEnforceMode() {
+	p.mode = enforceMode
+}
+
+// Before validates the safehttp.IncomingRequest using the Fetch Metadata policy
+// provided by the  plugin. It only allows request to pass if they conform to
+// the policy or if the mode is set to "report", in which case the request is
+// allowed to pass but the violation is reported. If the browser does not have
+// Fetch Metadata support implemented, the policy will not be applied and all
+// requests will be allowed to pass.
+func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+	if !p.policy(r) {
+		switch p.mode {
+		case enforceMode:
+			return w.ClientError(safehttp.StatusForbidden)
+		case reportMode:
+			p.logger.Log(r)
+		}
+	}
+
+	return safehttp.Result{}
+}
--- a/safehttp/plugins/fetch_metadata/fetch_metadata.go
+++ b/safehttp/plugins/fetch_metadata/fetch_metadata.go
@@ -29,7 +29,7 @@ type LoggingService interface {
 type Plugin struct {
 	// Policy is the Fetch Metadata policy that decides whether a request should
 	// be allowed to pass or be rejected. This will be set to a default Resource
-	// Isolation Policy by default but can be changed by the user.
+	// Isolation Policy, but can be changed by the user.
 	Policy     func(*safehttp.IncomingRequest) bool
 	reportOnly bool
 	logger     LoggingService
@@ -62,7 +62,7 @@ func defaultPolicy(r *safehttp.IncomingRequest) bool {
 		// allowed to pass.
 		return true
 	}
-	if m := r.Method(); h.Get("Sec-Fetch-Mode") == "navigate" && (m == "GET" || m == "HEAD") {
+	if m := r.Method(); h.Get("Sec-Fetch-Mode") == "navigate" && (m == safehttp.MethodGet || m == safehttp.MethodHead) {
 		if dest := h.Get("Sec-Fetch-Dest"); dest == "object" || dest == "embed" {
 			// The request is cross-site and originates from <object> or <embed>
 			// so it is rejected.
--- a/safehttp/plugins/fetch_metadata/fetch_metadata.go
+++ b/safehttp/plugins/fetch_metadata/fetch_metadata.go
@@ -16,7 +16,6 @@ package fetchmetadata
 
 import (
 	"github.com/google/go-safeweb/safehttp"
-	"net/url"
 )
 
 // RequestLogger is a user-provided service for logging Fetch Metadata policy
@@ -56,7 +55,7 @@ type Plugin struct {
 	NavIsolation bool
 	// RedirectURL can optionally indicate an URL to which the user can be
 	// redirected in case the Navigation Isolation policy rejects the request.
-	RedirectURL   string
+	RedirectURL   *safehttp.URL
 	reportOnly    bool
 	logger        RequestLogger
 	corsProtected map[string]bool
@@ -76,11 +75,8 @@ func NewPlugin(endpoints ...string) *Plugin {
 
 func (p *Plugin) resourceIsolationPolicy(r *safehttp.IncomingRequest) bool {
 	h := r.Header
-	method := r.Method()
-	site := h.Get("Sec-Fetch-Site")
-	mode := h.Get("Sec-Fetch-Mode")
-	dest := h.Get("Sec-Fetch-Dest")
-	if site != "cross-site" {
+
+	if h.Get("Sec-Fetch-Site") != "cross-site" {
 		// The request is allowed to pass because one of the following applies:
 		// - Fetch Metadata is not supported by the browser
 		// - the request is same-origin, same-site or caused by the user
@@ -88,6 +84,9 @@ func (p *Plugin) resourceIsolationPolicy(r *safehttp.IncomingRequest) bool {
 		return true
 	}
 
+	method := r.Method()
+	mode := h.Get("Sec-Fetch-Mode")
+	dest := h.Get("Sec-Fetch-Dest")
 	// Allow CORS options requests if neither Mode nor Dest is set.
 	// https://github.com/w3c/webappsec-fetch-metadata/issues/35
 	// https://bugs.chromium.org/p/chromium/issues/detail?id=979946
@@ -96,7 +95,7 @@ func (p *Plugin) resourceIsolationPolicy(r *safehttp.IncomingRequest) bool {
 	}
 
 	if navigationalModes[mode] && navigationalDest[dest] && statePreservingMethods[method] {
-		// The rquest is cross-site, but a simple top-level navigation from a
+		// The request is cross-site, but a simple top-level navigation from a
 		// safe destination so we  allow it to pass .
 		return true
 	}
@@ -124,16 +123,15 @@ func (p *Plugin) SetEnforce() {
 }
 
 // Before validates the safehttp.IncomingRequest using the Resource Isolation
-// Policy and, if enabled, the Navigation Isolation Policy . It only allows
-// request to pass if they conform to the policy, if it's targeted to a
-// CORS-protected endpoint, or if the mode is set to
-// "report", in which case the request is allowed to pass but the violation is
-// reported. If a redirectURL was provided and the Navigation Isolation Policy
-// fails, the IncomingRequest will be redirected to the redirectURL.
+// Policy and, if enabled, the Navigation Isolation Policy. It only allows
+// requests to pass if they conform to the policy, if it's targeted to one of
+// the CORS-protected endpoints, specified when creating the plugin, or if the
+// mode is set to  "report", in which case the request is allowed to pass but
+// the  violation is reported. If a redirectURL was provided and the Navigation
+// Isolation Policy is enabled and fails, the IncomingRequest will be
+// redirected to redirectURL.
 func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
-	reject := false
 	endpoint := r.URL.String()
-
 	if _, ok := p.corsProtected[endpoint]; ok {
 		// The request is targeted to an endpoint on which Fetch Metadata
 		// policies are disabled because it is CORS-protected so we don't apply
@@ -141,18 +139,13 @@ func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) 
 		return safehttp.Result{}
 	}
 
-	if p.NavIsolation {
-		h := r.Header
-		if site, mode := h.Get("Sec-Fetch-Site"), h.Get("Sec-Fetch-Mode"); site == "cross-site" && navigationalModes[mode] {
-			if p.RedirectURL != "" {
-				u, err := url.Parse(p.RedirectURL)
-				if err != nil {
-					return w.ServerError(safehttp.StatusInternalServerError)
-				}
-				return w.Redirect(r, u.String(), safehttp.StatusMovedPermanently)
-			}
-			reject = true
+	reject := false
+	h := r.Header
+	if p.NavIsolation && h.Get("Sec-Fetch-Site") == "cross-site" && navigationalModes[h.Get("Sec-Fetch-Mode")] {
+		if p.RedirectURL != nil {
+			return w.Redirect(r, p.RedirectURL.String(), safehttp.StatusMovedPermanently)
 		}
+		reject = true
 	}
 
 	if reject || !p.resourceIsolationPolicy(r) {
--- a/safehttp/plugins/fetch_metadata/fetch_metadata.go
+++ b/safehttp/plugins/fetch_metadata/fetch_metadata.go
@@ -16,8 +16,8 @@ package fetchmetadata
 
 import (
 	"github.com/google/go-safeweb/safehttp"
-	"net/url"
 	"log"
+	"net/url"
 )
 
 // RequestLogger is a user-provided service for logging Fetch Metadata policy
@@ -59,7 +59,7 @@ var (
 
 // Plugin implements Fetch Metadata functionality.
 //
-// See https://www.w3.org/TR/fetch-metadata/ and 
+// See https://www.w3.org/TR/fetch-metadata/ and
 // https://web.dev/fetch-metadata/  for more details.
 type Plugin struct {
 	// NavIsolation indicates whether the Navigation Isolation Policy should
@@ -99,7 +99,7 @@ func (p *Plugin) resourceIsolationPolicy(r *safehttp.IncomingRequest) bool {
 	site := h.Get("Sec-Fetch-Site")
 	mode := h.Get("Sec-Fetch-Mode")
 	dest := h.Get("Sec-Fetch-Dest")
-	log.Print(method,site, mode,dest)
+	log.Print(method, site, mode, dest)
 	if site != "cross-site" {
 		// The request is allowed to pass because one of the following applies:
 		// - Fetch Metadata is not supported by the browser
--- a/safehttp/plugins/fetch_metadata/fetch_metadata.go
+++ b/safehttp/plugins/fetch_metadata/fetch_metadata.go
@@ -16,6 +16,8 @@ package fetchmetadata
 
 import (
 	"github.com/google/go-safeweb/safehttp"
+	"net/url"
+	"log"
 )
 
 // RequestLogger is a user-provided service for logging Fetch Metadata policy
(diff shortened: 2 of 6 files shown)


safehttp/incoming_request.go
```diff
@@ -43,16 +43,21 @@ func NewIncomingRequest(req *http.Request) *IncomingRequest {
...
+// Method specifies the HTTP method of an IncomingRequest.
+func (r *IncomingRequest) Method() string {
+	return r.req.Method
+}
+
 // Host returns the host the request is targeted to.
-// TODO(@mihalimara22): Remove this after the safehttp.URL type has been
-// implemented.
+// Method specifies the HTTP method of an IncomingRequest.
+// TODO(@mihalimara22): Remove this after the safehttp.URL type has been    implemented
```
mattiasgrenfeldt, line 53: Remove the `Method ...` line
Remove extra spaces before `implemented`.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,121 @@
...
+			// The request is cross-site and originates from <object> or <embed>
+			// so it is rejected.
+			return false
+		}
+		// The request is cross-site, but a simple top-level navigation so we
+		// allow it to pass.
+		return true
+	}
+	// The request is cross-site and not navigational so it is rejected.
+	return false
```
mattiasgrenfeldt, line 77: Rewrite this to have the happy path on the left.
```
if not(Fetch-Mode==navigate && (method=="GET" || method=="HEAD")) {
    return false
}
if Sec-Fetch-Dest stuff {
    return false
}
return true
```

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,121 @@
...
+		policy: defaultPolicy,
+	}
+}
+
+// defaultPolicy applies a Resource Isolation Policy to the
+// safehttp.IncomingRequest, allowing it to pass only if it conforms to it.
+//
+// See https://web.dev/fetch-metadata/ for more information.
+func defaultPolicy(r *safehttp.IncomingRequest) bool {
+	switch r.Header.Get("Sec-Fetch-Site") {
```
mattiasgrenfeldt, line 53: Create a local variable `h := r.Header` to use instead of `r.Header`. Since `r.Header` is used so much this will improve readability

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,121 @@
...
+
+// LoggingService TODO
+type LoggingService interface {
+	Log(ir *safehttp.IncomingRequest)
+}
+
+// Plugin implements Fetch Metadata functionality.
+// See https://www.w3.org/TR/fetch-metadata/ for more details.
+type Plugin struct {
+	mode   string
```
mattiasgrenfeldt, line 34: Change `mode` to a bool called `ReportOnly`. This will be false by default. That is, enforcement is on by default.

Or another solution: Remove `mode` and just check whether logger is nil or not to check which mode is enabled.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	// The request is cross-site and not navigational so it is rejected.
+	return false
+}
+
+// SetPolicy allows changing the default Fetch Metadata policy
+// to a user-provided policy.
+func (p *Plugin) SetPolicy(policy func(*safehttp.IncomingRequest) bool) {
+	p.policy = policy
+}
```
mattiasgrenfeldt, line 84: Remove this setter and make `policy` exported instead

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,401 @@
...
+			}
+			if diff := cmp.Diff(test.wantHeaders, map[string][]string(rec.Header())); diff != "" {
+				t.Errorf("rec.Header() mismatch (-want +got):\n%s", diff)
+			}
+			if got := rec.Body(); got != test.wantBody {
+				t.Errorf("response body got: %q want: %q", got, test.wantBody)
+			}
+		})
+	}
+
```
mattiasgrenfeldt, line 168: Remove empty line

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,401 @@
...
```
mattiasgrenfeldt, line 400: Remove empty line

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,401 @@
...
+import (
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-safeweb/safehttp"
+	"github.com/google/go-safeweb/safehttp/plugins/fetch_metadata"
+	"github.com/google/go-safeweb/safehttp/safehttptest"
+	"testing"
+)
+
+func TestEnforceMode(t *testing.T) {
```
mattiasgrenfeldt, line 25: Change this to only test the `defaultPolicy` function and not `Before`. Then add separate tests for `Before`

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -18,30 +18,28 @@ import (
...
 
 // Plugin implements Fetch Metadata functionality.
 // See https://www.w3.org/TR/fetch-metadata/ for more details.
 type Plugin struct {
-	mode   string
-	policy func(*safehttp.IncomingRequest) bool
-	logger LoggingService
+	// Policy is the Fetch Metadata policy that decides whether a request should
+	// be allowed to pass or be rejected. This will be set to a default Resource
+	// Isolation Policy by default but can be changed by the user.
```
mattiasgrenfeldt, line 32: `... set to a default ... by default ...`
Rephrase to remove one default

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -18,30 +18,28 @@ import (
...
+	Policy     func(*safehttp.IncomingRequest) bool
+	reportOnly bool
+	logger     LoggingService
 }
 
 // NewPlugin creates a new Fetch Metadata plugin using the defaultPolicy and
 // sets the mode to enforce.
```
mattiasgrenfeldt: `defaultPolicy` is not visible in godoc to the users who will use the library. Explain the policy in some other way.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -108,14 +101,13 @@ func (p *Plugin) SetEnforceMode() {
...
-		switch p.mode {
-		case enforceMode:
+	if !p.Policy(r) {
+		switch p.reportOnly {
+		case false:
 			return w.ClientError(safehttp.StatusForbidden)
-		case reportMode:
+		case true:
 			p.logger.Log(r)
 		}
```
mattiasgrenfeldt: This should be an if-statement

safehttp/plugins/fetch_metadata/fetch_metadata.go
mattiasgrenfeldt: Right, so you could have only the `logger` and remove `reportOnly`. And if `logger` is `nil` then enforcement is on and if it isn't `nil` then it's reporting mode.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,113 @@
...
+		return true
+	case "same-origin", "same-site", "none":
+		// The request originated from a site served by your own server
+		// ("same-origin"), a subdomain of your site ("same-site", e.g.
+		// bar.foo.com made a request to foo.com) or was caused by the user
+		// explicitly interacting with the user-agent ("none"). Therefore it is
+		// allowed to pass.
+		return true
+	}
+	if m := r.Method(); h.Get("Sec-Fetch-Mode") == "navigate" && (m == "GET" || m == "HEAD") {
```
mattiasgrenfeldt, line 65: Replace `"GET"` with `safehttp.MethodGet`. Same with all other methods

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,411 @@
...
+	p.SetReportMode(nil)
+}
+
+type fooLog struct {
+	report string
+}
+
+func (l *fooLog) Log(r *safehttp.IncomingRequest) {
+	l.report = r.Method() + " " + r.URL.Path()
+}
```
mattiasgrenfeldt, line 302: Move this above `TestReportModeWithLogger` since that is where it is first used.

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,411 @@
...
+	logger := &fooLog{}
+	p := fetchmetadata.NewPlugin()
+	req := safehttptest.NewRequest("POST", "/pizza", nil)
+	req.Header.Add("Sec-Fetch-Site", "cross-site")
+	req.Header.Add("Sec-Fetch-Mode", "navigate")
+	rec := safehttptest.NewResponseRecorder()
+
+	p.Before(rec.ResponseWriter, req)
+
+	if status := rec.Status(); status != 403 {
```
mattiasgrenfeldt, line 314: Replace `403` with `safehttp.StatusForbidden`. Same with all other statuses

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,411 @@
...
+type fooLog struct {
+	report string
+}
+
+func (l *fooLog) Log(r *safehttp.IncomingRequest) {
+	l.report = r.Method() + " " + r.URL.Path()
+}
+
+func TestChangeMode(t *testing.T) {
+	logger := &fooLog{}
```
mattiasgrenfeldt, line 305: Move this to right above `p.SetReportMode(logger)` since that is where it is first used.

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,411 @@
...
+		t.Errorf("status code got: %v want: %v", status, 403)
+	}
+	wantHeaders := map[string][]string{
+		"Content-Type":           {"text/plain; charset=utf-8"},
+		"X-Content-Type-Options": {"nosniff"},
+	}
+	if diff := cmp.Diff(wantHeaders, map[string][]string(rec.Header())); diff != "" {
+		t.Errorf("rec.Header() mismatch (-want +got):\n%s", diff)
+	}
+
```
mattiasgrenfeldt, line 324: nit: remove empty line

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,411 @@
...
+	if wantBody, got := "Forbidden\n", rec.Body(); got != wantBody {
```
mattiasgrenfeldt, line 325: nit: renamed `wantBody` to just `want`

mattiasgrenfeldt (changes requested):
I accidentally submitted my review when I wasn't done. So here are some more comments.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,155 @@
...
+		// Fetch Metadata is not supported by the browser so we allow the
+		// request to pass.
+		return true
+	case "same-origin", "same-site", "none":
+		// The request originated from a site served by your own server
+		// ("same-origin"), a subdomain of your site ("same-site", e.g.
+		// bar.foo.com made a request to foo.com) or was caused by the user
+		// explicitly interacting with the user-agent ("none"). Therefore it is
+		// allowed to pass.
+		return true
```
empijei, line 76: I would say to either separate "none" from the "same-*" block or to put all of them together.

As a general idea I would say that the simplest and most readable way to express this entire switch is

```go
if h.Get("Sec-Fetch-Site") != "cross-site" {
  // Here comment what we are handling here.
  return true
}
```

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	if m := r.Method(); h.Get("Sec-Fetch-Mode") == "navigate" && (m == safehttp.MethodGet || m == safehttp.MethodHead) {
+		if dest := h.Get("Sec-Fetch-Dest"); dest == "object" || dest == "embed" {
```
empijei, line 80: These two lines are hard to read. I suggest splitting them up and adding comments:

```go
	if method, mode, dest := r.Method(), h.Get("Sec-Fetch-Mode"), h.Get("Sec-Fetch-Dest"); 
		// Allow navigations
		(mode == "navigate" || mode == "nested-navigate") &&
		// On non-state-changing methods
		(method == safehttp.MethodGet || method == safehttp.MethodHead) &&
		// That are not potentially dangerous
		(dest != "object" && dest != "embed") {
			return true
		}
```

That said I think it would even better to have some of this logic factored out of the `if`.

Our internal policy looks more or less like this:
```go
    // (Omitted) read and assign the 3 sec-fetch headers and the method

    if site != "cross-site" {
      // Allow requests from browsers which don't send Fetch Metadata
      // and same-site, same-origin, and browser-initiated requests.
      return true;
    }

    // Allow CORS options requests if neither Mode nor Dest is set.
    // https://github.com/w3c/webappsec-fetch-metadata/issues/35
    // https://bugs.chromium.org/p/chromium/issues/detail?id=979946
    if mode == "" && dest == "" && method == "OPTIONS" {
      return true;
    }

    // Allow navigations from anywhere for non state changing requests.
    // Don't fail-open if mode is not set as e.g. safe browsing doesn't set Fetch Metadata mode.
    if navigationalModes[mode] && navigationalDests[dest] && statePreservingMethods[method] {
      return true;
    }

    return false;
```
Where the three conditions of the last `if` are accesses to maps of type `map[string]bool`

I think this would be a readability boost.

In general if you have a very complicated condition to express I strongly advise to have all nested `return` statements to return one value (`true` in this case) and the last one to return the opposite. It makes it easier to debug.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,155 @@
...
+		// allow it to pass.
+		return true
+	}
+
+	origin := h.Get("Origin")
+	if _, ok := allowedCORS[origin]; ok && h.Get("Sec-Fetch-Mode") == "cors" {
+		// The request is cross-site but sent from an
+		// origin where Cross-Origin Resource Sharing is explicitly allowed.
+		return true
+	}
```
empijei, line 95: I think this check should be the first thing we do: if an endpoint is CORS-allowed we should bail out without doing any checks.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,155 @@
...
+
+func navigationIsolationPolicy(r *safehttp.IncomingRequest, allowedCORS map[string]bool) bool {
+	h := r.Header
+	origin := h.Get("Origin")
+	if h.Get("Sec-Fetch-Site") == "cross-site" && h.Get("Sec-Fetch-Mode") == "navigate" {
+		if _, ok := allowedCORS[origin]; !ok {
+			return false
+		}
+	}
+	return true
```
empijei, line 110: The cors-allowed check should be the first one.

With regards to the policy you expressed, I don't think this is what we should go for.

Since this policy has to run on top of the RI one I would say this should be something like:
```go
// Omitted: check if CORS allowed

if site == "cross-site" && navigationalMode[mode] {
  return false
}
return resourceIsolationPolicy(r, allowedCORS)
```

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// SetReportMode sets the Fetch Metadata policy mode to "report". This will
+// allow requests that violate the policy to pass, but will report the violation
+// using the LoggingService. The method will panic if no LoggingService is
+// provided.
+func (p *Plugin) SetReportMode(logger LoggingService) {
```
empijei, line 117: SetReportOnly

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,155 @@
...
+	if logger == nil {
+		panic("logging service required for Fetch Metadata report mode")
+	}
+	p.logger = logger
+	p.reportOnly = true
+}
+
+// SetEnforceMode sets the Fetch Metadata policy mode to "enforce". This will
+// reject any requests that violate the policy provided by the plugin.
+func (p *Plugin) SetEnforceMode() {
```
empijei, line 127: SetEnforce.

Also, I would set the logger to nil here.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,155 @@
...
+		return safehttp.Result{}
+	}
+	if p.NavIsolation && !navigationIsolationPolicy(r, p.allowedCORS) {
+		if !p.reportOnly {
+			return w.ClientError(safehttp.StatusForbidden)
+		}
+		p.logger.Log(r)
+		return safehttp.Result{}
+	}
+	return safehttp.Result{}
```
empijei, line 154: A lot of these conditions are negated, and most of the important logic is indented. Moreover there is quite a bit of duplication in structure.

Can you try to refactor this to make it more linear to read?

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,633 @@
...
+	tests := []struct {
+		name    string
+		origins map[string]bool
+		req     *safehttp.IncomingRequest
+		allowed bool
+	}{
+		{
+			name:    `Sec-Fetch-Site: ""`,
+			origins: map[string]bool{},
+			req:     safehttptest.NewRequest("POST", "https://spaghetti.com/pizza", nil),
```
empijei, line 35: WHAT DO YOU MEAN BY SPAGHETTI PIZZA‽

empijei (changes requested):

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
mattiasgrenfeldt: 🍝🍕👌

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,186 @@
...
+		corsProtected: m,
+	}
+}
+
+func (p *Plugin) resourceIsolationPolicy(r *safehttp.IncomingRequest) bool {
+	h := r.Header
+	method := r.Method()
+	site := h.Get("Sec-Fetch-Site")
+	mode := h.Get("Sec-Fetch-Mode")
+	dest := h.Get("Sec-Fetch-Dest")
```
mattiasgrenfeldt, line 101: Move all variable declarations directly before they are used and remove all variables that are only used once. Instead just put the function call directly in their place.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// Allow CORS options requests if neither Mode nor Dest is set.
+	// https://github.com/w3c/webappsec-fetch-metadata/issues/35
+	// https://bugs.chromium.org/p/chromium/issues/detail?id=979946
+	if mode == "" && dest == "" && method == safehttp.MethodOptions {
+		return true
+	}
+
+	if navigationalModes[mode] && navigationalDest[dest] && statePreservingMethods[method] {
+		// The rquest is cross-site, but a simple top-level navigation from a
```
mattiasgrenfeldt, line 119: The request

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,399 @@
...
+
+import (
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-safeweb/safehttp"
+	"github.com/google/go-safeweb/safehttp/plugins/fetch_metadata"
+	"github.com/google/go-safeweb/safehttp/safehttptest"
+	"testing"
+)
+
+func TestResourceIsolationEnforceMode(t *testing.T) {
```
mattiasgrenfeldt, line 25: Split this into two table tests. One for requests that pass and one for requests that fail. That would be easier to read.

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,399 @@
...
+			wantStatus:  safehttp.StatusOK,
+			wantHeaders: map[string][]string{},
+			wantBody:    "",
+		},
+		{
+			name: "Sec-Fetch-Site: cross-site non-navigational",
+			req: func() *safehttp.IncomingRequest {
+				req := safehttptest.NewRequest("POST", "/", nil)
+				req.Header.Add("Sec-Fetch-Site", "cross-site")
+				req.Header.Add("Sec-Fetch-Site", "no-cors")
```
mattiasgrenfeldt, line 78: Should both of these be `Sec-Fetch-Site`?

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,399 @@
...
+	l.report = r.Method() + " " + r.URL.Path()
+}
+
+func TestResourceIsolationReportMode(t *testing.T) {
+	tests := []struct {
+		name        string
+		req         *safehttp.IncomingRequest
+		wantStatus  safehttp.StatusCode
+		wantHeaders map[string][]string
+		wantBody    string
```
mattiasgrenfeldt, line 184: All of these tests have the same value for these parameters. Hardcode them in the subtest and remove them from the table.

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
mattiasgrenfeldt: This is not resolved

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,399 @@
...
+	}
+	if want, got := "", rec.Body(); got != want {
+		t.Errorf("response body got: %q want: %q", got, want)
+	}
+	if want := "POST /pizza"; logger.report != want {
+		t.Errorf("logger.report: got %s, want %s", logger.report, want)
+	}
+}
+
+func TestEnableDisableNavIsolation(t *testing.T) {
```
mattiasgrenfeldt, line 336: In a real use case we would never get a request and then change from `NavIsolation` to not `NavIsolation` between requests. The plugins will be configured at start up and then run with the same configuration for the entire runtime of the server. So make a test that just has `NavIsolation` on from the start and don't change it I would say. Same principle with `TestChangeEnforceReportMode`.

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,399 @@
...
+		t.Errorf("rec.Header() mismatch (-want +got):\n%s", diff)
+	}
+	if want, got := "", rec.Body(); got != want {
+		t.Errorf("response body got: %q want: %q", got, want)
+	}
+}
+
+func TestCORSEndpoint(t *testing.T) {
+	methods := []string{"GET", "POST"}
+	for _, m := range methods {
```
mattiasgrenfeldt, line 379: This is a table test. So use `t.Run(...)` in each iteration of the loop.

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,399 @@
...
+		req := safehttptest.NewRequest(m, "https://spaghetti.com/pizza", nil)
+		req.Header.Add("Sec-Fetch-Site", "cross-site")
+		req.Header.Add("Sec-Fetch-Mode", "navigate")
+		req.Header.Add("Sec-Fetch-Dest", "image")
+		rec := safehttptest.NewResponseRecorder()
+		p := fetchmetadata.NewPlugin("https://spaghetti.com/pizza")
+		p.NavIsolation = true
```
mattiasgrenfeldt, line 386: I think this line shouldn't be here probably.

mattiasgrenfeldt (changes requested):

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,155 @@
...
+	// Isolation Policy, but can be changed by the user.
+	//
+	// See https://web.dev/fetch-metadata/ for more information on the default policy
+	Policy func(*safehttp.IncomingRequest, map[string]bool) bool
+	// NavIsolation indicates whether the Navigation Isolation Policy should
+	// also be applied to the request as an additional layer of checking. This
+	// provides a way to mitigate clickjacking and reflected XSS by rejecting
+	// all cross-site navigations unless coming from origins where CORS is
+	// explicitly enabled. This can be used independently from a Resource
+	// Isolation Policy.
```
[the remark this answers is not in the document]
mihalimara22: In the new version, Navigation Isolation is applied first, if enabled and then Resource Isolation.

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
mihalimara22: It should be because one of the requests breaks the navigation isolation policy and the other one the resource isolation policy. I configured the plugin in the most strict way.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,186 @@
...
+	// additional layer of checking. This provides a way to mitigate
+	// clickjacking and reflected XSS by rejecting all cross-site navigations
+	// unless targeted to endpoints that are CORS-protected.
+	//
+	// WARNING: This is still an experimental feature and will be disabled by
+	// default.
+	NavIsolation bool
+	// RedirectURL can optionally indicate an URL to which the user can be
+	// redirected in case the Navigation Isolation policy rejects the request.
+	RedirectURL   string
```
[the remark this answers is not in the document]
mihalimara22: Good shout!

safehttp/plugins/fetch_metadata/fetch_metadata.go
mihalimara22: I'd rather keep explicit functions to switch between report and enforce mode with explanations on what the user is about to do. :)

safehttp/plugins/fetch_metadata/fetch_metadata.go
mattiasgrenfeldt: The way you did it now looks good!

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,407 @@
...
+	tests := allowedRIPHeaders
+	for _, test := range allowedRIPNavHeaders {
+		tests = append(tests, testHeaders{
+			name:   test.name,
+			method: test.method,
+			site:   test.site,
+			mode:   test.mode,
+			dest:   test.dest,
+		})
+	}
```
mattiasgrenfeldt, line 170: You can append two slices using `append` like this:
```go
x = append(slice1, slice2...)
```

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+type fooLog struct {
+	report string
+}
+
+func (l *fooLog) Log(r *safehttp.IncomingRequest) {
+	l.report = r.Method()
+}
+func TestAllowedResourceIsolationReportMode(t *testing.T) {
```
mattiasgrenfeldt, line 241: You have `TestAllowedResourceIsolationReportMode` and `TestAllowedResourceIsolationEnforceMode`. But having `ReportOnly` turned on doesn't make a difference for requests that are allowed to pass. It only changes the behavior for requests that are disallowed. Therefore I would say you could remove one of these table tests without losing any test coverage.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,158 @@
...
+		// policies are disabled because it is CORS-protected so we don't apply
+		// the policies.
+		return safehttp.Result{}
+	}
+
+	reject := false
+	h := r.Header
+	if p.NavIsolation && h.Get("Sec-Fetch-Site") == "cross-site" && navigationalModes[h.Get("Sec-Fetch-Mode")] {
+		if p.RedirectURL != nil {
+			return w.Redirect(r, p.RedirectURL.String(), safehttp.StatusMovedPermanently)
```
mattiasgrenfeldt, line 146: You need a test that reaches this line as well

safehttp/url_test.go
```diff
@@ -203,3 +203,40 @@ func TestURLInvalidQuery(t *testing.T) {
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			url, err := Parse(test.url)
+			if url == nil {
+				t.Error("Parse(test.url): got nil, want URL")
+			}
+			if err != nil {
+				t.Errorf("Parse(test.url): got %v, want nil", err)
+			}
```
mattiasgrenfeldt, line 229: Add another if-statement that checks if `url.String() == test.url`

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,186 @@
...
+// Policy and, if enabled, the Navigation Isolation Policy . It only allows
+// request to pass if they conform to the policy, if it's targeted to a
+// CORS-protected endpoint, or if the mode is set to
+// "report", in which case the request is allowed to pass but the violation is
+// reported. If a redirectURL was provided and the Navigation Isolation Policy
+// fails, the IncomingRequest will be redirected to the redirectURL.
+func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+	reject := false
+	endpoint := r.URL.String()
+	if _, ok := p.corsProtected[endpoint]; ok {
```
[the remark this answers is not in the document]
mattiasgrenfeldt: Don't forget to fix this

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
mattiasgrenfeldt: Ah, I see, nice! :)

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,399 @@
...
+
+type fooLog struct {
+	report string
+}
+
+func (l *fooLog) Log(r *safehttp.IncomingRequest) {
+	l.report = r.Method() + " " + r.URL.Path()
+}
+
+func TestResourceIsolationReportMode(t *testing.T) {
```
empijei: please add a test for navigation isolation in report mode.

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
mihalimara22: Oh, nice, thanks :)

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
mihalimara22: I still want to make sure that nothing was written to `logger.report`

safehttp/plugins/fetch_metadata/fetch_metadata.go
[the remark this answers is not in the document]
empijei: Question: wouldn't URL contain also the domain here? Shouldn't "endpoint" be path-based?

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,158 @@
...
+// limitations under the License.
+
+package fetchmetadata
+
+import (
+	"github.com/google/go-safeweb/safehttp"
+)
+
+// RequestLogger is a user-provided service for logging Fetch Metadata policy
+// violations
```
kele, line 22: nit: missing dot at the and, and it's enough to say: `RequestLogger is used for logging Fetch Metadata policy violations`. "user-provided" doesn't add any meaning here.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,158 @@
...
+	// Allow CORS options requests if neither Mode nor Dest is set.
+	// https://github.com/w3c/webappsec-fetch-metadata/issues/35
+	// https://bugs.chromium.org/p/chromium/issues/detail?id=979946
+	if mode == "" && dest == "" && method == safehttp.MethodOptions {
+		return true
+	}
+
+	if navigationalModes[mode] && navigationalDest[dest] && statePreservingMethods[method] {
+		// The request is cross-site, but a simple top-level navigation from a
+		// safe destination so we  allow it to pass .
```
kele, line 99: nit: extra space

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,158 @@
...
+// will be allowed by default without the policies being applied.
+func NewPlugin(endpoints ...string) *Plugin {
+	m := map[string]bool{}
+	for _, e := range endpoints {
+		m[e] = true
+	}
+	return &Plugin{corsProtected: m}
+}
+
+func (p *Plugin) resourceIsolationPolicy(r *safehttp.IncomingRequest) bool {
```
kele, line 76: Could you come up with a better name for this function?

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,158 @@
...
+// Policy and, if enabled, the Navigation Isolation Policy. It only allows
+// requests to pass if they conform to the policy, if it's targeted to one of
+// the CORS-protected endpoints, specified when creating the plugin, or if the
+// mode is set to  "report", in which case the request is allowed to pass but
+// the  violation is reported. If a redirectURL was provided and the Navigation
+// Isolation Policy is enabled and fails, the IncomingRequest will be
+// redirected to redirectURL.
+func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+	endpoint := r.URL.String()
+	if _, ok := p.corsProtected[endpoint]; ok {
```
kele, line 135: You're using booleans there anyway, so use the fact that map will return a default value (false), i.e. `if p.CorsProtected[endpoint] {`.

safehttp/url.go
```diff
@@ -71,3 +73,17 @@ func (u URL) Port() string {
 func (u URL) Path() string {
 	return u.url.Path
 }
+
+// Parse parses a rawurl string into a URL structure.
+//
+// The rawurl may be relative (a path, without a host) or absolute
```
kele, line 79: rawURL

safehttp/url.go
```diff
@@ -71,3 +73,17 @@ func (u URL) Port() string {
...
+// (starting with a scheme). Trying to parse a hostname and path
+// without a scheme is invalid but may not necessarily return an
+// error, due to parsing ambiguities.
```
kele, line 82: nit: The line breaking here seems a bit excessive. It's not even wrapped at 80 lines but at less.

kele (changes requested):

safehttp/plugins/fetch_metadata/fetch_metadata.go
kele: Sure, but maybe then `checkResourceIsolationPolicy`?

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	navigationalDest = map[string]bool{
+		"document":        true,
+		"nested-document": true,
+	}
+	statePreservingMethods = map[string]bool{
+		safehttp.MethodGet:  true,
+		safehttp.MethodHead: true,
+	}
+)
```
kele, line 39: Nice and clean! :)

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,163 @@
...
+// See https://www.w3.org/TR/fetch-metadata/ and
+// https://web.dev/fetch-metadata/  for more details.
+type Plugin struct {
+	// NavIsolation indicates whether the Navigation Isolation Policy should
+	// be applied to the request before the Resource Isolation Policy as an
+	// additional layer of checking. This provides a way to mitigate
+	// clickjacking and reflected XSS by rejecting all cross-site navigations
+	// unless targeted to endpoints that are CORS-protected.
+	//
+	// WARNING: This is still an experimental feature and will be disabled by
```
kele, line 52:
```suggestion
	// WARNING: This is still an experimental feature and is disabled by
```

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,163 @@
...
+// the  violation is reported. If a redirectURL was provided and the Navigation
+// Isolation Policy is enabled and fails, the IncomingRequest will be
+// redirected to redirectURL.
+func (p *Plugin) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+	if p.corsProtected[r.URL.Path()] {
+		// The request is targeted to an endpoint on which Fetch Metadata
+		// policies are disabled because it is CORS-protected so we don't apply
+		// the policies.
+		return safehttp.Result{}
+	}
```
kele, line 142: Hm... Why is this information stored in a plugin at all?

Shouldn't this work like:

```
m := Mux()
m.Install(fetchmetadata.Plugin)
m.Handle(..., fetchmetadata.Disable())
```

?

kele (changes requested):

safehttp/plugins/fetch_metadata/fetch_metadata.go
mihalimara22: This will then be handled and tested after the per-mux configurations have been finalised and we have an orchestrator.

kele: Good call.

safehttp/plugins/fetch_metadata/fetch_metadata.go
```diff
@@ -0,0 +1,165 @@
...
+func (p *Plugin) checkNavigationIsolationPolicy(r *safehttp.IncomingRequest) bool {
+	if h := r.Header; p.NavIsolation && h.Get("Sec-Fetch-Site") == "cross-site" && navigationalModes[h.Get("Sec-Fetch-Mode")] {
+		return false
+	}
+	return true
+}
+
+// SetReportOnly sets the Fetch Metadata policy mode to "report". This will
+// allow requests that violate the policy to pass, but will report the violation
+// using the RequestLogger. The method will panic if RequestLogger is nil.
```
kele, line 114: `if Logger is nil`. There is no `RequestLogger` field.

safehttp/url.go
```diff
@@ -66,8 +68,21 @@ func (u URL) Port() string {
...
 func (u URL) Path() string {
 	return u.url.Path
 }
+
+// Parse parses a rawURL string into a URL structure.
+//
+// The rawURl may be relative (a path, without a host) or absolute (starting
+// with a scheme). Trying to parse a hostname and path without a scheme is
+// invalid but may not necessarily return an error, due to parsing ambiguities.
+func Parse(rawurl string) (*URL, error) {
```
kele, line 82: This will be used as `safehttp.Parse`. Maybe change it to `safehttp.ParseURL`?

safehttp/url_test.go
```diff
@@ -203,3 +203,43 @@ func TestURLInvalidQuery(t *testing.T) {
...
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			url, err := Parse(test.url)
+			if url == nil {
+				t.Error("Parse(test.url): got nil, want URL")
+			}
+			if err != nil {
+				t.Errorf("Parse(test.url): got %v, want nil", err)
```
kele, line 228: Use "got err %v" instead.

kele (changes requested):

safehttp/url.go
mihalimara22: Would definitely be more intuitive for the user :)

safehttp/url.go
```diff
@@ -66,8 +68,21 @@ func (u URL) Port() string {
 //
 // Note that the path is stored in decoded form: /%47%6f%2f
 // becomes /Go/. A consequence is that it is impossible to tell
-// which slashes in the path were slashes in the raw URL and which
+// which slashes in the path were slashes in the rawURL and which
```
mattiasgrenfeldt, line 71: Why this change? Do we have a `rawURL` field? I think `raw URL` was better.

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,521 @@
...
+			p.NavIsolation = true
+			p.SetReportOnly()
+			p.Before(rec.ResponseWriter, req)
+
+			if want, got := safehttp.StatusOK, safehttp.StatusCode(rec.Status()); want != got {
+				t.Errorf("status code got: %v want: %v", got, want)
+			}
+			if want, got := safehttp.StatusOK, safehttp.StatusCode(rec.Status()); got != want {
+				t.Errorf("status code got: %v want: %v", got, want)
+			}
```
mattiasgrenfeldt, line 414: I think this got duplicated

safehttp/plugins/fetch_metadata/fetch_metadata_test.go
```diff
@@ -0,0 +1,521 @@
...
+			p.Logger = logger
+			p.SetReportOnly()
+			p.Before(rec.ResponseWriter, req)
+
+			if want, got := safehttp.StatusOK, safehttp.StatusCode(rec.Status()); want != got {
+				t.Errorf("status code got: %v want: %v", got, want)
+			}
+			if want, got := safehttp.StatusOK, safehttp.StatusCode(rec.Status()); got != want {
+				t.Errorf("status code got: %v want: %v", got, want)
+			}
```
mattiasgrenfeldt, line 314: Here it is also duplicated

mattiasgrenfeldt (changes requested):

kele (approved):
LGTM, please resolve Mattias' comments.

mattiasgrenfeldt (commented):
Please remove the `safehttp/plugins/fetch_metadata/debug.test` file.

Otherwise looks good to me!

mattiasgrenfeldt (approved):
Right, forgot to click the Approve checkbox

PR #98, opened 2020-08-19
Make sure that ResponseWriter is only written to once, otherwise panic.

### LINKED ISSUE #97: Prevent double-writing to the ResponseWriter
This is one of the most critical security features.

Double-writing should likely result in panic.

mattiasgrenfeldt:
Fixes #97 

To be able to store the state whether or not the `ResponseWriter` was written to or not the signature of `Handler` and `Interceptor.Before` had to change to accept a `*ResponseWriter` instead of a `ResponseWriter`. This needed to be changed in **a lot** of places.

### DIFF
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -178,7 +178,8 @@ func (m methodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}()
 
 	for _, intercep := range m.muxInterceps {
-		if res := intercep.Before(rw, ir); res.written {
+		intercep.Before(rw, ir)
+		if rw.written {
 			return
 		}
 	}
--- a/safehttp/response_writer.go
+++ b/safehttp/response_writer.go
@@ -27,6 +27,7 @@ type ResponseWriter struct {
 	// easily overwrite the struct bypassing all our safety guarantees.
 	header       Header
 	muxInterceps map[string]Interceptor
+	written      bool
 }
 
 // NewResponseWriter creates a ResponseWriter from a safehttp.Dispatcher, an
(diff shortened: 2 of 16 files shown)


safehttp/mux.go
```diff
@@ -165,7 +165,8 @@ func (m methodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	for _, intercep := range m.muxInterceps {
-		if res := intercep.Before(rw, ir); res.written {
+		intercep.Before(rw, ir)
+		if rw.written {
 			return
```
kele, line 170: Hm... What if other interceptors don't write? Should they run?

Also, what if the handler doesn't write?

kele (changes requested):

safehttp/mux.go
mattiasgrenfeldt: A handler that doesn't write any response seems rather pointless. Maybe we should panic if nothing has been written when the handler is done executing?

If we assume that `Before` will only write something when there is some sort of error, then I don't see the point of running more `Before` methods when we already know the response is an error.

I can't come up with any `Before`-phases that might want to write something that isn't an error. But if they exist then we might want to continue running other `Before`-phases. If in that case a later `Before` method also writes a response then it's OK if it panics since panics will be recovered and a 500 Internal Server Error will be returned.

So should I remove the
```go
if rw.written {
    return
}
```
?

mattiasgrenfeldt:
Rebased on top of master so that tests don't fail

safehttp/response_writer_test.go
```diff
@@ -50,3 +53,57 @@ func TestResponseWriterSetInvalidCookie(t *testing.T) {
...
 	}
 }
+
+func TestResponseWriterWriteTwicePanic(t *testing.T) {
+	tests := []struct {
+		name  string
+		write func(w *safehttp.ResponseWriter)
+	}{
+		{
+			name: "Write",
```
mihalimara22, line 63: Have a more detailed name to give more context to the reader.

safehttp/response_writer.go
```diff
@@ -51,52 +52,66 @@ func (w *ResponseWriter) Interceptor(key string) Interceptor {
...
+	}
 	http.Redirect(w.rw, r.req, url, int(code))
-	return Result{written: true}
+	return Result{}
+}
+
+// markWritten sets written to true. If written was already true, it panics.
+func (w *ResponseWriter) markWritten() {
+	if w.written {
+		panic("ResponseWriter already written to")
```
mihalimara22, line 112: Missing was

safehttp/response_writer.go
```diff
@@ -51,52 +52,66 @@ func (w *ResponseWriter) Interceptor(key string) Interceptor {
...
+	w.markWritten()
+	if code < 300 || code >= 400 {
+		panic("wrong method called")
+	}
 	http.Redirect(w.rw, r.req, url, int(code))
-	return Result{written: true}
+	return Result{}
+}
+
+// markWritten sets written to true. If written was already true, it panics.
```
mihalimara22, line 109: Have a comment that describes how the function is useful not its exact implementation.

mihalimara22 (changes requested):

safehttp/response_writer.go
```diff
@@ -51,52 +52,67 @@ func (w *ResponseWriter) Interceptor(key string) Interceptor {
...
-	return Result{written: true}
+	return Result{}
 }
 
 // Redirect responds with a redirect to a given url, using code as the status code.
 func (w *ResponseWriter) Redirect(r *IncomingRequest, url string, code StatusCode) Result {
+	w.markWritten()
+	if code < 300 || code >= 400 {
+		panic("wrong method called")
+	}
```
empijei, line 104: Maybe we should have a type for each family of statuses?

Like 

type RedirectStatus Status
type ServerErrorStatus Status

So that we can check with the typesystem that some things don't happen.

@kele WDYT?

kele: I'm not sure this is a good enough reason.

safehttp/response_writer_test.go
```diff
@@ -50,3 +53,57 @@ func TestResponseWriterSetInvalidCookie(t *testing.T) {
...
+
+func TestResponseWriterWriteTwicePanic(t *testing.T) {
+	tests := []struct {
+		name  string
+		write func(w *safehttp.ResponseWriter)
+	}{
+		{
+			name: "Call Write twice",
+			write: func(w *safehttp.ResponseWriter) {
+				w.Write(safehtml.HTMLEscaped("<h1>Escaped, so not really a heading</h1>"))
```
kele, line 65: Suggestion: make `write` be the full "handler", i.e. in this example, do `w.Write()` twice.

Why? This way you can also add tests for "400, then 200" cases (i.e. `w.ServerError(); w.Write()`). Also, it feels somewhat surprising to see the description say "Call Write twice" and then a line below see it called once.

kele (approved):

safehttp/response_writer_test.go
mattiasgrenfeldt: I agree it is a little confusing, but I want to be able to defer the recover between the writes, so that it only detects if the seconds write panics. If I make the whole "handler" part of the table then the only part left in the loop is creating the `ServeMux`. Like this:
-Handler: Write, Defer, Write.
-Loop: Create ServeMux, call handler.
In that case it would make more sense to just create a separate test for each case and ditch the table test since the tests won't share any logic anymore.

I could have a `write1` and a `write2` in the table.

And should I add all the 5*5 combinations of ways to write twice?

kele: It's fine if you do the recover at the beginning. We have other tests to check that things don't panic on a first write.

kele (approved):

kele:
@mihalimara22 PTAL since you requested changes.

@mattiasgrenfeldt please rebase to resolve conflicts.

mihalimara22 (approved):

kele:
Needs rebasing, then I can merge.

mattiasgrenfeldt:
Done! @kele

PR #104, opened 2020-08-21
Added first version of CORS plugin

### LINKED ISSUE #96: Add a CORS plugin
On CORS enabled endpoints this plugin should handle the incoming headers:

- Origin
- Access-Control-Request-Method
- Access-Control-Request-Headers

And set the response headers accordingly:

- Access-Control-Allow-Origin
- Access-Control-Expose-Headers
- Access-Control-Max-Age
- Access-Control-Allow-Credentials
- Access-Control-Allow-Methods
- Access-Control-Allow-Headers

See https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS for more info

mattiasgrenfeldt:
A CORS interceptor is implemented that handles preflights and applies the specified CORS rules on incoming requests.

The request methods `GET`, `HEAD` and `POST` are banned.
The Content-Types `application/x-www-form-urlencoded`, `multipart/form-data` and `text/plain` are also banned.
Each request must also contain either `X-Cors: 1` or `Sec-Fetch-Mode: cors`. All of this is to prevent XSRF on CORS requests.

Fixes #96


safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+var disallowedContentTypes = map[string]bool{
+	"application/x-www-form-urlencoded": true,
+	"multipart/form-data":               true,
+	"text/plain":                        true,
+}
+
+// Interceptor handles CORS requests based on its settings. The request methods
+// GET, HEAD and POST are banned and will be responded to with 405 Method Not Allowed.
```
mihalimara22, line 40: "will result in a 405 Method Not Allowed response." and similarly below

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,203 @@
...
+// SetAllowedHeaders sets which headers should be allowed in the Access-Control-Allow-Headers
+// header. The headers are first canonicalized using textproto.CanonicalMIMEHeaderKey.
+func (it *Interceptor) SetAllowedHeaders(headers ...string) {
+	it.allowedHeaders = map[string]bool{}
+	for _, h := range headers {
+		it.allowedHeaders[textproto.CanonicalMIMEHeaderKey(h)] = true
+	}
+}
+
+// Before handle the incoming request and sets headers or responds with an error
```
mihalimara22, line 90: handles

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,203 @@
...
+		return it.preflight(setHeaders, w, r)
+	} else {
+		return it.actualRequest(setHeaders, w, r)
+	}
+}
+
+// preflight handles requests that have the method OPTIONS.
+func (it *Interceptor) preflight(setHeaders func(), w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+	rh := r.Header
+	method := rh.Get("Access-Control-Request-Method")
```
mihalimara22, line 136: Can't this be `""` as well?

safehttp/plugins/cors/cors_test.go
```diff
@@ -0,0 +1,603 @@
...
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-safeweb/safehttp"
+	"github.com/google/go-safeweb/safehttp/plugins/cors"
+	"github.com/google/go-safeweb/safehttp/safehttptest"
+)
+
+func TestActualRequest(t *testing.T) {
```
mihalimara22, line 26: I don't think `Actual` is needed :)

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,203 @@
...
+
+import (
+	"net/textproto"
+	"strconv"
+	"strings"
+
+	"github.com/google/go-safeweb/safehttp"
+)
+
+const customHeader string = "X-Cors"
```
mihalimara22, line 25: Can you explain this decision to me? :)

safehttp/plugins/cors/cors_test.go
```diff
@@ -0,0 +1,603 @@
...
+				t.Errorf("rr.Header() mismatch (-want +got):\n%s", diff)
+			}
+			if got := rr.Body(); got != "" {
+				t.Errorf(`rr.Body() got: %q want: ""`, got)
+			}
+		})
+	}
+}
+
+func TestActualRequestNoSFMOrXCorsHeader(t *testing.T) {
```
mihalimara22, line 170: Same here with `Actual`

mihalimara22 (changes requested):

safehttp/plugins/cors/cors.go
mattiasgrenfeldt: It is pretty arbitrary. @empijei suggested `Cors`. Then someone in the mail-thread mentioned that it should be a `X-` header, so I picked `X-Cors`

safehttp/plugins/cors/cors.go
mattiasgrenfeldt: Yes, it can be empty. But a browser will always set it automatically. So if it is empty then someone is sending requests manually to our server in which case CORS is pointless. (Unless there is a way to send a CORS request without specifying the header, which would be a bug in browsers.)

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,208 @@
...
+
+	// Used to set headers in preflight() or request() if no error occurs.
+	setHeaders := func() {
+		allowOrigin([]string{origin})
+		vary([]string{"Origin"})
+		if r.Header.Get("Cookie") != "" && it.AllowCredentials {
+			// TODO: handle other credentials than cookies:
+			// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials
+			allowCredentials([]string{"true"})
+		}
```
empijei, line 128: Are you sure this is how it works?

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+}
+
+// Before handles the incoming request and sets headers or responds with an error
+// accordingly.
+func (it *Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+	origin := r.Header.Get("Origin")
+	if !it.AllowedOrigins[origin] {
+		return w.ClientError(safehttp.StatusForbidden)
+	}
```
empijei, line 101: Does same-origin `fetch` add a `Origin` header? I think it might not. I think Origin might only be sent when cross-origin.

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,208 @@
...
+	allowCredentials, err := h.Claim("Access-Control-Allow-Credentials")
+	if err != nil {
+		return w.ServerError(safehttp.StatusInternalServerError)
+	}
+	if r.Header.Get("Cookie") != "" && !it.AllowCredentials {
+		return w.ClientError(safehttp.StatusForbidden)
+	}
+
+	// Used to set headers in preflight() or request() if no error occurs.
+	setHeaders := func() {
```
empijei, line 121: why does this need to be a closure? Can't you pass the stuff you need in a different way? After the `setHeaders` call in both functions you pass this to there is *one line* of code. So maybe you can find an alternative to keep the control flow linear to read?

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,208 @@
...
+	origin := r.Header.Get("Origin")
+	if !it.AllowedOrigins[origin] {
+		return w.ClientError(safehttp.StatusForbidden)
+	}
+	h := w.Header()
+	allowOrigin, err := h.Claim("Access-Control-Allow-Origin")
+	if err != nil {
+		return w.ServerError(safehttp.StatusInternalServerError)
+	}
+	vary, err := h.Claim("Vary")
```
empijei, line 107: We will need `Vary` for many things, we cannot claim it. I think we might need to just set it (and preserve what was there already).

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,208 @@
...
+
+	if r.Method() == safehttp.MethodOptions {
+		return it.preflight(setHeaders, w, r)
+	} else {
+		return it.request(setHeaders, w, r)
+	}
+}
+
+// preflight handles requests that have the method OPTIONS.
+func (it *Interceptor) preflight(setHeaders func(), w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
```
empijei, line 139: This is quite hard to read. Let's discuss it together tomorrow.

empijei (changes requested):

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// SetAllowedHeaders sets which headers should be allowed in the Access-Control-Allow-Headers
+// header. The headers are first canonicalized using textproto.CanonicalMIMEHeaderKey.
+func (it *Interceptor) SetAllowedHeaders(headers ...string) {
+	it.allowedHeaders = map[string]bool{}
+	for _, h := range headers {
+		it.allowedHeaders[textproto.CanonicalMIMEHeaderKey(h)] = true
+	}
+}
```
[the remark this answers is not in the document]
mattiasgrenfeldt: Why `"*"`?

And why `"X-Cors"`? It will always be an allowed header by default. Letting users specify it in `allowedHeaders` doesn't make a difference, right?

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,208 @@
...
+		return w.ServerError(safehttp.StatusInternalServerError)
+	}
+
+	allowCredentials, err := h.Claim("Access-Control-Allow-Credentials")
+	if err != nil {
+		return w.ServerError(safehttp.StatusInternalServerError)
+	}
+	if r.Header.Get("Cookie") != "" && !it.AllowCredentials {
+		return w.ClientError(safehttp.StatusForbidden)
+	}
```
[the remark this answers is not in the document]
mattiasgrenfeldt: True, but it doesn't hurt checking it on the server side as well, right?

safehttp/plugins/cors/cors.go
mattiasgrenfeldt: Yes, I think so. Could you elaborate on what would be wrong?

safehttp/plugins/cors/cors.go
mattiasgrenfeldt: Yup, I know. To be able to set it we need a `isClaimed` method on the `Header` type though. I talked about this with @kele and he said that we should just claim it for now to not add to much to the API and that other things that needed to set the Vary header should instead talk with the CORS plugin.

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,217 @@
...
+//
+// Constraints
+//
+// - The request methods GET, HEAD and POST are banned and will result in a 405
+//   Method Not Allowed response.
+// - The content types "application/x-www-form-urlencoded", "multipart/form-data"
+//   and "text/plain" are also banned and will result in a  415 Unsupported Media
+//   Type response.
+// - Each CORS request must contain the header "X-Cors: 1" or "Sec-Fetch-Mode: cors".
+// All of this is to prevent XSRF.
```
kele, line 45: Please take a look at https://blog.golang.org/godoc and fix the formatting.

safehttp/plugins/cors/cors.go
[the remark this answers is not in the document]
kele: If you can delete some code, always do. :)

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,217 @@
...
+	if status != 0 && status != safehttp.StatusNoContent {
+		if 400 <= status && status < 500 {
+			return w.ClientError(status)
+		} else {
+			return w.ServerError(status)
+		}
+	}
+
+	if origin != "" {
+		allowOrigin([]string{origin})
```
kele, line 138: We might need to revisit what the `Claim` returns. As it is right now, there is no way to *add* a header value, you always need to replace the whole thing.

mattiasgrenfeldt: You could technically read the currently set header, add to it and then set it again. But I agree. this needs to be revised.

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,214 @@
...
+	} else {
+		status = it.request(w, r)
+	}
+
+	if status != 0 && status != safehttp.StatusNoContent {
+		if 400 <= status && status < 500 {
+			return w.ClientError(status)
+		} else {
+			return w.ServerError(status)
+		}
```
empijei, line 131: We should probably create a helper to do this code, I have a feeling it will be a common occurrence.

Mind opening an issue for this?

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,214 @@
...
+		return safehttp.StatusForbidden
+	}
+	wh := w.Header()
+	allowMethods, err := wh.Claim("Access-Control-Allow-Methods")
+	if err != nil {
+		return safehttp.StatusInternalServerError
+	}
+
+	headers := rh.Get("Access-Control-Request-Headers")
+	if headers != "" {
```
empijei, line 164: This `if` is not needed.

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		return safehttp.StatusInternalServerError
+	}
+
+	maxAge, err := wh.Claim("Access-Control-Max-Age")
+	if err != nil {
+		return safehttp.StatusInternalServerError
+	}
+
+	allowMethods([]string{method})
```
empijei, line 182: Where do you check if the method is allowed? We still want to block HEAD requests.

mattiasgrenfeldt:
Please ignore the failing CI, it is broken (See #128). If I "fix the broken tests" my local tests will fail instead.

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	maxAge([]string{strconv.Itoa(n)})
+	return safehttp.StatusNoContent
+}
+
+// request handles all requests that are not preflight requests.
+func (it *Interceptor) request(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.StatusCode {
+	h := r.Header
+	if h.Get(requiredHeader) != "1" {
+		return safehttp.StatusPreconditionFailed
```
empijei, line 198: We *might* want to tell the client smth, like a message saying that header is required.

safehttp/plugins/cors/cors.go
mattiasgrenfeldt: You opened #127 which would address this, so I won't open another issue. :+1:

safehttp/plugins/cors/cors.go
mattiasgrenfeldt: Can't do that right now since there is no way of checking if the header is claimed without also setting it. :/ We would leak headers to error responses.

safehttp/plugins/cors/cors.go
mattiasgrenfeldt: Can you send HEAD requests with `X-Cors: 1` and a non-standard Content-Type cross site without triggering a preflight? It's for this reason that we allowed GET and POST, so why ban HEAD?

safehttp/plugins/cors/cors.go
mattiasgrenfeldt: True, but at the moment there is no way of sending error responses with messages. I see that your issue #127 would address this, but currently this is not possible.

safehttp/plugins/cors/cors.go
mattiasgrenfeldt: Yes, it is. Running `strings.Split("", ", ")` gives a slice with a single empty string as an element (`[]string{""}`) which will make the loop run and will cause a 403 Forbidden response since `""` isn't an allowed header. So either we have this check or we add `""` as an allowed header.

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,215 @@
...
+	"application/x-www-form-urlencoded": true,
+	"multipart/form-data":               true,
+	"text/plain":                        true,
+}
+
+// Interceptor handles CORS requests based on its settings.
+//
+// For more info about CORS, see: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+//
+// Constraints
```
mihalimara22, line 37: This seems like an unfinished sentence.

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,215 @@
...
+	// is exposed.
+	ExposedHeaders []string
+	// AllowCredentials determines if Access-Control-Allow-Credentials should be
+	// set to true, which would allow cookies to be attached to requests.
+	AllowCredentials bool
+	// MaxAge sets the Access-Control-Max-Age header. MaxAge indicates how long,
+	// in seconds, that the results of a preflight request can be cached.
+	// MaxAge=0 results in MaxAge: 5.
+	// This default of 5 seconds is set because it is the default used by Chromium
+	// according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age
```
mihalimara22, line 65:
```suggestion
	// MaxAge sets the Access-Control-Max-Age header, indicating how many seconds the results of a preflight
	// request can be cached.
	// Max Age=0 results in MaxAge:5,  the default used by Chromium
	// according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age
```
or something along those lines

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,215 @@
...
+	for _, o := range allowedOrigins {
+		ao[o] = true
+	}
+	return &Interceptor{
+		AllowedOrigins: ao,
+		allowedHeaders: map[string]bool{},
+	}
+}
+
+// SetAllowedHeaders sets which headers should be allowed in the Access-Control-Allow-Headers
```
mihalimara22, line 86:
```suggestion
// SetAllowedHeaders sets the headers allowed in the Access-Control-Allow-Headers
```

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,215 @@
...
+	it.allowedHeaders = map[string]bool{}
+	for _, h := range headers {
+		if h == "*" {
+			continue
+		}
+		it.allowedHeaders[textproto.CanonicalMIMEHeaderKey(h)] = true
+	}
+}
+
+// Before handles the incoming request and sets headers or responds with an error
```
mihalimara22, line 98: Can you be more detailed about what happens in Before, it's not very intuitive to the user.

mihalimara22 (changes requested):

safehttp/plugins/cors/cors.go
mattiasgrenfeldt: It's a godoc header
image

kele:
Leaving this for Rob and Mara.

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,223 @@
...
+	return &Interceptor{
+		AllowedOrigins: ao,
+		allowedHeaders: map[string]bool{},
+	}
+}
+
+// SetAllowedHeaders sets the headers allowed in the Access-Control-Allow-Headers
+// header. The headers are first canonicalized using textproto.CanonicalMIMEHeaderKey.
+func (it *Interceptor) SetAllowedHeaders(headers ...string) {
+	it.allowedHeaders = map[string]bool{}
```
mihalimara22, line 89: Is this to overwrite the content of `allowedHeaders` if any? Asking because you are already instantiating the map in `Default`.

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,223 @@
...
+	for _, h := range headers {
+		if h == "*" {
+			continue
```
mihalimara22, line 92: I would add a comment here or in godoc to explain that we avoid allowing all headers.

safehttp/plugins/cors/cors_test.go
```diff
@@ -0,0 +1,574 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			rr := safehttptest.NewResponseRecorder()
+
+			it := cors.Default("https://foo.com")
+			it.AllowCredentials = tt.allowCredentials
+			it.ExposedHeaders = tt.exposedHeaders
+			it.Before(rr.ResponseWriter, tt.req)
+
```
mihalimara22, line 174: Why is the status code not checked?

mattiasgrenfeldt: Well this path through the plugin never sets a status, it just returns a `safehttp.Result{}`, leaving the setting of the status up to later plugins or the handler. So there isn't really anything to check. We should probably check that the status is unset, but currently the `safehttptest.ResponseRecorder` sets 200 OK as the default response status. Maybe we should change this to be 0 instead? So that we could check for these cases.

safehttp/plugins/cors/cors.go
mattiasgrenfeldt: Yup. `SetAllowedHeaders` wants to overwrite everything that is already there. And we need a non-nil map by default, otherwise the plugin will panic. So we will need a new fresh map in each case.

safehttp/plugins/cors/cors_test.go
maramihali: Hmm, I think that'd be a good idea to also avoid future bugs (e.g. we expect something to set the status code to 200 OK, but in fact it doesn't and tests pass because  `safehttptest.ResponseRecorder` AND `safehttp_test.responseRecorder` set the status to 200 OK by default). However, I think that should be in a separate clean-up PR not here.

maramihali (approved):

safehttp/plugins/cors/cors.go
empijei: Are we still blocked on this?

safehttp/plugins/cors/cors.go
empijei: We have observed over the years that head requests could be crafted via browser bugs.

Also no well intended application I have ever seen uses HEAD requests.

safehttp/plugins/cors/cors.go
mattiasgrenfeldt: No, so I will fix it

mattiasgrenfeldt:
@empijei I fixed the two most recent comments you left about appending Vary and disallowing HEAD.

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,211 @@
...
+//
+// All of this is to prevent XSRF.
+type Interceptor struct {
+	// AllowedOrigins determines which origins should be allowed in the
+	// Access-Control-Allow-Origin header.
+	AllowedOrigins map[string]bool
+	allowedHeaders map[string]bool
+	// ExposedHeaders determines which headers should be set in the
+	// Access-Control-Expose-Headers header. This controls which headers in the
+	// response should be accessible by JavaScript.
```
maramihali, line 53:
```suggestion
	// Access-Control-Expose-Headers header. This controls which headers are
	//  accessible by JavaScript in the response.
```

maramihali (approved):
One nit in the comment, overall looks good to me! :)

safehttp/plugins/cors/cors.go
```diff
@@ -0,0 +1,211 @@
...
+
+	var status safehttp.StatusCode
+	m := r.Method()
+	if m == safehttp.MethodOptions {
+		status = it.preflight(w, r)
+	} else if m == safehttp.MethodHead {
+		status = safehttp.StatusMethodNotAllowed
+	} else {
+		status = it.request(w, r)
+	}
```
empijei, line 130: Maybe `switch m`?

safehttp/plugins/cors/cors_test.go
mattiasgrenfeldt: Opened #152

empijei (approved):

jub0bs:
@mattiasgrenfeldt 

> The Content-Types `application/x-www-form-urlencoded`, `multipart/form-data` and `text/plain` are also banned.

Is that to force preflight? In that case, that security control is ineffectual, as it can be bypassed with something like:

```http
Content-Type: application/x-www-form-urlencoded; foo
```

More about this in https://jub0bs.com/posts/2022-02-08-cve-2022-21703-writeup/#bypassing-content-type-validation-and-avoiding-cors-preflight

PR #106, opened 2020-08-24
Changed safehttptest.ResponseRecorder.Status() to return a safehttp.StatusCode instead of an int.

### LINKED ISSUE #105: Modify ResponseRecorder.Status() to return a safehttp.StatusCode instead of int
Currently, calling `ResponseRecorder.Status` function returns the status code as an `int` that we then have to cast to `safehttp.StatusCode` in every test. We should make it return an `safehttp.StatusCode` by default and modify the tests.

mattiasgrenfeldt:
Fixes #105 

This is so that we can use the `safehttp.StatusXxxx` constants instead of number literals in tests without having to cast the constants from `safehttp.StatusCode` to `int` in every test we write that uses `safehttptest`.

### DIFF
--- a/safehttp/safehttptest/recorder.go
+++ b/safehttp/safehttptest/recorder.go
@@ -84,7 +84,7 @@ func (r *ResponseRecorder) Header() http.Header {
 
 // Status returns the recorded response status code.
 func (r *ResponseRecorder) Status() safehttp.StatusCode {
-	return safehttp.StatusCode(r.rw.status)
+	return r.rw.status
 }
 
 // Body returns the recorded response body.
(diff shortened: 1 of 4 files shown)


mihalimara22 (approved):

empijei (approved):

safehttp/safehttptest/recorder.go
```diff
@@ -83,8 +83,8 @@ func (r *ResponseRecorder) Header() http.Header {
 }
 
 // Status returns the recorded response status code.
-func (r *ResponseRecorder) Status() int {
-	return r.rw.status
+func (r *ResponseRecorder) Status() safehttp.StatusCode {
+	return safehttp.StatusCode(r.rw.status)
```
empijei, line 87: Maybe let's also change the ResponseRecorder.status to be of this type?

mattiasgrenfeldt:
@empijei this can be merged :)

PR #114, opened 2020-08-25
Per handler interceptor configurations

### LINKED ISSUE #87: Per-handler configuration for plugins
Plugins should be configurable per handler. This is especially important for plugins which use interceptors.

Examples:
1. `mux.Handler(..., csp.Disable())` -- CSP plugin is installed on the mux, but disabled for a given handler.
2. `mux.Handler(..., cors.Enable())` -- CORS plugin is installed on the mux, but by default it is not applied to any handlers, therefore we need to enable it.
3. `mux.Handler(..., auth.AllowAnonymous())` -- to change how access control is done.

This need stems from the design decision of not having per-handler plugins. The problems of having per-handler plugins (as opposed to mux-level plugins and per-plugin configurations):
- what should be the order of interceptors execution?
- overriding of plugins installed on handler level by the ones on mux level - how to *remove* a plugin?

Overall, having configurations seems like an easier option.

mihalimara22:
Fixes #87 

Added interceptor configurations that enable the user to modify the behaviour of interceptors per handler. Interceptors need to be installed before registering handlers otherwise the interceptor won't take effect. The configuration type is expected to be implemented for every interceptor that is configurable.

### DIFF
--- a/safehttp/config.go
+++ b/safehttp/config.go
@@ -1,0 +1,24 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package safehttp
+
+// Config provide additional configurations to Interceptors when needed.
+type Config interface {
+	// Apply applies the configurations to an Interceptor if it's of the
+	// appropriate type. The bool indicates whether the configuration took
+	// effect. If that's the case, the modified interceptor should be returned
+	// and otherwise Apply should return a nil Interceptor.
+	Apply(Interceptor) (Interceptor, bool)
+}
--- a/safehttp/config.go
+++ b/safehttp/config.go
@@ -16,9 +16,7 @@ package safehttp
 
 // Config provide additional configurations to Interceptors when needed.
 type Config interface {
-	// Apply applies the configurations to an Interceptor, if it's of the
-	// appropriate type. The bool indicates whether the configuration took
-	// effect, in which case, the modified interceptor will be returned.
-	// Otherwise, Apply returns the unmodified Interceptor.
-	Apply(Interceptor) (Interceptor, bool)
+	// Match checks whether this Config is meant to be used with the provided
+	// Interceptor.
+	Match(Interceptor) bool
 }
--- a/safehttp/config.go
+++ b/safehttp/config.go
@@ -16,9 +16,9 @@ package safehttp
 
 // Config provide additional configurations to Interceptors when needed.
 type Config interface {
-	// Apply applies the configurations to an Interceptor if it's of the
+	// Apply applies the configurations to an Interceptor, if it's of the
 	// appropriate type. The bool indicates whether the configuration took
-	// effect. If that's the case, the modified interceptor should be returned
-	// and otherwise Apply should return a nil Interceptor.
+	// effect, in which case, the modified interceptor will be returned.
+	// Otherwise, Apply returns the unmodified Interceptor.
 	Apply(Interceptor) (Interceptor, bool)
 }
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -106,11 +106,11 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
 // interceptors on a registered handler. Passing a Config whose corresponding
 // Interceptor was not installed will produce no effect.
 func (m *ServeMux) Handle(pattern string, method string, h Handler, configs ...Config) {
-	var changed bool
 	interceps := map[string]Interceptor{}
 	for k, i := range m.interceps {
 		interceps[k] = i
 	}
+	var changed bool
 	for _, c := range configs {
 		for k, i := range interceps {
 			if i, ok := c.Apply(i); ok {
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -80,7 +80,8 @@ type ServeMux struct {
 	handlers map[string]methodHandler
 
 	// Maps interceptor key to interceptor.
-	interceps map[string]Interceptor
+	interceps      map[string]Interceptor
+	intercepsOrder []string
 }
 
 // NewServeMux allocates and returns a new ServeMux
@@ -105,27 +106,21 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
 // Configs can be optionally passed in order to modify the behavior of the
 // interceptors on a registered handler. Passing a Config whose corresponding
 // Interceptor was not installed will produce no effect.
-func (m *ServeMux) Handle(pattern string, method string, h Handler, configs ...Config) {
-	interceps := map[string]Interceptor{}
-	for k, i := range m.interceps {
-		interceps[k] = i
-	}
-	var changed bool
-	for _, c := range configs {
-		for k, i := range interceps {
-			if i, ok := c.Apply(i); ok {
-				interceps[k] = i
-				changed = true
+func (m *ServeMux) Handle(pattern string, method string, h Handler, cfgs ...Config) {
+	cfgMap := map[string]Config{}
+	for _, cfg := range cfgs {
+		for k, it := range m.interceps {
+			if cfg.Match(it) {
+				cfgMap[k] = cfg
 			}
 		}
 	}
-	if !changed {
-		interceps = m.interceps
-	}
 	hi := handlerWithInterceptors{
-		handler:   h,
-		interceps: interceps,
-		disp:      m.disp,
+		handler:        h,
+		interceps:      m.interceps,
+		disp:           m.disp,
+		configs:        cfgMap,
+		intercepsOrder: m.intercepsOrder,
 	}
 
 	mh, ok := m.handlers[pattern]
@@ -157,6 +152,7 @@ func (m *ServeMux) Install(key string, i Interceptor) {
 		panic("interceptor with same key already installed")
 	}
 	m.interceps[key] = i
+	m.intercepsOrder = append(m.intercepsOrder, key)
 }
 
 // ServeHTTP dispatches the request to the handler whose method matches the
@@ -192,9 +188,11 @@ func (m methodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // handlerWithInterceptors encapsulates a handler and its corresponding
 // interceptors.
 type handlerWithInterceptors struct {
-	handler   Handler
-	interceps map[string]Interceptor
-	disp      Dispatcher
+	handler        Handler
+	configs        map[string]Config
+	interceps      map[string]Interceptor
+	intercepsOrder []string
+	disp           Dispatcher
 }
 
 // ServeHTTP calls the Before method of all the interceptors and then calls the
@@ -212,8 +210,8 @@ func (h handlerWithInterceptors) ServeHTTP(w http.ResponseWriter, r *http.Reques
 		}
 	}()
 
-	for _, it := range h.interceps {
-		it.Before(rw, ir)
+	for _, key := range h.intercepsOrder {
+		h.interceps[key].Before(rw, ir, h.configs[key])
 		if rw.written {
 			return
 		}
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -102,9 +102,11 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
 // Handle registers a handler for the given pattern and method. If another
 // handler is already registered for the same pattern and method, Handle panics.
 //
-// Configs can be optionally passed which will modify the behavior of the
-// interceptors on the handler registered.
+// Configs can be optionally passed in order to modify the behavior of the
+// interceptors on a registered handler. Passing a Config whose corresponding
+// Interceptor was not installed will produce no effect.
 func (m *ServeMux) Handle(pattern string, method string, h Handler, configs ...Config) {
+	var changed bool
 	interceps := map[string]Interceptor{}
 	for k, i := range m.interceps {
 		interceps[k] = i
@@ -113,10 +115,13 @@ func (m *ServeMux) Handle(pattern string, method string, h Handler, configs ...C
 		for k, i := range interceps {
 			if i, ok := c.Apply(i); ok {
 				interceps[k] = i
+				changed = true
 			}
 		}
 	}
-
+	if !changed {
+		interceps = m.interceps
+	}
 	hi := handlerWithInterceptors{
 		handler:   h,
 		interceps: interceps,
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -101,14 +101,33 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
 
 // Handle registers a handler for the given pattern and method. If another
 // handler is already registered for the same pattern and method, Handle panics.
-func (m *ServeMux) Handle(pattern string, method string, h Handler) {
+//
+// Configs can be optionally passed which will modify the behavior of the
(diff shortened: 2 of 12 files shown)


safehttp/mux.go
```diff
@@ -97,14 +97,33 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
...
+	for k, i := range m.interceps {
+		interceps[k] = i
+	}
+	for _, c := range configs {
+		for k, i := range interceps {
+			if i, ok := c.Apply(i); ok {
+				interceps[k] = i
+			}
+		}
+	}
```
empijei, line 114: Do we need to copy the map over if no configs have been passed?
E.g. if we never get `true` we could just discard the new map.

safehttp/mux.go
```diff
@@ -97,14 +97,33 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
...
+
```
empijei, line 115: Is it deliberate to not handle the case in which a config is passed, but for a plugin that is not installed? If so, please add a comment about it. If not, handle it.

tests/integration/hsts/hsts_test.go
```diff
@@ -68,13 +68,13 @@ func (r *responseRecorder) Write(data []byte) (int, error) {
 func TestHSTSServeMuxInstall(t *testing.T) {
 	mux := safehttp.NewServeMux(&dispatcher{}, "foo.com")
 
+	mux.Install("hsts", hsts.Default())
+
```
empijei, line 72: We have to do something to avoid users making the same mistake. I opened #117

safehttp/config.go
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+
+package safehttp
+
+// Config provide additional configurations to Interceptors when needed.
+type Config interface {
+	// Apply applies the configurations to an Interceptor if it's of the
+	// appropriate type. The bool indicates whether the configuration took
+	// effect. If that's the case, the modified interceptor should be returned
+	// and otherwise Apply should return a nil Interceptor.
```
empijei, line 22: Why return nil and not the previous one?

empijei (changes requested):
Looks good overall, but please take a look at the comments.

safehttp/mux.go
mihalimara22: Well, each config implementation wshould contain a type assertion that should check whether the interceptor received as parameter is of the right type (e.g. `Foo` for `FooConfig`). Passing a config that doesn't apply to any interceptor, because it wasn't installed for example, will just produce no effect. I will add this in the function's godoc.

safehttp/config.go
mihalimara22: We thought that the result can just be ignored if no modifications were applied, but we can also return the unmodified Interceptor. WDYT?

tests/integration/hsts/hsts_test.go
mihalimara22: When we call `mux.Install` to install a plugin, we could check whether any handler was registered, in which case we panic because plugins should be installed before handlers in order to take effect. WDYT?

safehttp/mux.go
```diff
@@ -97,14 +97,38 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
 
 // Handle registers a handler for the given pattern and method. If another
 // handler is already registered for the same pattern and method, Handle panics.
-func (m *ServeMux) Handle(pattern string, method string, h Handler) {
+//
+// Configs can be optionally passed in order to modify the behavior of the
+// interceptors on a registered handler. Passing a Config whose corresponding
+// Interceptor was not installed will produce no effect.
+func (m *ServeMux) Handle(pattern string, method string, h Handler, configs ...Config) {
+	var changed bool
```
mattiasgrenfeldt, line 105: Move this to right before the nested for-loop since that is where it is first used.

tests/integration/hsts/hsts_test.go
empijei: sounds like a *meh* solution. We should discuss this with @kele 

Panicking on startup is fine but wouldn't it be better if we solved this on a type level?

kele:
@chressie @bsiegert @stapelberg The idea behind these configurations is similar to Goa's per-action options. I'd really appreciate your feedback here.

Some of the important things to consider:
- when should "applying" the options happen in the lifecycle of the server?
- are the types right (i.e. maybe we need more/less context?)

safehttp/config.go
```diff
@@ -0,0 +1,24 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package safehttp
+
+// Config provide additional configurations to Interceptors when needed.
+type Config interface {
```
bsiegert, line 18: Maybe `Option` would be a better name?

bsiegert (commented):
The way this works is a bit different from how Goa does it. In Goa, a HandlerOption passes values through the context to the interceptors:

```
type Option interface {
	Set(context.Context) context.Context
}
```

This has the advantage that you avoid copying the interceptor map on each request (although we end up doing this copy elsewhere in the stack, unfortunately).

safehttp/config.go
```diff
@@ -0,0 +1,24 @@
...
+	// Apply applies the configurations to an Interceptor if it's of the
+	// appropriate type. The bool indicates whether the configuration took
+	// effect. If that's the case, the modified interceptor should be returned
```
chressie: Should it be made more explicit that the returned interface must represent a _copy_? In particular if the type that implements Interceptor is a pointer type, the corresponding Config must ensure that it returns a deep copy and not the same interface again. I think the Handle method could even enforce this by comparing `i==<returned interceptor>`.

safehttp/config.go
```diff
@@ -16,9 +16,9 @@ package safehttp
 
 // Config provide additional configurations to Interceptors when needed.
 type Config interface {
-	// Apply applies the configurations to an Interceptor if it's of the
+	// Apply applies the configurations to an Interceptor, if it's of the
 	// appropriate type. The bool indicates whether the configuration took
-	// effect. If that's the case, the modified interceptor should be returned
-	// and otherwise Apply should return a nil Interceptor.
+	// effect, in which case, the modified interceptor will be returned.
+	// Otherwise, Apply returns the unmodified Interceptor.
```
chressie: Maybe just do not specifying what Interceptor is when ok==false. Another question that arises is why we need to have OK at all. Why isn't `nil` a fine contract to say "I couldn't do anything with this type". What's the benefit of the bool? The implementation ignores the returned Interceptor anyway, if ok==false.

Also is it possible that Apply returns an error? What would an implementation do if Config results in an invalid Interceptor ... or should this be handled when creating the Config, so it shouldn't be Handler registrations business?

safehttp/config.go
mihalimara22: We considered adding the options to the context. The drawback of this is that retrieving a value from the context is a linear search.

safehttp/mux.go
```diff
@@ -164,11 +182,28 @@ func (m methodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	disp      Dispatcher
+}
+
+// ServeHTTP calls the Before method of all the interceptors and then calls the
+// underlying handler.
+func (h handlerWithInterceptors) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	rw := NewResponseWriter(h.disp, w, h.interceps)
+	ir := NewIncomingRequest(r)
+
+	for _, it := range h.interceps {
```
kele, line 202: This line makes it nondeterministic (I actually have missed this issue before, as it was a `map` earlier too).

safehttp/mux.go
```diff
@@ -164,11 +182,28 @@ func (m methodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	for _, intercep := range m.muxInterceps {
-		if res := intercep.Before(rw, ir); res.written {
+	h.ServeHTTP(w, r)
+}
+
+// handlerWithInterceptors encapsulates a handler and its corresponding
+// interceptors.
+type handlerWithInterceptors struct {
+	handler   Handler
+	interceps map[string]Interceptor
```
kele, line 192: Have you considered doing this instead:

1. For each interceptor, storing both a `Interceptor` and `cfg interface{}`.
2. Pass `cfg` to `Before` and all other methods, i.e. `Before(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest, cfg interface{})`?

This way we:
1. Avoid the awkward "deep-copy an interceptor" required by `Apply` right now.
2. Don't need the linear search in `context.Context` that Goa does.

kele (changes requested):

safehttp/mux.go
```diff
@@ -97,14 +97,38 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
...
+	var changed bool
+	for _, c := range configs {
+		for k, i := range interceps {
+			if i, ok := c.Apply(i); ok {
+				interceps[k] = i
+				changed = true
+			}
+		}
+	}
+	if !changed {
```
kele, line 118: This seems like a premature micro-optimization at first glance. What's the purpose of this?

kele (changes requested):

safehttp/config.go
kele: I'd rather drop the `Interceptor` return value altogether and have separate storage for plugin-specific configs. See https://github.com/google/go-safeweb/pull/114#discussion_r479628063

tests/integration/hsts/hsts_test.go
kele: #117

safehttp/config.go
kele: This response was likely meant for another comment.

With regards to `Option` vs `Config`... I'd rather stick with `Config` for one specific reason here: it would seem that one `Interceptor` should have one `Config`, but could have multiple `Option`s. I'm leaning towards adjusting the interceptor's behavior just once.

safehttp/mux.go
```diff
@@ -97,14 +97,38 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
...
+// interceptors on a registered handler. Passing a Config whose corresponding
+// Interceptor was not installed will produce no effect.
+func (m *ServeMux) Handle(pattern string, method string, h Handler, configs ...Config) {
+	interceps := map[string]Interceptor{}
+	for k, i := range m.interceps {
+		interceps[k] = i
+	}
+	var changed bool
+	for _, c := range configs {
+		for k, i := range interceps {
```
stapelberg, line 111: I assume that there will typically be more interceptors than configs, so it might be better to swap these two loops around: iterate over interceptors only once, and then apply all the configs.

safehttp/config.go
```diff
@@ -0,0 +1,24 @@
...
+
+package safehttp
+
+// Config provide additional configurations to Interceptors when needed.
+type Config interface {
+	// Apply applies the configurations to an Interceptor, if it's of the
+	// appropriate type. The bool indicates whether the configuration took
+	// effect, in which case, the modified interceptor will be returned.
+	// Otherwise, Apply returns the unmodified Interceptor.
+	Apply(Interceptor) (Interceptor, bool)
```
stapelberg, line 23: Does Apply need to be exported? The answer is yes if external types need to be able to implement it, but I’m not sure that’s the case?

safehttp/mux.go
mattiasgrenfeldt: Why is that better? Is it because iterating over a map is more expensive than iterating over a slice and therefore it's better to iterate over the slice many times and over the map just once?

mattiasgrenfeldt:
We changed the design of this in the latest push. Now it is more similar to what @kele suggested [here](https://github.com/google/go-safeweb/pull/114#discussion_r479628063)

`Before` now takes an optional `config Config` argument instead of applying the config to the interceptor at registration time. We have also added a slice of interceptor keys to iterate over when running the interceptors to make the order deterministic, instead of iterating over a map like we did before.

safehttp/config.go
mattiasgrenfeldt: We have now removed `Apply`.

safehttp/mux.go
mattiasgrenfeldt: We now have a slice of interceptor keys to iterate over instead.

safehttp/mux.go
mattiasgrenfeldt: It's been removed in the latest implementation

safehttp/config.go
mattiasgrenfeldt: Since each config and interceptor have a different meaning of "applying the config" I think exporting the `Apply` method was needed, but we have now removed `Apply` in the latest design.

safehttp/mux.go
```diff
@@ -80,7 +80,8 @@ type ServeMux struct {
 	handlers map[string]methodHandler
 
 	// Maps interceptor key to interceptor.
-	interceps map[string]Interceptor
+	interceps      map[string]Interceptor
+	intercepsOrder []string
```
kele, line 84: Note: this is an artifact of having keyed interceptors. We've decided today that this should not be the case, so in a subsequent PR this should be changed to `interceptors []Interceptor`, or even better `interceptors []appliedInterceptor` where:

```
type appliedInterceptor struct {
  interceptor Interceptor
  cfg interface{}
}
```

safehttp/mux.go
```diff
@@ -177,12 +210,12 @@ func (m methodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}()
 
-	for _, intercep := range m.muxInterceps {
-		intercep.Before(rw, ir)
+	for _, key := range h.intercepsOrder {
```
kele, line 213: Leaving this as it is for now, but this should be just one slice in the end. The `h.interceps[key]` lookup has some tiny cost and is rather unnecessary. Please note it down for a future PR.

safehttp/mux_test.go
```diff
@@ -336,3 +316,156 @@ func TestMuxInterceptors(t *testing.T) {
...
+type setHeaderConfig struct {
+	name  string
+	value string
+}
+
+func (setHeaderConfig) Match(i safehttp.Interceptor) bool {
+	if _, ok := i.(setHeaderConfigInterceptor); !ok {
+		return false
+	}
+	return true
```
kele, line 329:
```suggestion
	_, ok := i.(setHeaderConfigInterceptor)
	return ok
```

kele (approved):
Approving. Further changes expected in subsequent PRs.

safehttp/mux_test.go
```diff
@@ -336,3 +316,154 @@ func TestMuxInterceptors(t *testing.T) {
...
+
+			if got := b.String(); got != tt.wantBody {
+				t.Errorf("response body: got %q want %q", got, tt.wantBody)
+			}
+		})
+	}
+}
+
+type interceptorOne struct {
+}
```
empijei, line 408:
```suggestion
type interceptorOne struct {}
```

empijei (approved):

safehttp/mux.go
```diff
@@ -168,6 +182,25 @@ func (m methodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
+	h.ServeHTTP(w, r)
+}
+
+// handlerWithInterceptors encapsulates a handler and its corresponding
+// interceptors.
+type handlerWithInterceptors struct {
+	handler        Handler
+	configs        map[string]Config
+	interceps      map[string]Interceptor
```
stapelberg, line 193: (Why?) Is it required to keep a map here?

Couldn’t you turn the interceptors into an ordered slice (with their config) and avoid the map altogether in the `handlerWithInterceptors` type?

For serving, I don’t see the need, and a map lookup is more expensive than just iterating through a slice.

stapelberg: Ah, I see that kele raced me to it with the appliedInterceptor remark.

safehttp/mux.go
stapelberg: Yes, that’s one aspect. Another one I had in mind is memory locality: it’s better to do a few changes at once to each interceptor, rather than go through all the interceptors over and over again.

Probably doesn’t matter here at all though, so feel free to ignore me.

(I had in mind that when working with images, it’s better to iterate over Y in the outer loop, and X in the inner loop, for memory locality.)

PR #120, opened 2020-08-27
Added a safehttp.NotWritten result so that safehttp.Result{} can stop being used.

### LINKED ISSUE #109: Add a safehttp.NotWritten Result
This needs to be well documented and requires to change all plugins to use it.

We should also think of a mechanism to prevent future code to use `safehttp.Result{}` instead.

mattiasgrenfeldt:
Fixes #109 

I added a `safehttp.NotWritten` variable which we can use everywhere instead of `safehttp.Result{}`. If a `Handler` returns `NotWritten` then a `204 No Content` response is automatically set by the ServeMux.

### DIFF
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -218,4 +218,7 @@ func (h handlerWithInterceptors) ServeHTTP(w http.ResponseWriter, r *http.Reques
 	}
 
 	h.handler.ServeHTTP(rw, ir)
+	if !rw.written {
+		rw.NoContent()
+	}
 }
--- a/safehttp/response_writer.go
+++ b/safehttp/response_writer.go
@@ -54,13 +54,15 @@ func (w *ResponseWriter) Interceptor(key string) Interceptor {
 // Result TODO
 type Result struct{}
 
-// NotWritten is a Result which indicates that nothing has been written yet. It
+// NotWritten returns a Result which indicates that nothing has been written yet. It
 // can be used in all functions that return a Result, such as in the ServeHTTP method
 // of a Handler or in the Before method of an Interceptor. When returned, NotWritten
 // indicates that the writing of the response should take place later. When this
 // is returned by the Before method in Interceptors the next Interceptor in line
 // is run. When this is returned by a Handler, a 204 No Content response is written.
-var NotWritten Result
+func NotWritten() Result {
+	return Result{}
+}
 
 // Write TODO
 func (w *ResponseWriter) Write(resp Response) Result {
--- a/safehttp/response_writer.go
+++ b/safehttp/response_writer.go
@@ -54,6 +54,14 @@ func (w *ResponseWriter) Interceptor(key string) Interceptor {
 // Result TODO
 type Result struct{}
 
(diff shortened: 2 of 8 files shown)


safehttp/mux.go
```diff
@@ -170,5 +170,7 @@ func (m methodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	h.ServeHTTP(rw, ir)
+	if res := h.ServeHTTP(rw, ir); !res.written {
+		rw.NoContent()
```
empijei, line 174: The go default behavior is to return a 200 OK with no body, maybe let's use that?

safehttp/response_writer.go
```diff
@@ -55,6 +55,14 @@ type Result struct {
 	written bool
 }
 
+// NotWritten is a Result which indicates that nothing has been written yet. It
+// can be used in all functions that return a Result, such as in the ServeHTTP method
+// of a Handler or in the Before method of an Interceptor. When returned, NotWritten
+// indicates that the writing of the response should take place later. When this
+// is returned by the Before method in Interceptors the next Interceptor in line
+// is run. When this is returned by a Handler, a 204 No Content response is written.
```
empijei, line 63: if you change the default behavior above change this doc too.

safehttp/mux.go
mattiasgrenfeldt: Right, makes sense. But how?
```go
rw.Write(nil) // nil must be a valid safe response in the dispatcher
```
```go
rw.rw.WriteHeader(200) // sidestep dispatcher just like in NoContent
```
Or add a new `WriteNoContent` method?

kele:
Why is this better than `safehttp.Result{}`?

mattiasgrenfeldt:
@empijei asked us one day if it would be nicer to code the plugins with a `NotWritten` constant being returned instead of `Result{}` all the time. Me and @mihalimara22  thought this would be clearer for the users of the framework. This way you could change the Result type in the future easily without having all the plugins change their returns, you just change the `NotWritten` variable.

kele:
Please change it to a constructor then, i.e. `safehttp.NotWritten()` call. I can imagine parts of the framework modify the `safehttp.Result` type along the chain, and they'd all modify the same package-level variable.

kele (changes requested):
Added a PR comment, but also need to write one here because GitHub won't let me Request Changes otherwise. :)

In this case, here's a corny joke:

```
Did you know the first French fries weren't actually cooked in France? They were cooked in Greece.
```

kele (approved):
Needs a sync to master.

safehttp/mux.go
empijei: I would WriteHeader(200) or do nothing. The http package will take care of setting a 200.

mattiasgrenfeldt:
Done! @kele

safehttp/mux.go
mattiasgrenfeldt: Sorry, I missed your response. But I opened #153 so this won't be forgotten.

PR #123, opened 2020-08-27
Added token generation to the XSRF plugin.

### LINKED ISSUE #122: XSRF Token Generation
In its current format, the XSRF plugin only supports verification of the XSRF token in  the `safehttp.IncomingRequest` body. We also need to generate tokens and add them to a `safehttp.IncomingRequest` context. These will then be injected  by the `htmlinject` plugin as hidden inputs to forms.

mihalimara22:
Fixes #122 

Extended the XSRF plugin to support token generation in the `xsrf.Before()` function. This will then be added to the context of a `safehttp.IncomingRequest` and can be retrieved using the `xsrf.Token()` getter.

### DIFF
--- a/safehttp/incoming_request.go
+++ b/safehttp/incoming_request.go
@@ -47,6 +47,7 @@ func NewIncomingRequest(req *http.Request) *IncomingRequest {
 // Method specifies the HTTP method of an IncomingRequest.
 func (r *IncomingRequest) Method() string {
 	return r.req.Method
+<<<<<<< HEAD
 }
 
 // Host returns the host the request is targeted to.
@@ -61,6 +62,8 @@ func (r *IncomingRequest) Host() string {
 // implemented.
 func (r *IncomingRequest) Path() string {
 	return r.req.URL.Path
+=======
+>>>>>>> Added token generation functionality to the XSRF plugin. The generated token will be added to each of the safehttp.IncomingRequest contexts and subsequently injected as a hidden input in form by the htmlinject plugin
 }
 
 // PostForm parses the form parameters provided in the body of a POST, PATCH or
--- a/safehttp/incoming_request.go
+++ b/safehttp/incoming_request.go
@@ -47,23 +47,6 @@ func NewIncomingRequest(req *http.Request) *IncomingRequest {
 // Method specifies the HTTP method of an IncomingRequest.
 func (r *IncomingRequest) Method() string {
 	return r.req.Method
-<<<<<<< HEAD
-}
-
-// Host returns the host the request is targeted to.
-// TODO(@mihalimara22): Remove this after the safehttp.URL type has been implemented
-func (r *IncomingRequest) Host() string {
-	return r.req.Host
-}
-
-// Path returns the relative path of the URL in decoded format (e.g. %47%6f%2f
-// becomes /Go/).
-// TODO(@mihalimara22): Remove this after the safehttp.URL type has been
-// implemented.
-func (r *IncomingRequest) Path() string {
-	return r.req.URL.Path
-=======
->>>>>>> Added token generation functionality to the XSRF plugin. The generated token will be added to each of the safehttp.IncomingRequest contexts and subsequently injected as a hidden input in form by the htmlinject plugin
 }
 
 // PostForm parses the form parameters provided in the body of a POST, PATCH or
--- a/safehttp/plugins/xsrf/xsrf.go
+++ b/safehttp/plugins/xsrf/xsrf.go
@@ -44,10 +44,10 @@ type UserIdentifier interface {
 // Interceptor implements XSRF protection.
 type Interceptor struct {
 	// AppKey uniquely identifies each registered service and should have high
-	// entropy as it is use in generating the XSRF token.
+	// entropy as it is used for generating the XSRF token.
 	AppKey string
-	// Identifier supports retrieving the user ID of application's user based on
-	// incoming requests. This is needed in generating the XSRF token.
+	// Identifier supports retrieving the user ID based on the incoming
+	// request. This is needed for generating the XSRF token.
 	Identifier UserIdentifier
 }
 
@@ -84,6 +84,9 @@ func (i *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingReq
 	if needsValidation {
 		f, err := r.PostForm()
 		if err != nil {
+			// We fallback to checking whether the form is multipart. Both types
+			// are valid in an incoming request as long as the XSRF token is
+			// present.
 			mf, err := r.MultipartForm(32 << 20)
 			if err != nil {
 				return w.ClientError(safehttp.StatusBadRequest)
--- a/safehttp/plugins/xsrf/xsrf.go
+++ b/safehttp/plugins/xsrf/xsrf.go
@@ -79,8 +79,9 @@ func (i *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingReq
 		return w.ClientError(safehttp.StatusUnauthorized)
 	}
 
-	if m := r.Method(); statePreservingMethods[m] {
-		tok := xsrftoken.Generate(i.AppKey, userID, r.URL.String())
+	actionID := r.Method() + " " + r.URL.Path()
+	if statePreservingMethods[r.Method()] {
+		tok := xsrftoken.Generate(i.AppKey, userID, actionID)
 		r.SetContext(context.WithValue(r.Context(), tokenCtxKey{}, tok))
 		return safehttp.Result{}
 	}
@@ -99,11 +100,11 @@ func (i *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingReq
 		return w.ClientError(safehttp.StatusUnauthorized)
 	}
 
-	if ok := xsrftoken.Valid(tok, i.AppKey, userID, r.URL.String()); !ok {
+	if ok := xsrftoken.Valid(tok, i.AppKey, userID, actionID); !ok {
 		return w.ClientError(safehttp.StatusForbidden)
 	}
 
-	tok = xsrftoken.Generate(i.AppKey, userID, r.URL.String())
+	tok = xsrftoken.Generate(i.AppKey, userID, actionID)
 	r.SetContext(context.WithValue(r.Context(), tokenCtxKey{}, tok))
 
 	return safehttp.Result{}
--- a/safehttp/plugins/xsrf/xsrf.go
+++ b/safehttp/plugins/xsrf/xsrf.go
@@ -41,13 +41,13 @@ type UserIdentifier interface {
 	UserID(*safehttp.IncomingRequest) (string, error)
 }
 
-// Interceptor implements XSRF protection. It requires an application key and a
-// storage service. The AppKey uniquely identifies each registered service and
-// should have high entropy. The storage service supports retrieving ID's of the
-// application's users. Both the AppKey and user ID are used in the XSRF
-// token generation and verification algorithm.
+// Interceptor implements XSRF protection.
 type Interceptor struct {
-	AppKey     string
+	// AppKey uniquely identifies each registered service and should have high
+	// entropy as it is use in generating the XSRF token.
+	AppKey string
+	// Identifier supports retrieving the user ID of application's user based on
+	// incoming requests. This is needed in generating the XSRF token.
 	Identifier UserIdentifier
 }
 
@@ -65,12 +65,15 @@ func Token(r *safehttp.IncomingRequest) (string, error) {
 
 // Before should be executed before directing the safehttp.IncomingRequest to
 // the handler to ensure it is not part of the Cross-Site Request
-// Forgery. In case of state changing requests (all except GET, HEAD and OPTIONS
-// requests), it checks for the presence of an xsrf-token in the request body
-// and validates it based on the userID associated with the request. For all
-// requests, it adds a cryptographically safe XSRF token to the
-// safehttp.IncomingRequest context.
-func (i *Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest, cfg interface{}) safehttp.Result {
+// Forgery attack.
+//
+// In case of state changing requests (all except GET, HEAD and OPTIONS), it
+// checks for the presence of an XSRF token in the request and validates it
+// based on the user ID associated with the request.
+//
+// For authorized requests, it adds a cryptographically safe XSRF token to the
+// incoming request. It can be later extracted using Token.
+func (i *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest, cfg interface{}) safehttp.Result {
 	userID, err := i.Identifier.UserID(r)
 	if err != nil {
 		return w.ClientError(safehttp.StatusUnauthorized)
--- a/safehttp/plugins/xsrf/xsrf.go
+++ b/safehttp/plugins/xsrf/xsrf.go
@@ -43,9 +43,9 @@ type UserIdentifier interface {
 
 // Interceptor implements XSRF protection.
 type Interceptor struct {
-	// AppKey uniquely identifies each registered service and should have high
+	// SecretAppKey uniquely identifies each registered service and should have high
 	// entropy as it is used for generating the XSRF token.
-	AppKey string
+	SecretAppKey string
 	// Identifier supports retrieving the user ID based on the incoming
 	// request. This is needed for generating the XSRF token.
 	Identifier UserIdentifier
@@ -99,12 +99,12 @@ func (i *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingReq
 			return w.ClientError(safehttp.StatusUnauthorized)
 		}
 
-		if ok := xsrftoken.Valid(tok, i.AppKey, userID, actionID); !ok {
+		if ok := xsrftoken.Valid(tok, i.SecretAppKey, userID, actionID); !ok {
 			return w.ClientError(safehttp.StatusForbidden)
 		}
 	}
 
-	tok := xsrftoken.Generate(i.AppKey, userID, actionID)
+	tok := xsrftoken.Generate(i.SecretAppKey, userID, actionID)
 	r.SetContext(context.WithValue(r.Context(), tokenCtxKey{}, tok))
 
 	return safehttp.Result{}
--- a/safehttp/plugins/xsrf/xsrf.go
+++ b/safehttp/plugins/xsrf/xsrf.go
@@ -26,6 +26,11 @@ const (
 	TokenKey = "xsrf-token"
 )
 
+var stateChangingMethods = map[string]bool{
+	safehttp.MethodPost:  true,
+	safehttp.MethodPatch: true,
+}
+
 // UserIdentifier provides the web application users' identifiers,
 // needed in generating the XSRF token.
 type UserIdentifier interface {
@@ -68,13 +73,6 @@ func (i *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingReq
 		return w.ClientError(safehttp.StatusUnauthorized)
 	}
 
-	tok := xsrftoken.Generate(i.AppKey, userID, r.URL.String())
-	r.SetContext(context.WithValue(r.Context(), tokenCtxKey{}, tok))
-
-	stateChangingMethods := map[string]bool{
-		safehttp.MethodPost:  true,
-		safehttp.MethodPatch: true,
-	}
 	if m := r.Method(); stateChangingMethods[m] {
 		f, err := r.PostForm()
 		if err != nil {
@@ -94,6 +92,9 @@ func (i *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingReq
 			return w.ClientError(safehttp.StatusForbidden)
 		}
 	}
+
+	tok := xsrftoken.Generate(i.AppKey, userID, r.URL.String())
+	r.SetContext(context.WithValue(r.Context(), tokenCtxKey{}, tok))
 
 	return safehttp.Result{}
 }
--- a/safehttp/plugins/xsrf/xsrf.go
+++ b/safehttp/plugins/xsrf/xsrf.go
@@ -53,33 +53,32 @@ type Interceptor struct {
 
 type tokenCtxKey struct{}
 
-// Token returns the XSRF token from the safehttp.IncomingRequest Context and a
-// nil error, if present, and, otherwise, returns an empty string and a non-nil
-// error.
+// Token extracts the XSRF token from the incoming request. If it is not
+// present, it returns a non-nil error.
 func Token(r *safehttp.IncomingRequest) (string, error) {
 	tok := r.Context().Value(tokenCtxKey{})
 	if tok == nil {
-		return "", errors.New("xsrf token not found in safehttp.IncomingRequest context")
+		return "", errors.New("xsrf token not found")
 	}
 	return tok.(string), nil
 }
 
 // Before should be executed before directing the safehttp.IncomingRequest to
 // the handler to ensure it is not part of the Cross-Site Request
-// Forgery. In case of state changing methods, it checks for the
-// presence of an xsrf-token in the request body and validates it based on the
-// userID associated with the request. It also adds a cryptographically safe
-// XSRF token to the safehttp.IncomingRequest context.
-func (i *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest, cfg interface{}) safehttp.Result {
+// Forgery. In case of state changing requests (all except GET, HEAD and OPTIONS
+// requests), it checks for the presence of an xsrf-token in the request body
+// and validates it based on the userID associated with the request. For all
+// requests, it adds a cryptographically safe XSRF token to the
+// safehttp.IncomingRequest context.
+func (i *Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest, cfg interface{}) safehttp.Result {
 	userID, err := i.Identifier.UserID(r)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


safehttp/plugins/xsrf/xsrf.go
```diff
@@ -34,40 +35,64 @@ type UserIdentifier interface {
...
+	userID, err := i.Identifier.UserID(r)
 	if err != nil {
 		return w.ClientError(safehttp.StatusUnauthorized)
 	}
-	f, err := r.PostForm()
-	if err != nil {
-		mf, err := r.MultipartForm(32 << 20)
+
+	tok := xsrftoken.Generate(i.AppKey, userID, r.URL.String())
+	r.SetContext(context.WithValue(r.Context(), tokenCtxKey{}, tok))
```
mattiasgrenfeldt, line 72: This is entirely optional but maybe we should move the generation of the new token after the validation of the old token. This way you can't accidentally in some code change in the future validate the new token instead of the old one. It might sound crazy, but I mean, bugs happen. :p

safehttp/plugins/xsrf/xsrf_test.go
```diff
@@ -30,30 +31,25 @@ func (userIdentifier) UserID(r *safehttp.IncomingRequest) (string, error) {
...
-		path       string
-		userID     string
-		wantStatus int
-		wantHeader map[string][]string
-		wantBody   string
+var (
+	stateChangingMethods = []string{safehttp.MethodPost, safehttp.MethodPatch}
+	formTokenTests       = []struct {
+		name, userID, actionID, wantBody string
+		wantStatus                       int
```
mattiasgrenfeldt, line 38: Use `safehttp.StatusCode`

safehttp/plugins/xsrf/xsrf_test.go
```diff
@@ -163,87 +91,204 @@ func TestXSRFTokenMultipart(t *testing.T) {
...
-		req := safehttptest.NewRequest("POST", "http://foo.com/pizza", strings.NewReader(b))
-		req.Header.Set("Content-Type", `multipart/form-data; boundary="123"`)
+)
 
-		rec := safehttptest.NewResponseRecorder()
-		i.Before(rec.ResponseWriter, req)
+func TestTokenPost(t *testing.T) {
+	for _, method := range stateChangingMethods {
+		for _, test := range formTokenTests {
+			t.Run(method+test.name, func(t *testing.T) {
```
mattiasgrenfeldt, line 99: Maybe should have a space between `method` and `test.name`
```go
method + " " + test.name
```

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -34,40 +35,64 @@ type UserIdentifier interface {
...
+// storage service. The AppKey uniquely identifies each registered service and
 // should have high entropy. The storage service supports retrieving ID's of the
-// application's users. Both the appKey and user ID are used in the XSRF
-// token generation algorithm.
+// application's users. Both the AppKey and user ID are used in the XSRF
+// token generation and verification algorithm.
 type Interceptor struct {
 	AppKey     string
 	Identifier UserIdentifier
 }
```
mattiasgrenfeldt, line 45: Maybe we should provide a default constructor for `Interceptor` which takes an `Identifier` and generates a safe `AppKey`

safehttp/plugins/xsrf/xsrf_test.go
```diff
@@ -30,30 +31,25 @@ func (userIdentifier) UserID(r *safehttp.IncomingRequest) (string, error) {
...
 			wantHeader: map[string][]string{},
 			wantBody:   "",
 		},
 		{
-			name:       "Invalid host in token generation",
+			name:       "Token mismatch for invalid host",
 			userID:     "1234",
-			host:       "bar.com",
-			path:       "/pizza",
+			actionID:   "https://bar.com/pizza",
```
mattiasgrenfeldt, line 52: Since `actionID` is just a string you could have just one test for "invalid actionID".
It feels like testing for invalid host, path and scheme is a little to redundant.

safehttp/plugins/xsrf/xsrf_test.go
```diff
@@ -163,87 +91,204 @@ func TestXSRFTokenMultipart(t *testing.T) {
...
-		if rec.Status() != test.wantStatus {
-			t.Errorf("response status: got %v, want %v", rec.Status(), test.wantStatus)
-		}
-		if diff := cmp.Diff(test.wantHeader, map[string][]string(rec.Header())); diff != "" {
-			t.Errorf("rw.header mismatch (-want +got):\n%s", diff)
+				i.Before(rec.ResponseWriter, req)
+
+				if rec.Status() != test.wantStatus {
+					t.Errorf("response status: got %v, want %v", rec.Status(), test.wantStatus)
+				}
```
mattiasgrenfeldt, line 110: It's probably good to have a `got` here since you are making a function call. So that we are 100% sure that the thing being compared is the same as the thing being printed.

The error message could also start with `"rec.Status(): ..."`

safehttp/plugins/xsrf/xsrf_test.go
```diff
@@ -163,87 +91,204 @@ func TestXSRFTokenMultipart(t *testing.T) {
...
+				if diff := cmp.Diff(test.wantHeader, map[string][]string(rec.Header())); diff != "" {
+					t.Errorf("rec.Header() mismatch (-want +got):\n%s", diff)
+				}
+				if got := rec.Body(); got != test.wantBody {
+					t.Errorf("response body: got %q want %q", got, test.wantBody)
```
mattiasgrenfeldt, line 115: This error message could start with `"rec.Body(): ..."`

safehttp/plugins/xsrf/xsrf_test.go
```diff
@@ -163,87 +91,204 @@ func TestXSRFTokenMultipart(t *testing.T) {
...
 		}
-		if got := rec.Body(); got != test.wantBody {
-			t.Errorf("response body: got %q want %q", got, test.wantBody)
+		if want, got := "Unauthorized\n", rec.Body(); got != want {
+			t.Errorf("response body: got %q want %q", got, want)
 		}
 	}
 }
+
+func TestTokenRequestContext(t *testing.T) {
```
mattiasgrenfeldt, line 223: It feels like this test is testing `xsrftoken.Valid` and not `xsrf.Token`. We are already testing `xsrftoken.Valid` in `TestTokenPost` and `TestTokenMultipart`. I would suggest replacing this with two small tests. One which creates a `IncomingRequest`, places something in the context and then retrieves it with `xsrf.Token`. And another test in which you send a request to `Before`, just like you do in this test and checks afterwards if the `IncomingRequest` contains the token, without actually validating the token.

mattiasgrenfeldt (changes requested):
I really like how you structured `TestTokenPost` and `TestTokenMultipart`! They are very easy to read :)

safehttp/plugins/xsrf/xsrf_test.go
mihalimara22: We should make sure all our tests use either `response status` and `response body` or the function names.

mattiasgrenfeldt: Agreed, and I think it should be the function names

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -34,40 +35,64 @@ type UserIdentifier interface {
...
-		mf, err := r.MultipartForm(32 << 20)
+
+	tok := xsrftoken.Generate(i.AppKey, userID, r.URL.String())
+	r.SetContext(context.WithValue(r.Context(), tokenCtxKey{}, tok))
+
+	stateChangingMethods := map[string]bool{
+		safehttp.MethodPost:  true,
+		safehttp.MethodPatch: true,
+	}
+	if m := r.Method(); stateChangingMethods[m] {
```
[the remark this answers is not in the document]
mihalimara22: AFAIU, you can't make PUT and DELETE requests  from HTML forms. We also don't allow cross-site state changing requests unless directed to an endpoint where CORS is specifically allowed, in which case the Fetch Metedata policies don't take effect. So the only requests we expect to contain tokens in the body are POST and PATCH requests.

safehttp/plugins/xsrf/xsrf_test.go
mihalimara22: Opened #126

safehttp/plugins/xsrf/xsrf.go
mihalimara22: Well, you could have multiple `Identifier`s for the same application I guess, but you might want the key to be unique across the application, IDK. @empijei, @kele WDYT?

mattiasgrenfeldt (approved):

safehttp/plugins/xsrf/xsrf.go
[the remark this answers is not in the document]
mattiasgrenfeldt: Right, but can't you make PUT and DELETE requests using JS and `fetch` to your own site. Shouldn't those requests contain XSRF tokens? Maybe not, because if they would be made cross-site then that would trigger a CORS preflight(?), which we won't handle.

I don't feel like I have enough context to judge this, so I trust what you've made. :) :+1: If there is a problem then hopefully @empijei and @kele will notice it.

empijei (changes requested):

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -25,6 +26,11 @@ const (
 	TokenKey = "xsrf-token"
 )
 
+var stateChangingMethods = map[string]bool{
+	safehttp.MethodPost:  true,
+	safehttp.MethodPatch: true,
+}
```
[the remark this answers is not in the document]
mihalimara22: My understanding was that only POST and PATCH requests can make requests from HTML forms, but will modify.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -34,24 +42,47 @@ type UserIdentifier interface {
...
 type Interceptor struct {
 	AppKey     string
 	Identifier UserIdentifier
 }
 
+type tokenCtxKey struct{}
+
+// Token returns the XSRF token from the safehttp.IncomingRequest Context and a
+// nil error, if present, and, otherwise, returns an empty string and a non-nil
+// error.
```
kele, line 58: "Token extract the XSRF token from the incoming request. If it is not present, returns a non-nil error." is enough.

- no need to write `safehttp.IncomingRequest` or that it gets the token from it's `Context`
- in Go, it is the convention that nil error indicates success, so no need to write about it
- in Go, you usually assume that if one of the return values is a non-nil error, you cannot depend on other return values

In general, try to think about comments in a more declarative, than imperative way. For instance, imagine a function `func MinInt(a, b int) int`. You have two ways of commenting:

1. "MinInt takes two integers and if the first one is smaller or equal to the second one, it returns the first. If the second one is smaller, it returns the second."
2. "MinInt takes two integers and returns the smaller one."

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -34,24 +42,47 @@ type UserIdentifier interface {
...
+func Token(r *safehttp.IncomingRequest) (string, error) {
+	tok := r.Context().Value(tokenCtxKey{})
+	if tok == nil {
+		return "", errors.New("xsrf token not found in safehttp.IncomingRequest context")
```
kele, line 62: Too much detail, just "XSRF token not found".

Try to think: "what should the user do with the error"? In this case, there is nothing they can do about this extra information. Whether this is stored in context or somewhere else is an internal implementation decision of the `xsrf` package.

safehttp/plugins/xsrf/xsrf.go
kele: Wholeheartedly agree with this sentiment. @mihalimara22, could you do this?

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -34,24 +42,47 @@ type UserIdentifier interface {
...
+}
+
 // Before should be executed before directing the safehttp.IncomingRequest to
-// the handler to ensure it is not part of the Cross Site Request
-// Forgery. It checks for the presence of an xsrf-token in the request body and
-// validates it based on the userID associated with the request.
-func (p *Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
-	userID, err := p.Identifier.UserID(r)
+// the handler to ensure it is not part of the Cross-Site Request
+// Forgery. In case of state changing methods, it checks for the
```
kele, line 69: How does the user of this package know which are state changing and which are not?

kele (changes requested):

safehttp/plugins/xsrf/xsrf.go
mihalimara22: Added an explanation to the godoc

mihalimara22:
Rebased on top of master.

safehttp/incoming_request.go
```diff
@@ -44,19 +44,9 @@ func NewIncomingRequest(req *http.Request) *IncomingRequest {
...
-// Path returns the relative path of the URL in decoded format (e.g. %47%6f%2f
-// becomes /Go/).
-// TODO(@mihalimara22): Remove this after the safehttp.URL type has been     //
-// implemented.
-func (r *IncomingRequest) Path() string {
-	return r.req.URL.Path
+// Method specifies the HTTP method of an IncomingRequest.
+func (r *IncomingRequest) Method() string {
+	return r.req.Method
 }
```
kele, line 50: For the future, avoid putting unrelated changes in a PR.

safehttp/plugins/csp/csp_test.go
```diff
@@ -193,7 +193,7 @@ func TestAlreadyClaimed(t *testing.T) {
 			it := Interceptor{}
 			it.Before(rr.ResponseWriter, req)
 
-			if got, want := rr.Status(), int(safehttp.StatusInternalServerError); got != want {
```
kele, line 196: Same here.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -34,24 +42,46 @@ type UserIdentifier interface {
 }
 
 // Interceptor implements XSRF protection. It requires an application key and a
-// storage service. The appKey uniquely identifies each registered service and
+// storage service. The AppKey uniquely identifies each registered service and
 // should have high entropy. The storage service supports retrieving ID's of the
-// application's users. Both the appKey and user ID are used in the XSRF
-// token generation algorithm.
+// application's users. Both the AppKey and user ID are used in the XSRF
+// token generation and verification algorithm.
```
kele, line 48: The comment mentions "storage service" but there's none to be found. Could you update the comment?

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -34,24 +42,46 @@ type UserIdentifier interface {
...
-// Forgery. It checks for the presence of an xsrf-token in the request body and
-// validates it based on the userID associated with the request.
-func (p *Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
-	userID, err := p.Identifier.UserID(r)
+// the handler to ensure it is not part of the Cross-Site Request
+// Forgery. In case of state changing requests (all except GET, HEAD and OPTIONS
+// requests), it checks for the presence of an xsrf-token in the request body
+// and validates it based on the userID associated with the request. For all
+// requests, it adds a cryptographically safe XSRF token to the
+// safehttp.IncomingRequest context.
```
kele, line 72:
```suggestion
// the handler to ensure it is not part of a Cross-Site Request Forgery attack.
//
// In case of state changing requests (all except GET, HEAD and OPTIONS),
// it checks for the presence of an XSRF token in the request
// and validates it based on the user ID associated with the request.
//
// For authorized requests, it adds a cryptographically safe XSRF token to the
// incoming request. It can be later extracted using Token.
```

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -34,24 +42,46 @@ type UserIdentifier interface {
...
+func (i *Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+	userID, err := i.Identifier.UserID(r)
 	if err != nil {
 		return w.ClientError(safehttp.StatusUnauthorized)
 	}
+
+	if m := r.Method(); statePreservingMethods[m] {
```
kele, line 79:
```suggestion
	if statePreservingMethods[r.Method()] {
```

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -66,9 +96,12 @@ func (p *Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequ
 		return w.ClientError(safehttp.StatusUnauthorized)
 	}
 
-	if ok := xsrftoken.Valid(tok, p.AppKey, userID, r.Host()+r.Path()); !ok {
+	if ok := xsrftoken.Valid(tok, i.AppKey, userID, r.URL.String()); !ok {
```
kele, line 99: What prompted this change?

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -34,24 +42,46 @@ type UserIdentifier interface {
...
+// Forgery. In case of state changing requests (all except GET, HEAD and OPTIONS
+// requests), it checks for the presence of an xsrf-token in the request body
+// and validates it based on the userID associated with the request. For all
+// requests, it adds a cryptographically safe XSRF token to the
+// safehttp.IncomingRequest context.
+func (i *Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+	userID, err := i.Identifier.UserID(r)
 	if err != nil {
 		return w.ClientError(safehttp.StatusUnauthorized)
 	}
```
kele, line 77: Alternatively you can (and probably should) do:

```
needsValidation := !statePreservingMethods[r.Method]

if needsValidation {
  // do the token extraction and validation
}

tok = xsrftoken.Generate(i.AppKey, userID, r.URL.String())
r.SetContext(context.WithValue(r.Context(), tokenCtxKey{}, tok))
return safehttp.Result{}
```

safehttp/plugins/xsrf/xsrf_test.go
```diff
@@ -87,164 +68,184 @@ func TestXSRFTokenPost(t *testing.T) {
...
 
-		if rec.Status() != test.wantStatus {
-			t.Errorf("response status: got %v, want %v", rec.Status(), test.wantStatus)
-		}
-		if diff := cmp.Diff(test.wantHeader, map[string][]string(rec.Header())); diff != "" {
-			t.Errorf("rec.Header() mismatch (-want +got):\n%s", diff)
-		}
-		if got := rec.Body(); got != test.wantBody {
-			t.Errorf("response body: got %q want %q", got, test.wantBody)
+				i.Before(rec.ResponseWriter, req)
```
kele, line 83: This is a nit, but try to somehow sort the unrelated lines. What would probably look better is:

1. Creating the Response Recorder.
2. Creating the token, creating the request.
3. Creating the interceptor and running it on the request.

This way you minimize the context needed for each line of code to be able to understand it.

safehttp/plugins/xsrf/xsrf_test.go
```diff
@@ -87,164 +68,184 @@ func TestXSRFTokenPost(t *testing.T) {
...
+				i := Interceptor{AppKey: "xsrf", Identifier: userIdentifier{}}
+				tok := xsrftoken.Generate("xsrf", test.userID, test.actionID)
+				b := "--123\r\n" +
+					"Content-Disposition: form-data; name=\"xsrf-token\"\r\n" +
+					"\r\n" +
+					tok + "\r\n" +
+					"--123--\r\n"
+				req := safehttptest.NewRequest(method, "https://foo.com/pizza", strings.NewReader(b))
+				req.Header.Set("Content-Type", `multipart/form-data; boundary="123"`)
+				rec := safehttptest.NewResponseRecorder()
```
kele, line 112: Similar comments as above.

safehttp/plugins/xsrf/xsrf_test.go
```diff
@@ -87,164 +68,184 @@ func TestXSRFTokenPost(t *testing.T) {
...
-			},
-			wantBody: "Unauthorized\n",
 		},
 	}
 	for _, test := range tests {
-		i := xsrf.Interceptor{AppKey: "xsrf", Identifier: userIdentifier{}}
+		i := Interceptor{AppKey: "xsrf", Identifier: userIdentifier{}}
 		rec := safehttptest.NewResponseRecorder()
+
 		i.Before(rec.ResponseWriter, test.req)
```
kele, line 182: Similar comments as above.

kele (changes requested):

safehttp/plugins/csp/csp_test.go
mihalimara22: This was needed for checks not to fail.

safehttp/plugins/xsrf/xsrf.go
mihalimara22: The `actionID` should indicate the action the user is taking and having the URL seemed better than doing the string concatenation twice as before. Alternatively, we can also add the HTTP method used.

kele: The URL is not just host + path, so you need to be careful here. It's also query params and and fragments, right?

safehttp/plugins/xsrf/xsrf.go
mihalimara22: Initially, that's the way it was (and I prefer it as well because I don't need to do token generation twice), but Rob suggested moving token verification outside the if-clause so I don't have most of the function logic indented. Which one do you prefer? @kele @empijei

safehttp/plugins/xsrf/xsrf.go
mihalimara22: Oh right, that could introduce problems, I will follow the example `xsrftoken` gives the `Generate` godoc (HTTP method and path). WDYT?

safehttp/plugins/xsrf/xsrf.go
kele: I don't have a strong preference. You can also have a `Interceptor.extractToken()` method and have less lines in this one.

safehttp/plugins/xsrf/xsrf.go
kele: I'd keep the host too.

@empijei WDYT?

safehttp/plugins/xsrf/xsrf_test.go
```diff
@@ -71,59 +70,55 @@ var (
...
-				}
-				if got := rec.Body(); got != test.wantBody {
-					t.Errorf("response body: got %q want %q", got, test.wantBody)
-				}
-			})
-		}
+	for _, test := range formTokenTests {
+		t.Run(test.name, func(t *testing.T) {
+			rec := safehttptest.NewResponseRecorder()
+			tok := xsrftoken.Generate("xsrf", test.userID, test.actionID)
```
kele, line 76: BTW. Maybe use a different app key than "xsrf"? Maybe "testAppKey".

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -33,42 +41,68 @@ type UserIdentifier interface {
...
-	if err != nil {
-		mf, err := r.MultipartForm(32 << 20)
+
+	actionID := r.Method() + " " + r.URL.Path()
+	needsValidation := !statePreservingMethods[r.Method()]
+	if needsValidation {
+		f, err := r.PostForm()
 		if err != nil {
-			return w.ClientError(safehttp.StatusBadRequest)
+			mf, err := r.MultipartForm(32 << 20)
```
kele, line 87: Add a comment about why we're falling back to MultipartForm (we fallback because from our perspective it doesn't matter what's the type of the form, as long as the XSRF token is there).

safehttp/plugins/xsrf/xsrf_test.go
kele: Please update. Here and all other tests.

kele (approved):
A few nits.

kele:
Make sure to rebase before rerequesting a review.

safehttp/plugins/xsrf/xsrf_test.go
kele: Again, please fix.

kele (approved):
Well done! :)

safehttp/plugins/xsrf/xsrf.go
kele: @mihalimara22 Please file an issue about this and resolve. I can merge then. :)

safehttp/plugins/xsrf/xsrf.go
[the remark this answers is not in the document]
kele: Mara has resolved this.

safehttp/plugins/xsrf/xsrf.go
empijei: factoring it out in a function works for me, otherwise I prefer to have it unindented.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -33,42 +41,71 @@ type UserIdentifier interface {
...
-// Interceptor implements XSRF protection. It requires an application key and a
-// storage service. The appKey uniquely identifies each registered service and
-// should have high entropy. The storage service supports retrieving ID's of the
-// application's users. Both the appKey and user ID are used in the XSRF
-// token generation algorithm.
+// Interceptor implements XSRF protection.
 type Interceptor struct {
-	AppKey     string
+	// AppKey uniquely identifies each registered service and should have high
+	// entropy as it is used for generating the XSRF token.
```
empijei, line 47: I would say that this should be secret, and to do that I would rename the field to SecretAppKey.

empijei (approved):

PR #124, opened 2020-08-27
Now we use safehttp.MethodXxxx and safehttp.StatusXxxx everywhere instead of strings and ints

### LINKED ISSUE #121: Cleanup tests to only use safehttp.StatusCode and the request methods provided in mux.go
Currently, some of the frameworks tests don't use `safehttp.StatusCode`s and `safehttp.MethodGet`,  `safehttp.MethodHead`, etc but rather `string`s and `int`s. We should clean up tests for uniformity.

mattiasgrenfeldt:
Fixes #121

### DIFF
--- a/safehttp/form_test.go
+++ b/safehttp/form_test.go
@@ -90,7 +90,7 @@ func TestFormValidInt(t *testing.T) {
 		rec := newResponseRecorder(&strings.Builder{})
 		mux.ServeHTTP(rec, test.req)
 
-		if want := 200; rec.status != want {
+		if want := safehttp.StatusOK; rec.status != want {
 			t.Errorf("response status: got %v, want %v", rec.status, want)
 		}
 	}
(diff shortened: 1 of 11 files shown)


mihalimara22 (commented):
What do you think about having the `safehttptest.ResponseRecorder.Status()` also returning a `safehttp.StatusCode` since we are doing the casting in tests anyway.

mattiasgrenfeldt:
That has already been fixed in #106, so we don't need to fix it here.

mihalimara22 (changes requested):
You've forgot the other files in integration tests that contain a `responseRecorder` whose `status` field is an `int`. Please fix them as well in this PR.

mihalimara22 (approved):

kele:
Sync to master and we can merge. Good job! :)

kele (approved):

mattiasgrenfeldt:
Done! @kele

bsiegert (approved):

PR #125, opened 2020-08-28
Splitting form_test.go so that it only contains unit tests for the Form type and adding unit tests for the IncomingRequest Type separately.

### LINKED ISSUE #57: form_test.go tests the functionality of both form.go and incoming_request.go, this should be split.
`safehttp/form_test.go` tests both `safehttp/form.go` and the functions `PostForm` and `MultipartForm` in `safehttp/incoming_request.go`. I think this should be split into two test files `form_test.go` and `incoming_request_test.go` testing the functionality implemented in the files separately. Integration tests could also be added to test the combined functionality of both files.

mattiasgrenfeldt:
Fixes #57 

**Test Coverage**
- form.go: 73% -> 100%
- incoming_request.go: 69.8% -> 95.3%

**Lines**
- Before: form_test.go + incoming_request_test.go = 1147 + 196 =  1343
- After: form_test.go + incoming_request_test.go = 488 + 360 = 848
- Diff: -495


safehttp/form_test.go
```diff
@@ -12,1135 +12,476 @@
...
-						"--123--\r\n"
-					multipartReq := httptest.NewRequest("POST", "http://foo.com/", strings.NewReader(multipartReqBody))
-					multipartReq.Header.Set("Content-Type", `multipart/form-data; boundary="123"`)
-					return multipartReq
-				}(),
-			},
-			want: 0,
-		},
+func TestFormInvalidInt64(t *testing.T) {
+	tests := []string{
```
mihalimara22, line 52: Please add names to this tests so it's clearer on what input the test failed, same below.

safehttp/form_test.go
```diff
@@ -12,1135 +12,476 @@
...
-			}
-
-			if diff := cmp.Diff(tc.want, tc.placeholder); diff != "" {
-				t.Errorf("form.Slice: diff (-want +got): \n%v", diff)
-			}
+			f.Slice("x", tt.placeholder)
 
-			return safehttp.Result{}
-		}))
+			// TODO: add a check here that tt.placeholder is nil. I (grenfeldt@)
```
mihalimara22, line 348: Keep the placeholder as an interface and do a `cmp.Diff` with a nil interface (`var want interface{}`).

safehttp/form_test.go
```diff
@@ -12,1135 +12,476 @@
...
-		"{{ range . -}}" +
-			"--123\r\n" +
-			"Content-Disposition: form-data; name=\"pizza\"\r\n" +
-			"\r\n" +
-			"{{ . }}\r\n" +
-			"{{ end }}" +
-			"--123--\r\n"))
+func TestFormSliceInvalidPointerType(t *testing.T) {
+	f := Form{values: map[string][]string{"x": {"x"}}}
+	f.Slice("x", &[]int16{})
```
mihalimara22, line 360: Make sure the result is nil here.

safehttp/incoming_request_test.go
```diff
@@ -16,11 +16,14 @@ package safehttp_test
...
 	"net/http/httptest"
+	"strings"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-safeweb/safehttp"
+	"github.com/google/go-safeweb/safehttp/safehttptest"
 )
 
 func TestIncomingRequestCookie(t *testing.T) {
```
mihalimara22, line 29: We have some tests names in which we include the file name `TestIncomingRequest...` and some, especially in plugins, where we don't. We should also probably make this uniform.

safehttp/form_test.go
```diff
@@ -12,1135 +12,476 @@
...
-					stringMaxInt + "\r\n" +
-					"--123--\r\n"
-				multipartReq := httptest.NewRequest("POST", "http://foo.com/", strings.NewReader(multipartReqBody))
-				multipartReq.Header.Set("Content-Type", `multipart/form-data; boundary="123"`)
-				return multipartReq
-			}(),
-			want: math.MaxInt64,
-		},
+		{val: "123", want: 123},
+		{val: "9223372036854775807", want: math.MaxInt64},
```
mihalimara22, line 31: This is optional, but I'd use `strconv...` to make the input, same for other tests.

mihalimara22 (changes requested):
Overall looks good, thanks for doing this. :)

safehttp/incoming_request_test.go
mattiasgrenfeldt: I think the way we are doing it now works well. There will be many tests in `safehttp`, therefore the test names needs to be prefixed with what exactly they are testing. But in the plugins there aren't as many tests that might overlap. So there we can have shorter test names, which is easier to read.

safehttp/form_test.go
mattiasgrenfeldt: I don't agree, that's harder to read.

safehttp/form_test.go
mattiasgrenfeldt: `f.Slice` doesn't set pointers of unknown type to nil.

safehttp/form_test.go
mattiasgrenfeldt: So, this is very tricky. If i have something like this:
```go
var want interface{}
diff := cmp.Diff(want, tt.placeholder)
```
I get the following diff:
```
form_test.go:363: DIFF (-want +got)  interface{}(
            + 	&[]int64(nil),
              )
```
If I try to create a `&[]int(nil)` to compare with in the table then it complains that I "can't take the address of nil".
If I try to create a `[]int(nil)` to compare with then that works.
```go
diff := cmp.Diff(&tt.want, tt.placeholder)
```
But then I get the following diff which compares pointer addresses :/
```
form_test.go:363: DIFF (-want +got)   interface{}(
            - 	&⟪0xc000066768⟫[]int(nil),
            +  &⟪0xc00000f120⟫[]int(nil),
              )
```
And if instead to try dereference `tt.placeholder` then that is really annoying as well since you have to do a bunch of type assertions.
Maybe there is some `cmpopt` that let's me compare pointers without comparing their addresses. I'll look into that.
Does @empijei or @kele have any tips on this?

mihalimara22: Played around with it a bit and realised the solution I thought I had from some weeks ago doesn't work :(

safehttp/form_test.go
mihalimara22: Ah right, sorry, my bad.

safehttp/form_test.go
kele: I'm not sure I get why you want a `nil` placeholder check. What's the rationale?

kele:
While I agree that having these functions unit tested is good, I think some of the "integration" tests should stay.

This is quite important because just having a simple `a=b` thing in the request parsing + having more complex values than `b` in form parsing from strings is not enough (even though test coverage increases). The reason is: what if the key/value extraction phase of the parser struggles with things like `-123`? What if in this case, `a=-123` the parser messes up and returns `("a", "123")`? I understand this is not the case now, but the goal was to have these tests check end-to-end behavior.

safehttp/form_test.go
mattiasgrenfeldt: `f.Slice` will set the slice pointer to `nil` if the data was invalid, I want to check that it does.

mattiasgrenfeldt:
I see. I thought this parsing was tested by `tests/formparams/form_params_test.go`. I don't test this parsing in `safehttp/header_test.go` either but I have the tests in `tests/headers/*` to cover how they are actually parsed into a map. I see 3 options:

1. I add back some integration tests.
2. We close this and open an issue to add this later.
3. We drop the changes in this PR.

WDYT?

kele:
2) SGTM. Please sync and I can merge.

kele (approved):

safehttp/incoming_request.go
```diff
@@ -96,13 +97,13 @@ func (r *IncomingRequest) PostForm() (*Form, error) {
...
-	// passed as part of the request. If this is bigger than 32 MB, the rest
-	// will be stored on disk.
-	const defaultMaxMemory = 32 << 20
 	r.multipartParseOnce.Do(func() {
-		if r.req.Method != "POST" && r.req.Method != "PATCH" && r.req.Method != "PUT" {
-			err = fmt.Errorf("got request method %s, want POST/PATCH/PUT", r.req.Method)
+		// Ensures no more than 32 MB are stored in memory when a form file is
+		// passed as part of the request. If this is bigger than 32 MB, the rest
+		// will be stored on disk.
+		const defaultMaxMemory = 32 << 20
```
empijei, line 104: If you want you can move this closer to its use, on line 113

empijei (approved):

safehttp/incoming_request.go
mattiasgrenfeldt: This is getting removed in #135 so i'm not gonna bother.

mattiasgrenfeldt:
Done! @kele

mattiasgrenfeldt:
This is ready to be merged @kele @empijei

PR #131, opened 2020-08-31
Removed casting in CSP tests that is breaking master

mihalimara22:
Fixed commit #5c6aabe that is causing `go vet` and `go test` to fail because of a type mismatich (`int` to `safehttp.StatusCode` comparison in the CSP tests).


kele:
Thanks for a quick fix!

kele (approved):

PR #134, opened 2020-09-01
Added set of hashes to CSP plugin

### LINKED ISSUE #111: Add a set of hashes to the CSP plugin
The `script-src` directive in CSP can contain hashes of scripts that will be allowed. This should be added to the CSP plugin.

More info here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src

I will add this after #93 has been merged.

mattiasgrenfeldt:
Fixes #111 

You can now specify a set of hashes in the CSP plugin. An example of such a hash is `sha256-CihokcEcBW4atb/CW/XWsvWwbTjqwQlE9nj9ii5ww5M=`

### DIFF
--- a/safehttp/plugins/csp/csp.go
+++ b/safehttp/plugins/csp/csp.go
@@ -84,6 +84,12 @@ type StrictCSPBuilder struct {
 	// ReportURI controls the report-uri directive. If ReportUri is empty, no report-uri
 	// directive will be set.
 	ReportURI string
+	// Hashes adds a set of hashes to script-src. An example of a hash would be:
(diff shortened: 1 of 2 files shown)


mihalimara22:
Can you explain why we need this set of hashes in the PR description?

mihalimara22 (approved):

mattiasgrenfeldt:
Well I'm not actually sure why we need the set of hashes when we have the nonces. @empijei who wanted me to add this maybe has some insight?

I think having a hash instead of using a nonce works just as well security-wise, but you need to update the hashes all the time when the scripts change. So therefore nonces are more convenient. With hashes you also have to send much data to the client on each response.

empijei (approved):

PR #138, opened 2020-09-02
Improving Header ergonomics by removing error return value from Add, Set, Del and Claim and panicking instead.

### LINKED ISSUE #119: Improve ergonomics of the header type
Currently the header API is very hard to use:

```go
	wh := w.Header()
	allowMethods, err := wh.Claim("Access-Control-Allow-Methods")
	if err != nil {
		return w.ServerError(safehttp.StatusInternalServerError)
	}

	headers := rh.Get("Access-Control-Request-Headers")
	if headers != "" {
		for _, h := range strings.Split(headers, ", ") {
			h = textproto.CanonicalMIMEHeaderKey(h)
			if !it.allowedHeaders[h] && h != customHeader {
				return w.ClientError(safehttp.StatusForbidden)
			}
		}
	}
	allowHeaders, err := wh.Claim("Access-Control-Allow-Headers")
	if err != nil {
		return w.ServerError(safehttp.StatusInternalServerError)
	}

	maxAge, err := wh.Claim("Access-Control-Max-Age")
	if err != nil {
		return w.ServerError(safehttp.StatusInternalServerError)
	}
```

We should make this simpler to use and access.

mattiasgrenfeldt:
Fixes #119 

- Error return values have been removed from `Add`, `Set`, `Del` and `Claim`. They instead panic if an error occurs.
- Added a `IsClaimed` method.


safehttp/header.go
```diff
@@ -39,61 +39,66 @@ func newHeader(h http.Header) Header {
 // which can be used to set the header. The name is first canonicalized
 // using textproto.CanonicalMIMEHeaderKey. Other methods in
 // the struct can't write to, change or delete the header with this
-// name. These methods will instead fail when applied on an claimed
+// name. These methods will instead panic when applied on an claimed
```
kele, line 42: On *a* claimed

safehttp/header.go
```diff
@@ -39,61 +39,66 @@ func newHeader(h http.Header) Header {
...
 			return
 		}
 		h.wrapped[name] = v
-	}, nil
+	}
+}
+
+// IsClaimed returns whether the provided header is already claimed. The name is
+// first canonicalized using textproto.CanonicalMIMEHeaderKey. The Set-Cookie header
+// is always pre-claimed.
```
kele, line 61: treated as claimed.

kele (approved):

safehttp/header.go
```diff
@@ -39,61 +39,66 @@ func newHeader(h http.Header) Header {
...
 	return func(v []string) {
 		if v == nil {
 			return
 		}
 		h.wrapped[name] = v
-	}, nil
+	}
+}
+
+// IsClaimed returns whether the provided header is already claimed. The name is
```
mihalimara22, line 59: reports rather than returns maybe?

safehttp/header.go
```diff
@@ -39,61 +39,66 @@ func newHeader(h http.Header) Header {
...
+	return err != nil
 }
 
 // Set sets the header with the given name to the given value.
 // The name is first canonicalized using textproto.CanonicalMIMEHeaderKey.
 // This method first removes all other values associated with this
-// header before setting the new value. Returns an error when
-// applied on claimed headers or on the Set-Cookie header.
-func (h Header) Set(name, value string) error {
+// header before setting the new value. Panics when applied on claimed headers
```
mihalimara22, line 71: "It panics .." or "and panics..", same for below.

mihalimara22 (approved):

Issue #136, opened 2020-09-01
Decide what the actionID should be in the XSRF plugin


maramihali:
The `actionID` is defined as "the action the user is taking" in the `xsrftoken` package. We use this package for token generation and verification in the XSRF plugin. Currently, this is build as the HTTP method and Path from the incoming request. We should decide whether to leave it as it is or change it (e.g. to also include the Host).

PR #139, opened 2020-09-02
Added pre-login XSRF protection to the XSRF plugin

### LINKED ISSUE #130: Add pre-login XSRF protection
The XSRF plugin will need to protect login forms from attacks as this can also introduce vulnerabilities See [this link](https://stackoverflow.com/questions/6412813/do-login-forms-need-tokens-against-csrf-attacks) for more context.

maramihali:
Fixes #130, #136 

The XSRF protection can now also protect login forms by setting a custom cookie to distinguish between users. The cookie will then be used as an identifier in the verification and generation algorithm.

### DIFF
--- a/safehttp/plugins/xsrf/xsrf.go
+++ b/safehttp/plugins/xsrf/xsrf.go
@@ -58,9 +58,8 @@ func Token(r *safehttp.IncomingRequest) (string, error) {
 
 func addCookieID(w *safehttp.ResponseWriter) (*safehttp.Cookie, error) {
 	buf := make([]byte, 20)
-	_, err := rand.Read(buf)
-	if err != nil {
-		panic(fmt.Errorf("crypto/rand.Read: %v", err))
+	if _, err := rand.Read(buf); err != nil {
+		return nil, fmt.Errorf("crypto/rand.Read: %v", err)
 	}
 
 	c := safehttp.NewCookie(cookieIDKey, base64.StdEncoding.EncodeToString(buf))
--- a/safehttp/plugins/xsrf/xsrf.go
+++ b/safehttp/plugins/xsrf/xsrf.go
@@ -16,8 +16,10 @@ package xsrf
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/base64"
 	"errors"
-
+	"fmt"
 	"github.com/google/go-safeweb/safehttp"
 	"golang.org/x/net/xsrftoken"
 )
@@ -25,7 +27,8 @@ import (
 const (
 	// TokenKey is the form key used when sending the token as part of POST
 	// request.
-	TokenKey = "xsrf-token"
+	TokenKey    = "xsrf-token"
+	tokenCookie = "xsrf-cookie"
 )
 
 var statePreservingMethods = map[string]bool{
@@ -34,22 +37,11 @@ var statePreservingMethods = map[string]bool{
 	safehttp.MethodOptions: true,
 }
 
-// UserIdentifier provides the web application users' identifiers,
-// needed in generating the XSRF token.
-type UserIdentifier interface {
-	// UserID returns the user's identifier based on the
-	// safehttp.IncomingRequest received.
-	UserID(*safehttp.IncomingRequest) (string, error)
-}
-
 // Interceptor implements XSRF protection.
 type Interceptor struct {
-	// SecretAppKey uniquely identifies each registered service and should have high
-	// entropy as it is used for generating the XSRF token.
+	// SecretAppKey uniquely identifies each registered service and should have
+	// high entropy as it is used for generating the XSRF token.
 	SecretAppKey string
-	// Identifier supports retrieving the user ID based on the incoming
-	// request. This is needed for generating the XSRF token.
-	Identifier UserIdentifier
 }
 
 type tokenCtxKey struct{}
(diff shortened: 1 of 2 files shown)


mattiasgrenfeldt (approved):
One nit :+1:

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -63,24 +56,52 @@ func Token(r *safehttp.IncomingRequest) (string, error) {
...
 // In case of state changing requests (all except GET, HEAD and OPTIONS), it
 // checks for the presence of an XSRF token in the request and validates it
-// based on the user ID associated with the request.
+// based on the custom cookie.
 //
 // For authorized requests, it adds a cryptographically safe XSRF token to the
-// incoming request. It can be later extracted using Token.
+// incoming request. This can be later extracted using Token.
 func (i *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest, cfg interface{}) safehttp.Result {
-	userID, err := i.Identifier.UserID(r)
```
kele, line 77: I don't get this part. Why is this removed?

mihalimara22: So, the previous plugin version was vulnerable to pre-login CSRF. An attack would go as follows: I am logged into your bank account and the attacker also has an actual account on the banking website. Then, the attacker can forge a login request in my browser and trick me into using their account. Then, they'll have access to stuff like my transaction history while I am logged in into their account.
 
This would be allowed by the previous plugin because the call to `Identifier.UserID(r)` request would return the ID corresponding to the attacker account even if I am the one navigating and the token that is injected in the form has also been generated for the attacker's account.

Therefore, we need a way to identify that I am the one navigating the attacker's account. One way to do this is: the first time I navigate the banking website with a state preserving request, a noonce-based cookie is generated with SameSite Strict. Then, any xsrf token validation fails because the token was generated with the attacker's cookie and we validate with our cookie.

Hopefully this made sense (@empijei please correct me if I misunderstood something).

kele: But now you've taken away the feature of extracting a user ID from a request, possibly via some lookup operations.

mihalimara22: AFAIU that is not needed anymore.

kele: Discussed this with @empijei offline.

Long story short:
- you're right, we don't really need the `UserIdentifier` - it wasn't clear to me why
- please document why need both the cookies + token injection in the response

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -63,24 +56,52 @@ func Token(r *safehttp.IncomingRequest) (string, error) {
...
+	needsValidation := !statePreservingMethods[r.Method()]
+	c, err := r.Cookie(tokenCookie)
```
kele, line 91: The "tokenCookie" doesn't really tell match what it is. In reality, it is some kind of a user ID / session ID. Could you try to find a better name?

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -63,24 +56,52 @@ func Token(r *safehttp.IncomingRequest) (string, error) {
 	return tok.(string), nil
 }
 
+func addTokenCookie(w *safehttp.ResponseWriter) (*safehttp.Cookie, error) {
+	buf := make([]byte, 20)
+	_, err := rand.Read(buf)
+	if err != nil {
+		panic(fmt.Errorf("failed to generate entropy using crypto/rand/RandReader: %v", err))
```
kele, line 63:
```suggestion
		panic(fmt.Errorf("crypto/rand.Read: %v", err))
```

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -63,24 +56,52 @@ func Token(r *safehttp.IncomingRequest) (string, error) {
...
+		if err != nil {
+			// An error is returned when the plugin fails to Set the Set-Cookie
+			// header in the response writer as this is a server misconfiguration.
+			return w.ServerError(safehttp.StatusInternalServerError)
+		}
 	}
 
-	actionID := r.Method() + " " + r.URL.Path()
-	needsValidation := !statePreservingMethods[r.Method()]
+	actionID := r.URL.Path()
```
kele, line 104: Why this change?

safehttp/plugins/xsrf/xsrf.go
kele, line 91: Consider a longer name than "c", it is used over ~30 lines.

kele (changes requested):

safehttp/plugins/xsrf/xsrf.go
maramihali: Having the method there is error-prone since we could visit the same path using different HTTP methods. I discussed offline with @empijei who said that it's fine if the `actionID` is just the path.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -63,24 +56,55 @@ func Token(r *safehttp.IncomingRequest) (string, error) {
 	return tok.(string), nil
 }
 
+func addCookieID(w *safehttp.ResponseWriter) (*safehttp.Cookie, error) {
+	buf := make([]byte, 20)
+	_, err := rand.Read(buf)
+	if err != nil {
+		panic(fmt.Errorf("crypto/rand.Read: %v", err))
```
bsiegert, line 63: I am nervous about this `panic` in the request path. Can this return the error instead?

bsiegert (approved):

safehttp/plugins/xsrf/xsrf.go
maramihali: I changed it to return an error instead, but shouldn't a system failure result in a panic?

empijei (approved):

kele (approved):
Thank you. :)

kele:
Please rebase and we can merge.

maramihali:
Rebased, ready for merge.

PR #144, opened 2020-09-03
Removed Interceptor keys from ServeMux, now interceptors and their configs are stored in a slice.

### LINKED ISSUE #143: Get rid of Interceptor keys in ServeMux, since they aren't needed.
Currently `Interceptors` are installed and stored using unique keys in the `ServeMux`. We have decided these keys aren't needed and should be removed. If interaction is needed between a `Handler` and an `Interceptor` then the interceptors needs to implement package level functions that take `ResponseWriter`s and `IncomingRequest`s and modify the context when state needs to be passed.

mattiasgrenfeldt:
Now Interceptors are stored together with their configs in slices instead of maps. To interact with an interceptor from the handler that interceptor needs to implement package level functions which take ResponseWriters and IncomingRequests and do things. State can be stored in the context of the request.

Fixes #143

### DIFF
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -202,8 +202,8 @@ func (h handlerWithInterceptors) ServeHTTP(w http.ResponseWriter, r *http.Reques
 		}
 	}()
 
-	for _, ai := range h.interceps {
-		ai.it.Before(rw, ir, ai.cfg)
+	for _, interceptor := range h.interceps {
+		interceptor.it.Before(rw, ir, interceptor.cfg)
 		if rw.written {
 			return
 		}
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -107,7 +107,7 @@ type appliedInterceptor struct {
 // Configs can be optionally passed in order to modify the behavior of the
 // interceptors on a registered handler. Passing a Config whose corresponding
 // Interceptor was not installed will produce no effect. If multiple Configs are
-// passed for the same Interceptor, only the first one is applied.
+// passed for the same Interceptor, only the first one will take effect.
 func (m *ServeMux) Handle(pattern string, method string, h Handler, cfgs ...Config) {
 	var interceps []appliedInterceptor
 	for _, it := range m.interceps {
@@ -121,9 +121,9 @@ func (m *ServeMux) Handle(pattern string, method string, h Handler, cfgs ...Conf
 		interceps = append(interceps, appliedInterceptor{it: it, cfg: cfg})
 	}
 	hi := handlerWithInterceptors{
-		handler:          h,
-		appliedInterceps: interceps,
-		disp:             m.disp,
+		handler:   h,
+		interceps: interceps,
+		disp:      m.disp,
 	}
 
 	mh, ok := m.handlers[pattern]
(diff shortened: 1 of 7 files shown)


safehttp/mux.go
```diff
@@ -92,35 +89,41 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
...
-		handlers:  map[string]methodHandler{},
-		interceps: map[string]Interceptor{},
+		mux:      http.NewServeMux(),
+		domains:  dm,
+		disp:     d,
+		handlers: map[string]methodHandler{},
 	}
 }
 
+type appliedInterceptor struct {
```
mihalimara22, line 99: Maybe `configuredInterceptor` or so? Also, add a brief comment to explain what this is (I know it's unexported but I think it would help people maintaining this package in the future).

safehttp/mux.go
```diff
@@ -92,35 +89,41 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
...
+}
+
 // Handle registers a handler for the given pattern and method. If another
 // handler is already registered for the same pattern and method, Handle panics.
 //
 // Configs can be optionally passed in order to modify the behavior of the
 // interceptors on a registered handler. Passing a Config whose corresponding
-// Interceptor was not installed will produce no effect.
+// Interceptor was not installed will produce no effect. If multiple Configs are
+// passed for the same Interceptor, only the first one is applied.
```
mihalimara22, line 110:
```suggestion
// passed for the same Interceptor, only the first one will take effect.
```

safehttp/mux.go
```diff
@@ -188,17 +182,15 @@ func (m methodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // handlerWithInterceptors encapsulates a handler and its corresponding
 // interceptors.
 type handlerWithInterceptors struct {
-	handler        Handler
-	configs        map[string]Config
-	interceps      map[string]Interceptor
-	intercepsOrder []string
-	disp           Dispatcher
+	handler          Handler
+	appliedInterceps []appliedInterceptor
```
mihalimara22, line 186: I think calling them `interceps` is enough here.

mihalimara22 (changes requested):

safehttp/mux.go
mattiasgrenfeldt: What needs to be explained in the comment? This is just a holder of data, it has no methods. There is no depth that needs to be explained. I think a comment would just add clutter in this case.

safehttp/mux.go
```diff
@@ -210,8 +202,8 @@ func (h handlerWithInterceptors) ServeHTTP(w http.ResponseWriter, r *http.Reques
 		}
 	}()
 
-	for _, key := range h.intercepsOrder {
-		h.interceps[key].Before(rw, ir, h.configs[key])
+	for _, ai := range h.interceps {
```
kele, line 205: `ai` isn't clear when you see `h.intercepts`. I'd suggest changing to `interceptor`.

safehttp/mux.go
```diff
@@ -92,35 +89,41 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
...
-				cfgMap[k] = cfg
+	var interceps []appliedInterceptor
+	for _, it := range m.interceps {
+		var cfg Config
+		for _, c := range cfgs {
+			if c.Match(it) {
+				cfg = c
+				break
 			}
 		}
```
kele, line 120: Could you file a feature request to ban adding two configurations to the same interceptor?

kele (changes requested):
Please mention in the PR description (and possibly title) that you've also updated how Configs are associated with interceptors, and that a map of interceptors have been removed in favor of a simple slice.

safehttp/mux.go
kele: For posterity: https://github.com/google/go-safeweb/issues/148.

kele (approved):

PR #146, opened 2020-09-03
Added the Commit phase

### LINKED ISSUE #145: Add a Commit phase to the Interceptor interface
As some plugins might want to modify/add headers or modify the body (e.g. CSP/XSRF nonce injection in HTML templates) exactly before the response is written, the interceptor interface needs to support a Commit phase.

mattiasgrenfeldt:
Fixes #145

Added the `Commit` phase to the Interceptor interface. This will be called by the `safehttp.ResponseWriter` before a response is written to the underlying `http.ResponseWriter`, for all interceptors that were installed on the handler. If a response  is written in any of the `Commit` phases,  then the `Commit` phases of the remaining interceptor won't execute.

As the `Commit` phase takes a `safehttp.Response` as argument (e.g. for injecting nonces), we also created a `safehttp.TemplateResponse` interface that encapsulates a safe template and the data that is applied to it.

Co-authored-by: Mihali Mara <maramihali@google.com>

### DIFF
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -96,7 +96,8 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
 	}
 }
 
-type appliedInterceptor struct {
+// AppliedInterceptor TODO
+type AppliedInterceptor struct {
 	it  Interceptor
 	cfg Config
 }
@@ -109,7 +110,7 @@ type appliedInterceptor struct {
 // Interceptor was not installed will produce no effect. If multiple Configs are
 // passed for the same Interceptor, only the first one will take effect.
 func (m *ServeMux) Handle(pattern string, method string, h Handler, cfgs ...Config) {
-	var interceps []appliedInterceptor
+	var interceps []AppliedInterceptor
 	for _, it := range m.interceps {
 		var cfg Config
 		for _, c := range cfgs {
@@ -118,7 +119,7 @@ func (m *ServeMux) Handle(pattern string, method string, h Handler, cfgs ...Conf
 				break
 			}
 		}
-		interceps = append(interceps, appliedInterceptor{it: it, cfg: cfg})
+		interceps = append(interceps, AppliedInterceptor{it: it, cfg: cfg})
 	}
 	hi := handlerWithInterceptors{
 		handler:   h,
@@ -183,21 +184,17 @@ func (m methodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // interceptors.
 type handlerWithInterceptors struct {
 	handler   Handler
-	interceps []appliedInterceptor
+	interceps []AppliedInterceptor
 	disp      Dispatcher
 }
 
 // ServeHTTP calls the Before method of all the interceptors and then calls the
-// underlying handler.
+// underlying handler. Any panics that occur during the Before phase, during handler
+// execution or during the Commit phase are recovered here and a 500 Internal Server
+// Error response is sent.
 func (h handlerWithInterceptors) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ir := NewIncomingRequest(r)
-	cp := &CommitPhase{
-		Req:            ir,
-		Interceps:      h.interceps,
-		Configs:        h.configs,
-		IntercepsOrder: h.intercepsOrder,
-	}
-	rw := NewResponseWriter(h.disp, w, h.interceps, cp)
+	rw := NewResponseWriter(h.disp, w, ir, h.interceps)
 
 	// The `net/http` package recovers handler panics, but we cannot rely on that behavior here.
 	// The reason is, we might need to run After/Commit stages of the interceptors before we
@@ -218,26 +215,5 @@ func (h handlerWithInterceptors) ServeHTTP(w http.ResponseWriter, r *http.Reques
 	h.handler.ServeHTTP(rw, ir)
 	if !rw.written {
 		rw.NoContent()
-	}
-}
-
-// CommitPhase contains the configured interceptors that are associated with
-// handler and the incoming request.
-type CommitPhase struct {
-	Req            *IncomingRequest
-	Configs        map[string]Config
-	Interceps      map[string]Interceptor
-	IntercepsOrder []string
-}
-
-func (c *CommitPhase) commit(w *CommitResponseWriter, resp Response) {
-	for i := len(c.IntercepsOrder) - 1; i >= 0; i-- {
-		key := c.IntercepsOrder[i]
-		it := c.Interceps[key]
-		cfg := c.Configs[key]
-		it.Commit(w, c.Req, resp, cfg)
-		if w.rw.written {
-			return
-		}
 	}
 }
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -96,8 +96,7 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
 	}
 }
 
-// AppliedInterceptor TODO
-type AppliedInterceptor struct {
+type appliedInterceptor struct {
 	it  Interceptor
 	cfg Config
 }
@@ -110,7 +109,7 @@ type AppliedInterceptor struct {
 // Interceptor was not installed will produce no effect. If multiple Configs are
 // passed for the same Interceptor, only the first one will take effect.
 func (m *ServeMux) Handle(pattern string, method string, h Handler, cfgs ...Config) {
-	var interceps []AppliedInterceptor
+	var interceps []appliedInterceptor
 	for _, it := range m.interceps {
 		var cfg Config
 		for _, c := range cfgs {
@@ -119,7 +118,7 @@ func (m *ServeMux) Handle(pattern string, method string, h Handler, cfgs ...Conf
 				break
 			}
 		}
-		interceps = append(interceps, AppliedInterceptor{it: it, cfg: cfg})
+		interceps = append(interceps, appliedInterceptor{it: it, cfg: cfg})
 	}
 	hi := handlerWithInterceptors{
 		handler:   h,
@@ -184,7 +183,7 @@ func (m methodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // interceptors.
 type handlerWithInterceptors struct {
 	handler   Handler
-	interceps []AppliedInterceptor
+	interceps []appliedInterceptor
 	disp      Dispatcher
 }
 
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -194,7 +194,8 @@ type handlerWithInterceptors struct {
 // Error response is sent.
 func (h handlerWithInterceptors) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ir := NewIncomingRequest(r)
-	rw := NewResponseWriter(h.disp, w, ir, h.interceps)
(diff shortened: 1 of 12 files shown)


kele:
What's the need for `CommitResponseWriter`? Why can't it be just `ResponseWriter`? If you plan to change something about the `ResponseWriter`, can it be done in a separate PR?

Similarly, can the `DataTemplate` be done in a separate PR?

mihalimara22:
We want the `Commit` phase to only have access to the `WriteError` method in the `ResponseWriter` and the ability to set headers. We considered some alternatives, but they were either very messy or required big changes to the whole design. So, we created the `CommitResponseWriter`, which is basically a wrapper around `ResponseWriter` that serves the purpose above. I think this should be in this PR since it doesn't make much sense without the `Commit` phase (which is its only use case). Similarly, the `DataTemplate` is a `Response` that bundles together `Template` and `data` so it can be passed to the `Commit` phase which might want to modify them (e.g. inject nonces). So it is also strictly related to the `Commit` phase (for now at least). 

They do need documentation, but we wanted to do that together with the documentation for `ResponseWriter`, `IncomingRequest`, etc.

 WDYT? Should they still be split?

kele:
> We want the `Commit` phase to only have access to the `WriteError` method in the `ResponseWriter` and the ability to set headers. (...)

Please introduce the `Commit` phase without any new features first.

> Similarly, the `DataTemplate` is a `Response` that bundles together `Template` and `data` so it can be passed to the `Commit` phase which might want to modify them (e.g. inject nonces). So it is also strictly related to the `Commit` phase (for now at least).

Sure, this seems unavoidable. I didn't think of it, sorry.

> WDYT? Should they still be split?

Yes, please just keep a simple `Commit` phase with support for whatever features we have right now (i.e. that's why `DataTemplate` is okay).

maramihali:
So the alternative would be to have the `ResponseWriter` passed as argument to the `Commit` phase and then in a subsequent PR have the `CommitResponseWriter` (or something with a better name). Would you prefer it that way (just to make sure I understood it correctly)?

kele:
Yup! From my perspective:

- `Commit` phase needs to be added
- `CommitResponseWriter` or whatever other restrictions regarding what `Commit` phase can and cannot do are a separate feature
- `DataTemplate` seems to be required, as adding `Commit` with support only for `Write` and not `ExecuteTemplate` seems like introducing an inconsistency
- I don't understand what `CommitPhase` does

safehttp/mux.go
```diff
@@ -198,8 +198,14 @@ type handlerWithInterceptors struct {
 // ServeHTTP calls the Before method of all the interceptors and then calls the
 // underlying handler.
```
empijei, line 199: Change docs accordingly.

safehttp/mux_test.go
```diff
@@ -324,16 +400,23 @@ func (setHeaderConfig) Match(i safehttp.Interceptor) bool {
...
-	w.Header().Set(name, value)
+	w.Header().Set("Pizza", value)
 	return safehttp.Result{}
 }
 
+func (p setHeaderConfigInterceptor) Commit(w *safehttp.CommitResponseWriter, r *safehttp.IncomingRequest, resp safehttp.Response, cfg interface{}) safehttp.Result {
+	value := "Fusili"
+	if c, ok := cfg.(setHeaderConfig); ok {
+		value = c.pastaValue
+	}
```
empijei, line 415: I would honestly use a type assertion without check. If this assertion fails we have a big bug in the framework and I'd rather have this panic.

empijei (commented):
I second kele comments, and I would add that we will also need a wrapper for JSON, not just templates, but I would suggest to call both JSONResponse and TemplateResponse. DataTemplate is not clear to me as a name.

I also think that the CommitPhase type might be superfluous, can't we do this by adding to types we already have?

safehttp/mux.go
```diff
@@ -183,15 +184,17 @@ func (m methodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	handler   Handler
-	interceps []appliedInterceptor
+	interceps []AppliedInterceptor
 	disp      Dispatcher
 }
 
 // ServeHTTP calls the Before method of all the interceptors and then calls the
-// underlying handler.
+// underlying handler. Any panics that occur during the Before phase, during handler
+// execution or during the Commit phase are recovered here and a 500 Internal Server
```
stapelberg, line 193: Recovering panics is a design decision that the writers of `net/http` regretted.

Wouldn’t it be safer to not make that mistake again?

Even if the answer is no, the comment should be updated to explain.

safehttp/response_writer.go
```diff
@@ -53,21 +57,54 @@ func NotWritten() Result {
 	return Result{}
 }
 
+// commitPhase calls the Commit phases of all the interceptors. This stage will
+// run before a response is written to the ResponseWriter. If a response is
+// written to the ResponseWriter in a Commit phase then the Commit phases of the
+// remaining interceptors won't execute.
+//
+// TODO: BIG WARNING, if ResponseWriter.Write and ResponseWriter.WriteTemplate
+// are called in Commit then this will recurse. CommitResponseWriter was an
```
stapelberg, line 66: What happened to CommitResponseWriter? Don’t leave the reader hanging :)

safehttp/mux.go
maramihali: Finding a better alternative to panics is one of our TODOs so they will disappear in v0 :)

safehttp/response_writer.go
maramihali: @kele suggested to leave this feature (and other possible restrictions to the `Commit` phase that we haven't foreseen) for another PR. :)

safehttp/mux_test.go
```diff
@@ -315,8 +371,8 @@ func TestMuxInterceptors(t *testing.T) {
 }
 
 type setHeaderConfig struct {
-	name  string
-	value string
```
kele, line 319: Please remove all unrelated changes.

safehttp/mux.go
```diff
@@ -96,7 +96,8 @@ func NewServeMux(d Dispatcher, domains ...string) *ServeMux {
 	}
 }
 
-type appliedInterceptor struct {
+// AppliedInterceptor TODO
+type AppliedInterceptor struct {
```
kele, line 100: Why is this part of the public API now?

safehttp/response.go
```diff
@@ -23,3 +23,12 @@ type Response interface{}
 type Template interface {
 	Execute(wr io.Writer, data interface{}) error
 }
+
+// TemplateResponse TODO
```
kele, line 27: Fix the TODOs.

safehttp/response_writer.go
```diff
@@ -25,18 +25,22 @@ type ResponseWriter struct {
 
 	// Having this field unexported is essential for security. Otherwise one can
 	// easily overwrite the struct bypassing all our safety guarantees.
-	header  Header
-	written bool
+	header           Header
+	appliedInterceps []AppliedInterceptor
+	req              *IncomingRequest
```
kele, line 30: Why is this a part of a `ResponseWriter` now?

safehttp/response_writer.go
```diff
@@ -25,18 +25,22 @@ type ResponseWriter struct {
...
+	written          bool
 }
 
 // NewResponseWriter creates a ResponseWriter from a safehttp.Dispatcher, an
 // http.ResponseWriter and a list of interceptors associated with a ServeMux.
-func NewResponseWriter(d Dispatcher, rw http.ResponseWriter) *ResponseWriter {
+func NewResponseWriter(d Dispatcher, rw http.ResponseWriter, req *IncomingRequest, interceps []AppliedInterceptor) *ResponseWriter {
```
kele, line 36: It doesn't seem right that the `ResponseWriter` has direct control over the interceptors. It should ask the mux or the handler wrapper that has the interceptors to do the work.

safehttp/response_writer.go
```diff
@@ -73,13 +110,20 @@ func (w *ResponseWriter) WriteTemplate(t Template, data interface{}) Result {
...
+		return Result{}
+	}
 	w.markWritten()
 	w.rw.WriteHeader(int(StatusNoContent))
 	return Result{}
 }
 
 // WriteError writes an error response (400-599) according to the provided status
-// code.
+// code. Any headers previously set will be removed.
```
kele, line 126: Is this still true?

safehttp/response_writer.go
mattiasgrenfeldt: Well, both `appliedInterceps` and `req` needed to be part of the `ResponseWriter` for our design where the `ResponseWriter` would run the `commitPhase`. In the newer design I'm working on now the `ResponseWriter` still needs to own the `req` but instead of `appliedInterceps` it has a reference to the `handlerWithInterceptors` and the `commitPhase` is moved to the `handlerWithInterceptors`.

safehttp/mux.go
mattiasgrenfeldt: It was part of the public API since we wanted to pass it as part of the `ResponseWriter` constructor and for that it needs to be exported, so that the constructor doesn't become unusable outside the package.

safehttp/mux_test.go
mattiasgrenfeldt: What do you mean with unrelated changes? We have updated the `SetHeaderInterceptor with config` and the `SetHeaderInterceptor with mismatching config` tests to also test the commit phase. Earlier they only tested the Before phase. How is this unrelated from the PR introducing the commit phase?

I looked through all our changes and found one unrelated change in `mux_test.go` in `TestMuxInterceptors`:
```
- wantStatus: 500,
+ wantStatus: safehttp.StatusInternalServerError,
```
We have in an earlier PR replaced all numeric status codes with constants and apparently I had missed this instance. How could it not be ok to fix these kinds of tiny nits when they are discovered? The other option would be to open a new issue and PR to just change this little thing.

safehttp/mux_test.go
mattiasgrenfeldt: Ah, yes, this check was to check if a valid config was passed. If there was no valid config passed then the default values would be used.

safehttp/mux_test.go
mattiasgrenfeldt: Fixed!

safehttp/response_writer.go
```diff
@@ -26,17 +26,22 @@ type ResponseWriter struct {
...
 	written bool
 }
 
 // NewResponseWriter creates a ResponseWriter from a safehttp.Dispatcher, an
-// http.ResponseWriter and a list of interceptors associated with a ServeMux.
-func NewResponseWriter(d Dispatcher, rw http.ResponseWriter) *ResponseWriter {
+// http.ResponseWriter and a reference to the current IncomingRequest being served.
+// The IncomingRequest will only be used by the commit phase, which only runs when
+// the ServeMux is used, and can be passed as nil in tests.
+func NewResponseWriter(d Dispatcher, rw http.ResponseWriter, req *IncomingRequest) *ResponseWriter {
```
kele, line 38: Let's finish this PR, but the `IncomingRequest` should not be added here, please leave a TODO.

kele (approved):

safehttp/response_writer.go
kele, line 38:
```suggestion
// TODO: remove the IncomingRequest parameter.
func NewResponseWriter(d Dispatcher, rw http.ResponseWriter, req *IncomingRequest) *ResponseWriter {
```

Issue #133, opened 2020-08-31
Create a default safe Dispatcher


maramihali:
In order to build a web server that uses the framework we need to provide a Dispatcher that supports safehtml, sets the Content-Type properly, has JSON support, etc.

PR #147, opened 2020-09-07
Extended the Dispatcher interface to support writing JSON and retrieving the response Content-Type and created a safe default Dispatcher.

### LINKED ISSUE #60: The dispatcher can write a response without providing a Content-Type header.

maramihali:
Fixes #60, #133

Extended the Dispatcher interface to support writing JSON safe responses and retrieving the Content-Type of safe response and added a . This way the ResponseWriter can query the Dispatcher to retrieve and set the Content-Type of a safe response before writing the response. Implemented and write tests for a default safe dispatcher.

### DIFF
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -56,9 +56,9 @@ func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
 // If the provided response is not valid JSON or writing the response fails, the
 // method will return an error.
 func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) error {
-	// TODO: XSSI protection needs to be added as this is not a safe response
-	// type yet
-	return json.NewEncoder(rw).Encode(resp.Data)
+	e := json.NewEncoder(rw)
+	e.Encode(")]}',")
+	return e.Encode(resp.Data)
 }
 
 // ExecuteTemplate applies a parsed template to the provided data object if the
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -56,9 +56,8 @@ func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
 // If the provided response is not valid JSON or writing the response fails, the
 // method will return an error.
 func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) error {
-	e := json.NewEncoder(rw)
-	e.Encode(")]}',")
-	return e.Encode(resp.Data)
+	io.WriteString(rw, ")]}',\n")
+	return json.NewEncoder(rw).Encode(resp.Data)
 }
 
 // ExecuteTemplate applies a parsed template to the provided data object if the
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -16,7 +16,7 @@ package safehttp
 
 import (
 	"encoding/json"
-	"errors"
+	"fmt"
 	"github.com/google/safehtml"
 	"github.com/google/safehtml/template"
 	"net/http"
@@ -26,8 +26,8 @@ import (
 // Response Writer if it's of a safe type.
 type DefaultDispatcher struct{}
 
-// ContentType returns the Content-Type of a safe respons if it's of a safe-type
-// and a non-nil error otherwise.
+// ContentType returns the Content-Type of a safe response if it's of a
+// safe-type and an error otherwise.
 func (DefaultDispatcher) ContentType(resp Response) (string, error) {
 	switch resp.(type) {
 	case safehtml.HTML, *template.Template:
@@ -35,7 +35,7 @@ func (DefaultDispatcher) ContentType(resp Response) (string, error) {
 	case JSONResponse:
 		return "application/json; charset=utf-8", nil
 	default:
-		return "", errors.New("not a safe response")
+		return "", fmt.Errorf("%T is not a safe response type, a Content-Type cannot be provided", resp)
 	}
 }
 
@@ -45,7 +45,7 @@ func (DefaultDispatcher) ContentType(resp Response) (string, error) {
 func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
 	x, ok := resp.(safehtml.HTML)
 	if !ok {
-		return errors.New("not a safe response type")
+		return fmt.Errorf("%T is not a safe response type and it cannot be written", resp)
 	}
 	_, err := rw.Write([]byte(x.String()))
 	return err
@@ -56,12 +56,9 @@ func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
 // provided response is not valid JSON or writing the response fails, the method
 // will return an error.
 func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) error {
-	obj, err := json.Marshal(resp.Data)
-	if err != nil {
-		return errors.New("invalid json")
-	}
-	_, err = rw.Write(obj)
-	return err
+	// TODO: XSSI protection needs to be added as this is not a safe response
+	// type yet
+	return json.NewEncoder(rw).Encode(resp.Data)
 }
 
 // ExecuteTemplate applies a parsed template to the provided data object if the
@@ -73,8 +70,7 @@ func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) er
 func (DefaultDispatcher) ExecuteTemplate(rw http.ResponseWriter, t Template, data interface{}) error {
 	x, ok := t.(*template.Template)
 	if !ok {
-		return errors.New("not a safe response type")
+		return fmt.Errorf("%T is not a safe template and it cannot be parsed and written", t)
 	}
-	err := x.Execute(rw, data)
-	return err
+	return x.Execute(rw, data)
 }
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -56,7 +56,7 @@ func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
 // If the provided response is not valid JSON or writing the response fails, the
 // method will return an error.
 func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) error {
-	io.WriteString(rw, ")]}',\n")
+	io.WriteString(rw, ")]}',\n") // Break parsing of JavaScript in order to prevent XSSI.
 	return json.NewEncoder(rw).Encode(resp.Data)
 }
 
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -1,0 +1,80 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package safehttp
+
+import (
+	"encoding/json"
+	"errors"
+	"github.com/google/safehtml"
+	"github.com/google/safehtml/template"
+	"net/http"
+)
+
+// DefaultDispatcher is responsible of writing the body of a response to an
+// Response Writer if it's of a safe type.
+type DefaultDispatcher struct{}
+
+// ContentType returns the Content-Type of a safe respons if it's of a safe-type
+// and a non-nil error otherwise.
+func (DefaultDispatcher) ContentType(resp Response) (string, error) {
+	switch resp.(type) {
+	case safehtml.HTML, *template.Template:
+		return "text/html; charset=utf-8", nil
+	case JSONResponse:
+		return "application/json; charset=utf-8", nil
+	default:
+		return "", errors.New("not a safe response")
+	}
+}
+
+// Write writes the response to the Response Writer if it's safe HTML. It
+// returns a non-nil error if the response is not safe HTML or if writing the
+// response fails.
+func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
+	x, ok := resp.(safehtml.HTML)
+	if !ok {
+		return errors.New("not a safe response type")
+	}
+	_, err := rw.Write([]byte(x.String()))
+	return err
+
+}
+
+// WriteJSON serialises and writes a JSON to the Response Writer. If the
+// provided response is not valid JSON or writing the response fails, the method
+// will return an error.
+func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) error {
+	obj, err := json.Marshal(resp.Data)
+	if err != nil {
+		return errors.New("invalid json")
+	}
+	_, err = rw.Write(obj)
+	return err
+}
+
+// ExecuteTemplate applies a parsed template to the provided data object if the
+// template is a safe HTML template, writing the output to the ResponseWriter.
+//
+// If an error occurs executing the template or writing its output,
+// execution stops, but partial results may already have been written to
+// the Response Writer.
+func (DefaultDispatcher) ExecuteTemplate(rw http.ResponseWriter, t Template, data interface{}) error {
+	x, ok := t.(*template.Template)
+	if !ok {
+		return errors.New("not a safe response type")
+	}
+	err := x.Execute(rw, data)
+	return err
+}
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -19,11 +19,11 @@ import (
 	"fmt"
 	"github.com/google/safehtml"
 	"github.com/google/safehtml/template"
+	"io"
 	"net/http"
 )
 
-// DefaultDispatcher is responsible of writing the body of a response to an
-// Response Writer if it's of a safe type.
+// DefaultDispatcher is responsible for writing safe responses.
 type DefaultDispatcher struct{}
 
 // ContentType returns the Content-Type of a safe response if it's of a
@@ -47,14 +47,14 @@ func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
 	if !ok {
 		return fmt.Errorf("%T is not a safe response type and it cannot be written", resp)
 	}
-	_, err := rw.Write([]byte(x.String()))
+	_, err := io.WriteString(rw, x.String())
 	return err
 
 }
 
-// WriteJSON serialises and writes a JSON to the Response Writer. If the
-// provided response is not valid JSON or writing the response fails, the method
-// will return an error.
+// WriteJSON serialises and writes a JSON to the underlying http.ResponseWriter.
+// If the provided response is not valid JSON or writing the response fails, the
+// method will return an error.
 func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) error {
 	// TODO: XSSI protection needs to be added as this is not a safe response
 	// type yet
@@ -62,7 +62,8 @@ func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) er
 }
 
 // ExecuteTemplate applies a parsed template to the provided data object if the
-// template is a safe HTML template, writing the output to the ResponseWriter.
+// template is a safe HTML template, writing the output to the  http.
+// ResponseWriter.
 //
 // If an error occurs executing the template or writing its output,
 // execution stops, but partial results may already have been written to
--- a/safehttp/default_dispatcher_test.go
+++ b/safehttp/default_dispatcher_test.go
@@ -58,7 +58,7 @@ func TestDefaultDispatcherValidResponse(t *testing.T) {
 				}{Field: "myField"}
 				return d.WriteJSON(w, safehttp.JSONResponse{Data: data})
 			},
-			wantBody: "{\"field\":\"myField\"}\n",
+			wantBody: ")]}',\n{\"field\":\"myField\"}\n",
 		},
 	}
 	for _, tt := range tests {
--- a/safehttp/default_dispatcher_test.go
+++ b/safehttp/default_dispatcher_test.go
@@ -83,6 +83,7 @@ func TestDefaultDispatcherInvalidResponse(t *testing.T) {
 	tests := []struct {
 		name  string
 		write func(w http.ResponseWriter) error
+		want  string
 	}{
 		{
 			name: "Unsafe HTML Response",
@@ -90,6 +91,7 @@ func TestDefaultDispatcherInvalidResponse(t *testing.T) {
 				d := &safehttp.DefaultDispatcher{}
 				return d.Write(w, "<h1>Hello World!</h1>")
 			},
+			want: "",
 		},
 		{
 			name: "Unsafe Template Response",
@@ -97,6 +99,7 @@ func TestDefaultDispatcherInvalidResponse(t *testing.T) {
 				d := &safehttp.DefaultDispatcher{}
 				return d.ExecuteTemplate(w, template.Must(template.New("name").Parse("<h1>{{ . }}</h1>")), "This is an actual heading, though.")
 			},
+			want: "",
 		},
 		{
 			name: "Invalid JSON Response",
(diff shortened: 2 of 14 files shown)


safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/google/safehtml"
+	"github.com/google/safehtml/template"
+	"net/http"
+)
+
+// DefaultDispatcher is responsible of writing the body of a response to an
+// Response Writer if it's of a safe type.
+type DefaultDispatcher struct{}
+
+// ContentType returns the Content-Type of a safe respons if it's of a safe-type
```
mattiasgrenfeldt, line 29: `response` with an `e`

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,80 @@
...
+// and a non-nil error otherwise.
```
mattiasgrenfeldt, line 30: `and an error otherwise`

safehttp/mux_test.go
```diff
@@ -482,3 +500,12 @@ func TestMuxHandlerReturnsNotWritten(t *testing.T) {
...
 	}
 }
+
+func TestMuxNilDispatcher(t *testing.T) {
+	defer func() {
+		if r := recover(); r == nil {
+			t.Errorf(`mux.NewServeMux(nil): expected panic`)
+		}
+	}()
+	_ = safehttp.NewServeMux(nil)
```
mattiasgrenfeldt, line 510: You can remove the `_ =`

safehttp/response_writer.go
```diff
@@ -64,21 +66,68 @@ func NotWritten() Result {
...
+// TODO: replace panics with proper error handling when getting the response
+// Content-Type or writing the response fails.
 func (w *ResponseWriter) Write(resp Response) Result {
-	w.markWritten()
+	ct, err := w.d.ContentType(resp)
+	if err != nil {
+		panic(err)
+	}
+	w.rw.WriteHeader(int(StatusOK))
+	w.rw.Header().Set("Content-Type", ct)
```
mattiasgrenfeldt, line 82: You need to set the ct header before calling `WriteHeader`. The headers of the response are written when you call `WriteHeader`, so they must be modified first.

The underlying responsewriter will write a 200 OK response if `WriteHeader` was never called, so you can remove the call to `WriteHeader`. And if there is a error during `w.d.Write` which results in a panic then something else will try to write to the responsewriter and in that case it would be bad if the response code is already set.

So remove the call to `WriteHeader` here, in `WriteJSON` and in `WriteTemplate`.

safehttp/response_writer_test.go
```diff
@@ -66,6 +67,18 @@ func TestResponseWriterWriteTwicePanic(t *testing.T) {
...
+			name: "Call WriteJSON twice",
+			write: func(w *safehttp.ResponseWriter) {
+				data := struct{ Field string }{Field: "myField"}
+				jsonObj, err := json.Marshal(data)
+				if err != nil {
+					t.Fatalf("invalid json object")
+				}
+				w.WriteJSON(jsonObj)
+				w.WriteJSON(jsonObj)
+			},
```
mattiasgrenfeldt, line 80: When you use `WriteJSON` you send the actual struct, right? The dispatcher will marshal it for you so that isn't needed here. Now you are sending just a string.

safehttp/response_writer.go
```diff
@@ -52,7 +52,9 @@ func (w *ResponseWriter) Interceptor(key string) Interceptor {
 }
 
 // Result TODO
-type Result struct{}
+type Result struct {
+	err error
+}
```
mattiasgrenfeldt, line 57: Is this field used anywhere? If it isn't, remove it for now.

safehttp/default_dispatcher_test.go
```diff
@@ -0,0 +1,124 @@
...
+		write func(w http.ResponseWriter) error
+	}{
+		{
+			name: "Unsafe HTML Response",
+			write: func(w http.ResponseWriter) error {
+				d := &safehttp.DefaultDispatcher{}
+				return d.Write(w, "<h1>Hello World!</h1>")
+			},
+		},
+
```
mattiasgrenfeldt, line 94: Remove empty line

mattiasgrenfeldt (changes requested):

safehttp/default_dispatcher.go
kele: FYI: @mattiasgrenfeldt check out the "add suggestion" option when making comments (the +- sign in the comment toolbar). This allows the PR author to apply the suggestions from GitHub's web UI with one click.

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,80 @@
...
+	_, err := rw.Write([]byte(x.String()))
+	return err
+
+}
+
+// WriteJSON serialises and writes a JSON to the Response Writer. If the
+// provided response is not valid JSON or writing the response fails, the method
+// will return an error.
+func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) error {
+	obj, err := json.Marshal(resp.Data)
```
kele, line 59: Add a `TODO` explaining that this is **not** safe yet, and XSSI protection needs to be added (where?).

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,80 @@
...
+// ContentType returns the Content-Type of a safe respons if it's of a safe-type
+// and a non-nil error otherwise.
+func (DefaultDispatcher) ContentType(resp Response) (string, error) {
+	switch resp.(type) {
+	case safehtml.HTML, *template.Template:
+		return "text/html; charset=utf-8", nil
+	case JSONResponse:
+		return "application/json; charset=utf-8", nil
+	default:
+		return "", errors.New("not a safe response")
```
kele, line 38: More descriptive errors please (e.g. include type).

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+// Write writes the response to the Response Writer if it's safe HTML. It
+// returns a non-nil error if the response is not safe HTML or if writing the
+// response fails.
+func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
+	x, ok := resp.(safehtml.HTML)
+	if !ok {
+		return errors.New("not a safe response type")
```
kele, line 48: More descriptive errors please (e.g. include type).

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,80 @@
...
+
+}
+
+// WriteJSON serialises and writes a JSON to the Response Writer. If the
+// provided response is not valid JSON or writing the response fails, the method
+// will return an error.
+func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) error {
+	obj, err := json.Marshal(resp.Data)
+	if err != nil {
+		return errors.New("invalid json")
```
kele, line 61: More descriptive errors please (e.g. include `err`, or simply return it).

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	_, err = rw.Write(obj)
+	return err
```
kele, line 64: Use `json.Encoder` instead, i.e.: `json.NewEncoder(rw).Encode(obj)`. Now the marshalling to an object is unnecessary, as it can be done on the fly by the `Encoder`.

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,80 @@
...
+// ExecuteTemplate applies a parsed template to the provided data object if the
+// template is a safe HTML template, writing the output to the ResponseWriter.
+//
+// If an error occurs executing the template or writing its output,
+// execution stops, but partial results may already have been written to
+// the Response Writer.
+func (DefaultDispatcher) ExecuteTemplate(rw http.ResponseWriter, t Template, data interface{}) error {
+	x, ok := t.(*template.Template)
+	if !ok {
+		return errors.New("not a safe response type")
```
kele, line 76: More descriptive errors please (e.g. include type).

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	err := x.Execute(rw, data)
+	return err
```
kele, line 79:
```suggestion
	return x.Execute(rw, data)
```

safehttp/response_writer.go
kele: 1. I agree that the headers need to be set before doing `WriteHeader`.
2. I'd still explicitly call `WriteHeader(200)` here, though.

safehttp/response_writer.go
```diff
@@ -64,21 +66,68 @@ func NotWritten() Result {
...
+	if err != nil {
+		panic(err)
+	}
+	w.rw.WriteHeader(int(StatusOK))
+	w.rw.Header().Set("Content-Type", ct)
 	if err := w.d.Write(w.rw, resp); err != nil {
-		panic("error")
+		panic(err)
 	}
+	w.markWritten()
```
kele, line 86: Why this change?

tests/integration/header/header_test.go
```diff
@@ -29,6 +32,17 @@ import (
...
+	switch resp.(type) {
+	case safehtml.HTML, *template.Template:
+		return "text/html; charset=utf-8", nil
+	case safehttp.JSONResponse:
+		return "application/json; charset=utf-8", nil
+	default:
+		return "", errors.New("not a safe response")
+	}
+}
+
```
kele, line 45: Please file an issue to carve out the `testDispatcher`, as it feels it should be the same for most of the tests.

kele (changes requested):

maramihali:
I realised I forgot to add the integration tests that test a `ServeMux` created with a `DefaultDispatcher`, I amended the last commit. Sorry about this!

safehttp/response_writer.go
maramihali: We decided in the simplifying session to only call `markWritten` after a response has been written and this change made its way in this PR as I was refactoring things.

safehttp/response_writer.go
mattiasgrenfeldt: Imagine that `w.d.Write` panics or returns an error, in which case `ResponseWriter.Write` will panic. This panic will be recovered by the framework (ServeMux) and a 500 Internal Server Error response will be written to the `ResponseWriter`. But the status code and the headers have already been written, so things will break.

We need to remove the call to `WriteHeader`.

maramihali: Right, we discussed this offline yesterday. We need a better way of handling errors instead of the panics. Right now, if the Dispatcher errs after writing half of the response, the status code and content type had already been set and there's not much we can do as we cannot reset it, right? An option is to buffer the response and only send it after we make sure everything works as intended, but IIRC that's a feature that we don't plan to introduce in V0.

Also, BTW, these are problems that don't come up in tests because we use the `responseRecorder`. This is not ideal IMO as it's easy to introduce bugs (e.g. my bad ordering of statements) and tests won't fail. Should we do something about it?

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Write writes the response to the Response Writer if it's safe HTML. It
+// returns a non-nil error if the response is not safe HTML or if writing the
+// response fails.
+func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
+	x, ok := resp.(safehtml.HTML)
+	if !ok {
+		return fmt.Errorf("%T is not a safe response type and it cannot be written", resp)
+	}
+	_, err := rw.Write([]byte(x.String()))
```
kele, line 50: Use `io.WriteString`. If `http.ResponseWriter` implements the `StringWriter` interface, this might save allocating a byte slice (and it seems it [does](https://golang.org/src/net/http/serve_test.go#L4015)).

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/google/safehtml"
+	"github.com/google/safehtml/template"
+	"net/http"
+)
+
+// DefaultDispatcher is responsible of writing the body of a response to an
+// Response Writer if it's of a safe type.
```
kele, line 26:
```suggestion
// DefaultDispatcher is responsible for writing safe responses.
```

safehttp/response_writer.go
kele: Please leave out this change for a separate PR.

safehttp/response_writer.go
```diff
@@ -122,5 +169,7 @@ func (w *ResponseWriter) SetCookie(c *Cookie) error {
 // Dispatcher TODO
 type Dispatcher interface {
 	Write(rw http.ResponseWriter, resp Response) error
+	WriteJSON(rw http.ResponseWriter, resp JSONResponse) error
 	ExecuteTemplate(rw http.ResponseWriter, t Template, data interface{}) error
+	ContentType(resp Response) (string, error)
```
kele, line 174: Start adding comments to the methods. You don't have to add to the ones that existed before your PR, we can do that later.

safehttp/response_writer_test.go
```diff
@@ -66,6 +66,14 @@ func TestResponseWriterWriteTwicePanic(t *testing.T) {
 				w.Write(safehtml.HTMLEscaped("<h1>Escaped, so not really a heading</h1>"))
 			},
 		},
+		{
+			name: "Call WriteJSON twice",
+			write: func(w *safehttp.ResponseWriter) {
+				obj := struct{ Field string }{Field: "myField"}
+				w.WriteJSON(obj)
+				w.WriteJSON(obj)
+			},
```
kele, line 75: Add a tests that tries to `w.Write` and then `w.WriteJSON`.

safehttp/safehttptest/recorder.go
```diff
@@ -26,6 +28,17 @@ import (
 
 type testDispatcher struct{}
```
kele, line 29: Does this mean you could use the `DefaultDispatcher` here now? Possibly in other tests too?

This can be done in a follow-up PR.

kele (changes requested):

safehttp/safehttptest/recorder.go
maramihali: Yep, I think so, and it would definitely make tests cleaner. We could use the `DefaultDispatcher` in tests by default (too many defaults in one sentence) and only create a `testDispatcher` for analysing some custom behaviour. 

But yeah, should be done in a clean-up PR and not in this one, I think.

safehttp/response_writer.go
maramihali: The big problem with this is that it introduces a bug, making the tests  panic. So, if we put `markWritten` at the beginning that will set `w.written` to `true`. Then if either `d.ContentType(...)` or `d.WriteSomething(...)` returns an error, we will panic and that will be recovered by the `ServeMux`. The `ServeMux` will then attempt to call`WriteError` with a 500, which attempts to call `markWritten` again, which will panic because `w.written` was already set to true which ends up in a recursive thing, I think :/

maramihali: Therefore, I'd suggest leaving it as it is in this PR and then fixing it properly in a follow-up PR (the alternative is `t.Skip()` the tests in `integration/mux/mux_test.go` that test the behaviour of passing unsafe responses until we fix it properly)

safehttp/response_writer.go
mattiasgrenfeldt: I am not sure what to do here.

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,77 @@
...
+	return err
+
+}
+
+// WriteJSON serialises and writes a JSON to the underlying http.ResponseWriter.
+// If the provided response is not valid JSON or writing the response fails, the
+// method will return an error.
+func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) error {
+	// TODO: XSSI protection needs to be added as this is not a safe response
+	// type yet
```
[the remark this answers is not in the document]
maramihali: Oh, how is that preventing XSSI?

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,77 @@
...
+	"io"
+	"net/http"
+)
+
+// DefaultDispatcher is responsible for writing safe responses.
+type DefaultDispatcher struct{}
+
+// ContentType returns the Content-Type of a safe response if it's of a
+// safe-type and an error otherwise.
+func (DefaultDispatcher) ContentType(resp Response) (string, error) {
```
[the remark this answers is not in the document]
maramihali: Wasn't sure whether that is okay, will add :)

safehttp/response_writer.go
```diff
@@ -53,20 +53,67 @@ func NotWritten() Result {
 	return Result{}
 }
 
-// Write TODO
+// Write dispatches the response to the Dispatcher, setting the Content-Type to
+// text/html; charset=utf-8 and response status to 200 OK if resp is a safe
```
mattiasgrenfeldt, line 57: This sets the `Content-Type` to whatever the dispatcher tells it to. Not necessarily `text/html`

safehttp/response_writer.go
```diff
@@ -53,20 +53,67 @@ func NotWritten() Result {
...
+// TODO: replace panics with proper error handling when getting the response
+// Content-Type or writing the response fails.
+func (w *ResponseWriter) WriteJSON(data interface{}) Result {
+	w.markWritten()
+	resp := JSONResponse{Data: data}
+	ct, err := w.d.ContentType(resp)
+	if err != nil {
+		panic(err)
+	}
+	w.rw.Header().Set("Content-Type", ct)
```
mattiasgrenfeldt, line 91: Shouldn't this method always set the Content-Type to `application/json`? Then it doesn't need to interrogate the Dispatcher

safehttp/response_writer.go
```diff
@@ -53,20 +53,67 @@ func NotWritten() Result {
...
+}
+
+// WriteTemplate dispatches a parsed template and a data object to the
+// Dispatcher to be executed and written to the underlying Response Writer, in
+// case the template is a safe HTML template. If t is a safe HTML Template, the
+// Content-Type will also be set to text/html; charset=utf-8 and the response
+// status to 200 OK.
+//
+// TODO: replace panics with proper error handling when getting the response
+// Content-Type or writing the response fails.
```
mattiasgrenfeldt, line 106: The `Content-Type` for safe html templates depends on the Dispatcher implementation and should be left out of this comment.

mattiasgrenfeldt (changes requested):

safehttp/response_writer.go
mattiasgrenfeldt: I agree that we need a better error strategy for the dispatcher, but as you said, that's not V0.

To detect these kinds of errors in tests we need thorough integration tests between the ServeMux, the ResponseWriter and a Dispatcher.

But I still think that the ResponseWriter should not call `WriteHeader`. The Dispatcher should start doing the actual writing to the network connection.

safehttp/response_writer_test.go
mattiasgrenfeldt: I opened an issue for testing every combination and ordering of writing methods. #157

safehttp/response_writer.go
```diff
@@ -119,8 +166,39 @@ func (w *ResponseWriter) SetCookie(c *Cookie) error {
...
+	// and writes the output to the http.ResponseWriter.
+	//
+	// This should return an error if the provided Template response is not a
+	// valid template or if an error occurs executing or writing the template.
 	ExecuteTemplate(rw http.ResponseWriter, t Template, data interface{}) error
+
+	// Content-Type returns the Content-Type of the provided response if it is
+	// of a type supported by the Dispatcher and should return an error
+	// otherwise.
+
```
kele, line 197:
```suggestion
	//
```

safehttp/response_writer.go
```diff
@@ -119,8 +166,39 @@ func (w *ResponseWriter) SetCookie(c *Cookie) error {
...
+	// Sending a response to the http.ResponseWriter without properly setting
+	// the is bug-prone and could introduce vulnerabilities. Therefore, this
+	// method should be used to set the Content-Type header before calling any
```
kele, line 200:
```suggestion
	// Sending a response to the http.ResponseWriter without properly setting
	// CT is error-prone and could introduce vulnerabilities. Therefore, this
	// method should be used to set the Content-Type header before calling any
```

safehttp/response_writer.go
```diff
@@ -119,8 +166,39 @@ func (w *ResponseWriter) SetCookie(c *Cookie) error {
...
+	// of the writing methods in the Dispatcher. Writing should not proceed
+	// ContentType returns an error.
```
kele, line 202:
```suggestion
	// if ContentType returns an error.
```

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,76 @@
...
+	switch resp.(type) {
+	case safehtml.HTML, *template.Template:
+		return "text/html; charset=utf-8", nil
+	case JSONResponse:
+		return "application/json; charset=utf-8", nil
+	default:
+		return "", fmt.Errorf("%T is not a safe response type, a Content-Type cannot be provided", resp)
```
kele, line 38:
```suggestion
		return "", fmt.Errorf("unknown Content-Type for %T", resp)
```

safehttp/default_dispatcher.go
```diff
@@ -0,0 +1,76 @@
...
+	_, err := io.WriteString(rw, x.String())
+	return err
+
+}
+
+// WriteJSON serialises and writes a JSON to the underlying http.ResponseWriter.
+// If the provided response is not valid JSON or writing the response fails, the
+// method will return an error.
+func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) error {
+	io.WriteString(rw, ")]}',\n")
```
kele, line 59:
```suggestion
	io.WriteString(rw, ")]}',\n") // Break parsing of JavaScript in order to prevent XSSI.
```

safehttp/response_writer.go
kele: I'd keep it for consistency.

safehttp/response_writer.go
```diff
@@ -119,8 +166,39 @@ func (w *ResponseWriter) SetCookie(c *Cookie) error {
 	return w.header.addCookie(c)
 }
 
-// Dispatcher TODO
+// Dispatcher is responsible for writing a response received from the
+// ResponseWriter to the underlying http.ResponseWriter.
+//
+// The implementation of a custom Dispatcher should be thoroughly reviewed by
+// the security team to avoid introducing vulnerabilities. Moreover/
```
kele, line 173: This comment ends abruptly.

safehttp/response_writer.go
```diff
@@ -119,8 +166,39 @@ func (w *ResponseWriter) SetCookie(c *Cookie) error {
...
 	Write(rw http.ResponseWriter, resp Response) error
+
+	// WriteJSON serialises and writes a JSON object to the http.ResponseWriter.
+	//
+	// This should return an error if the provided response is not a valid
+	// JSONResponse  or if writing the object to the http.ResponseWriter fails.
+	WriteJSON(rw http.ResponseWriter, resp JSONResponse) error
+
+	// ExecuteTemplate applies a parsed template to the provided data object
+	// and writes the output to the http.ResponseWriter.
```
kele, line 188:
```suggestion
	// ExecuteTemplate applies a template to the provided data object
	// and writes the output to the http.ResponseWriter.
```

safehttp/response_writer.go
```diff
@@ -119,8 +166,39 @@ func (w *ResponseWriter) SetCookie(c *Cookie) error {
...
+// The implementation of a custom Dispatcher should be thoroughly reviewed by
+// the security team to avoid introducing vulnerabilities. Moreover/
 type Dispatcher interface {
+	// Write writes a Response to the http.ResponseWriter.
+	//
+	// It should return an error if the writing operation fails or if the
+	// provided Response should not be written to the http.ResponseWriter
 	Write(rw http.ResponseWriter, resp Response) error
+
+	// WriteJSON serialises and writes a JSON object to the http.ResponseWriter.
```
kele, line 181:
```suggestion
	// WriteJSON writes a given object as JSON to the http.ResponseWriter.
```

kele (approved):

kele:
@googlebot I consent.

kele:
@empijei I think you need to do the `@googlebot I consent` thing

safehttp/response_writer.go
maramihali: Opened #164.

safehttp/safehttptest/recorder.go
maramihali: Opened #149

safehttp/response_writer.go
maramihali: I am not sure I understand the last sentence.

empijei:
@googlebot I consent.

Issue #132, opened 2020-08-31
Cleanup TODO godocs in the `safehttp` package


kele:
https://pkg.go.dev/github.com/google/go-safeweb/safehttp?tab=doc

kele:
Please make sure this is wrapped up. :)

PR #160, opened 2020-09-10
Added package level godocs to Collector, CORS, CSP, HSTS, staticheaders and safehttptest.

mattiasgrenfeldt:
As part of improving the documentation of the project ( #132 ) I've added package level godocs for the following packages:

- safehttp/plugins/collector
- safehttp/plugins/cors
- safehttp/plugins/csp
- safehttp/plugins/hsts
- safehttp/plugins/staticheaders
- safehttp/safehttptest

Package level comments left to write are for:

- safehttp/plugins/xsrf
- safehttp/plugins/fetch_metadata

I'll let @maramihali write those since she wrote those packages.

### DIFF
--- a/safehttp/plugins/cors/cors.go
+++ b/safehttp/plugins/cors/cors.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package cors provides a safehttp.Interceptor that handles CORS requests.
 package cors
 
 import (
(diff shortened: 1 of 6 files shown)


safehttp/plugins/cors/cors.go
```diff
@@ -42,12 +43,13 @@ var disallowedContentTypes = map[string]bool{
...
 // Each CORS request must contain the header "X-Cors: 1".
 //
+// The HEAD request method is disallowed.
+//
 // All of this is to prevent XSRF.
 type Interceptor struct {
 	// AllowedOrigins determines which origins should be allowed in the
 	// Access-Control-Allow-Origin header.
 	AllowedOrigins map[string]bool
-	allowedHeaders map[string]bool
```
mattiasgrenfeldt, line 50: This move is to make the godoc look better. If this is placed here then there will be an empty line in the godoc between AllowedOrigins and ExposedHeaders. Like this:
```go
// AllowedOrigins determines which origins should be allowed in the
// Access-Control-Allow-Origin header.
AllowedOrigins map[string]bool

// ExposedHeaders ...
```
But if this is moved to be below all of the exported fields then the empty line is removed.

kele (approved):

PR #161, opened 2020-09-11
Rewrite email in commits from personal email to google email.

mattiasgrenfeldt:
I discovered that I had accidentally used my personal email for most of the commits during this project. These commits should instead be made using my @google.com email. I have rewritten history so that the @google.com email is used everywhere.


mattiasgrenfeldt:
I realized that this can't be done using a PR but must be done by force pushing to master.

PR #166, opened 2020-09-14
Address linter comments in the safehttp package

maramihali:
The following changes have been done to address linters comments:
* Added missing documentation
* Replaced usage of `html/template` 
* ensure a `nil` context is not passed to a function directly

TODOS:
* Add godocs to `safesql/sqlwrap.go`.
* Add package level documentation in `safehttp/doc.go` to remove the TODO.

### DIFF
--- a/safehttp/incoming_request_test.go
+++ b/safehttp/incoming_request_test.go
@@ -193,6 +193,8 @@ func TestRequestSetNilContext(t *testing.T) {
 			t.Errorf(`ir.SetContext(nil): expected panic`)
 		}
 	}()
+
+	// Avoids linters complaint about a nil context being passed as argument
 	var nilContext context.Context
 	ir.SetContext(nilContext)
 }
--- a/safehttp/incoming_request_test.go
+++ b/safehttp/incoming_request_test.go
@@ -26,6 +26,8 @@ import (
 	"github.com/google/go-safeweb/safehttp/safehttptest"
 )
 
+var nilContext context.Context
+
 func TestIncomingRequestCookie(t *testing.T) {
 	var tests = []struct {
 		name      string
@@ -194,7 +196,7 @@ func TestRequestSetNilContext(t *testing.T) {
 		}
 	}()
 
-	ir.SetContext(nil)
+	ir.SetContext(nilContext)
 }
 
 func TestIncomingRequestPostForm(t *testing.T) {
--- a/safehttp/incoming_request_test.go
+++ b/safehttp/incoming_request_test.go
@@ -194,7 +194,8 @@ func TestRequestSetNilContext(t *testing.T) {
 		}
 	}()
 
-	// Avoids linters complaint about a nil context being passed as argument
+	// Avoids a linter complaint about a nil context being passed as argument.
+	// In this case, we explicitly want to test that a nil context results in an error.
 	var nilContext context.Context
 	ir.SetContext(nilContext)
 }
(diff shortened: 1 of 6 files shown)


safehttp/incoming_request_test.go
```diff
@@ -26,6 +26,8 @@ import (
 	"github.com/google/go-safeweb/safehttp/safehttptest"
 )
 
+var nilContext context.Context
```
kele, line 29: My understanding is you did that for testing only. Just create a local variable and add a comment why you're doing this.

safehttp/plugins/htmlinject/htmlinject_test.go
```diff
@@ -15,10 +15,10 @@
 package htmlinject
 
 import (
-	"html/template"
 	"os"
 	"strings"
 	"testing"
+	"text/template"
```
kele, line 21: Why?

safehttp/safehttptest/recorder.go
```diff
@@ -17,7 +17,7 @@ package safehttptest
 import (
 	"encoding/json"
 	"errors"
-	"html/template"
+	"github.com/google/safehtml/template"
```
kele, line 20: Why is this usage replaced? Was there a linter complaining about `html/template`?

kele (changes requested):

safehttp/incoming_request_test.go
maramihali: I want to ensure the function `SetContext` panics when a nil context is passed but the linter complains if a nil context is passed directly to the function. This is a hack that will stop the linter from complaining hopefully.

safehttp/plugins/htmlinject/htmlinject_test.go
maramihali: Linters complain when they spot usage of `html/template`

safehttp/safehttptest/recorder.go
maramihali: Yep.

safehttp/incoming_request_test.go
kele: I get that. Could you make it a *local* variable and add a comment in the code explaining what's happening?

safehttp/plugins/htmlinject/htmlinject_test.go
kele: Please keep it. Linters are here to help, but if they don't, we ignore them. Changing this to `text/template` makes it confusing, as this is a `htmlinject_test` file that's supposed to work with HTML.

safehttp/incoming_request_test.go
```diff
@@ -193,8 +193,8 @@ func TestRequestSetNilContext(t *testing.T) {
 			t.Errorf(`ir.SetContext(nil): expected panic`)
 		}
 	}()
-
-	ir.SetContext(nil)
+	var nilContext context.Context
+	ir.SetContext(nilContext)
```
mattiasgrenfeldt, line 197: Add a comment explaining why this is done.

mattiasgrenfeldt (approved):

safehttp/incoming_request_test.go
```diff
@@ -194,7 +194,9 @@ func TestRequestSetNilContext(t *testing.T) {
 		}
 	}()
 
-	ir.SetContext(nil)
+	// Avoids linters complaint about a nil context being passed as argument
```
kele, line 197:
```suggestion
	// Avoids a linter complaint about a nil context being passed as argument.
	// In this case, we explicitly want to test that a nil context results in an error.
```

kele (approved):

PR #167, opened 2020-09-14
Cleanup the repository to use DefaultDispatcher in tests and a singe implementation of http.ResponseWriter

### LINKED ISSUE #149: Use a single Dispatcher and http.ResponseWriter implementation in testing
Currently, the repository has many test files that declare a `testDispatcher`. This could be replaced with the `safehttp.DefaultDispatcher`. Moreover, we could have only one `responseRecorder` (implementation of `http.ResponseWriter`) so we don't have to define it in every test file.

maramihali:
Fixes #149 

Changed tests to use the `safehttp.DefaultDispatcher`, so a repeated `testDispatcher` does not have to be created in every test file that requires it. Exported the `http.ResponseWriter` implementation in `safehttptest` and renamed it to `TestResponseWriter` so it can be used in tests instead of repeatedly defining a `responseRecorder` in every test file that requires one.


safehttp/safehttptest/recorder.go
```diff
@@ -114,30 +69,40 @@ func (r *ResponseRecorder) Body() string {
 	return r.b.String()
 }
 
-// responseWriter is an implementation of the http.ResponseWriter interface used
-// for constructing an HTTP response.
-type responseWriter struct {
+// TestResponseWriter is an implementation of the http.ResponseWriter interface
+// used for constructing an HTTP response for testing purposes.
+type TestResponseWriter struct {
```
maramihali, line 74: This is really not a great name but I couldn't come up with a better one. :/

kele (approved):
Nice!

PR #177, opened 2020-09-18
Implemented the Commit phase in the CSP and XSRF plugins

### LINKED ISSUE #76: Add template nonce injection
The CSP and XSRF plugins currently put tokens in the request context.

* There needs to be a part of the plugins that can, later on, provide that nonce for a template to render.
* This should ideally work easily with different template implementations
* This will likely happen in the Commit phase of both plugins

maramihali:
Fixes #76 

Extended the `safehttp.TemplateResponse` to support adding elements to a function map. This will be executed before the template is parsed. The CSP nonce and XSRF Token values will be added to this map in the `xsrf.Commit` and `csp.Commit` phases.

### DIFF
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -62,7 +62,7 @@ func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) er
 
 // ExecuteTemplate applies the parsed template to the provided data object, if
 // the template is a safe HTML template, writing the output to the  http.
-// ResponseWriter. If the funcMap is non-nil, its elements can override the
+// ResponseWriter. If the funcMap is non-nil, its elements override the
 // existing names to functions mappings in the template. An attempt to define a
 // new name to function mapping that is not already in the template will result
 // in a panic. The template, data object and funcMap are contained in the
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -65,8 +65,7 @@ func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) er
 // ResponseWriter. If the funcMap is non-nil, its elements override the
 // existing names to functions mappings in the template. An attempt to define a
 // new name to function mapping that is not already in the template will result
-// in a panic. The template, data object and funcMap are contained in the
-// TemplateResponse.
+// in a panic.
 //
 // If an error occurs executing the template or writing its output,
 // execution stops, but partial results may already have been written to
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -60,27 +60,29 @@ func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) er
 	return json.NewEncoder(rw).Encode(resp.Data)
 }
 
-// ExecuteTemplate applies a parsed template to the provided data object, if the
-// template is a safe HTML template, writing the output to the  http.
+// ExecuteTemplate applies the parsed template to the provided data object, if
+// the template is a safe HTML template, writing the output to the  http.
 // ResponseWriter. If the funcMap is non-nil, its elements can override the
 // existing names to functions mappings in the template. An attempt to define a
 // new name to function mapping that is not already in the template will result
-// in a panic.
+// in a panic. The template, data object and funcMap are contained in the
+// TemplateResponse.
 //
 // If an error occurs executing the template or writing its output,
 // execution stops, but partial results may already have been written to
 // the Response Writer.
-func (DefaultDispatcher) ExecuteTemplate(rw http.ResponseWriter, t Template, data interface{}, funcMap map[string]interface{}) error {
+func (DefaultDispatcher) ExecuteTemplate(rw http.ResponseWriter, resp TemplateResponse) error {
+	t := *resp.Template
 	x, ok := t.(*template.Template)
 	if !ok {
-		return fmt.Errorf("%T is not a safe template and it cannot be parsed and written", t)
+		return fmt.Errorf("%T is not a safe template and it cannot be parsed and written", resp.Template)
 	}
-	if len(funcMap) == 0 {
-		return x.Execute(rw, data)
+	if len(resp.FuncMap) == 0 {
+		return x.Execute(rw, resp.Data)
 	}
 	cloned, err := x.Clone()
 	if err != nil {
 		return err
 	}
-	return cloned.Funcs(funcMap).Execute(rw, data)
+	return cloned.Funcs(resp.FuncMap).Execute(rw, resp.Data)
 }
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -60,9 +60,12 @@ func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) er
 	return json.NewEncoder(rw).Encode(resp.Data)
 }
 
-// ExecuteTemplate applies a parsed template to the provided data object if the
+// ExecuteTemplate applies a parsed template to the provided data object, if the
 // template is a safe HTML template, writing the output to the  http.
-// ResponseWriter.
+// ResponseWriter. If the funcMap is non-nil, its elements can override the
+// existing names to functions mappings in the template. An attempt to define a
+// new name to function mapping that is not already in the template will result
+// in a panic.
 //
 // If an error occurs executing the template or writing its output,
 // execution stops, but partial results may already have been written to
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -67,10 +67,17 @@ func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) er
 // If an error occurs executing the template or writing its output,
 // execution stops, but partial results may already have been written to
 // the Response Writer.
-func (DefaultDispatcher) ExecuteTemplate(rw http.ResponseWriter, t Template, data interface{}) error {
+func (DefaultDispatcher) ExecuteTemplate(rw http.ResponseWriter, t Template, data interface{}, funcMap map[string]interface{}) error {
 	x, ok := t.(*template.Template)
 	if !ok {
 		return fmt.Errorf("%T is not a safe template and it cannot be parsed and written", t)
 	}
-	return x.Execute(rw, data)
+	if len(funcMap) == 0 {
+		return x.Execute(rw, data)
+	}
+	cloned, err := x.Clone()
+	if err != nil {
+		return err
+	}
+	return cloned.Funcs(funcMap).Execute(rw, data)
 }
--- a/safehttp/plugins/csp/csp_test.go
+++ b/safehttp/plugins/csp/csp_test.go
@@ -200,6 +200,7 @@ func (errorReader) Read(b []byte) (int, error) {
 func TestPanicWhileGeneratingNonce(t *testing.T) {
 	randReader = errorReader{}
 	defer func() {
+		// TODO: avoid replacing the randReader per individual test cases
 		randReader = endlessAReader{}
 	}()
 	defer func() {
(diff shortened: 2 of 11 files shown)


safehttp/plugins/xsrf/xsrf.go
```diff
@@ -94,7 +94,7 @@ func addCookieID(w *safehttp.ResponseWriter) (*safehttp.Cookie, error) {
 // In case of state changing requests (all except GET, HEAD and OPTIONS), the
 // interceptor checks for the presence of the XSRF token in the request body
 // (expected to have been injected) and validates it.
-func (i *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest, cfg interface{}) safehttp.Result {
+func (it *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest, cfg interface{}) safehttp.Result {
```
bsiegert, line 97: nit: Why `it` instead of `i`? Single-letter receiver names are the recommended style.

tests/integration/csp/csp_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/google/go-safeweb/safehttp"
+	"github.com/google/go-safeweb/safehttp/plugins/csp"
+	"github.com/google/go-safeweb/safehttp/safehttptest"
+	safetemplate "github.com/google/safehtml/template"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestServeMuxInstallCsp(t *testing.T) {
```
bsiegert, line 27: nit: `TestServeMuxInstallCSP`

tests/integration/xsrf/xsrf_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/google/go-safeweb/safehttp"
+	"github.com/google/go-safeweb/safehttp/plugins/xsrf"
+	"github.com/google/go-safeweb/safehttp/safehttptest"
+	safetemplate "github.com/google/safehtml/template"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestServeMuxInstallXsrf(t *testing.T) {
```
bsiegert, line 27: nit: s/Xsrf/XSRF/

bsiegert (approved):

safehttp/plugins/xsrf/xsrf.go
maramihali: I made the modification to be consistent across the repository, since this was the only place where we used `i` instead of `it`. Especially in the functions from the `mux.go` file, it seemed a bit more readable to refer to interceptors as `it` instead of `i`.

safehttp/plugins/csp/csp_test.go
```diff
@@ -230,3 +233,119 @@ func TestNonceEmptyContext(t *testing.T) {
...
+	nonce, ok := tr.FuncMap["CSPNonce"]
+	if !ok {
+		t.Error(`tr.FuncMap["CSPNonce"]: got nil, want nonce`)
+	}
+
+	fn, ok := nonce.(func() string)
+	if !ok {
+		t.Errorf(`tr.FuncMap["CSPNonce"]: got %T, want "func() string"`, fn)
+	}
+	if want, got := "pizza", fn(); want != got {
```
kele, line 255: nit: Make the assignment inside the condition consistent with how you display it in tests, i.e. `got` before `want`.

safehttp/plugins/csp/csp_test.go
```diff
@@ -230,3 +233,119 @@ func TestNonceEmptyContext(t *testing.T) {
...
+		t.Errorf("rec.Body(): got %q want %q", got, want)
+	}
+
+}
+
+func TestBeforeCommit(t *testing.T) {
+	rec := safehttptest.NewResponseRecorder()
+	req := safehttptest.NewRequest(safehttp.MethodGet, "https://foo.com/pizza", nil)
+
+	i := Default("")
```
kele, line 326: `it`?

safehttp/plugins/csp/csp_test.go
```diff
@@ -230,3 +233,119 @@ func TestNonceEmptyContext(t *testing.T) {
...
+		t.Errorf("rec.Header() mismatch (-want +got):\n%s", diff)
+	}
+
+	if want, got := "", rec.Body(); got != want {
+		t.Errorf("rec.Body(): got %q want %q", got, want)
+	}
+
+}
+
+func TestBeforeCommit(t *testing.T) {
```
kele, line 322: What is this testing?

As a framework, we don't guarantee that e.g. the `ResponseWriter` or the request are going to stay the same between these calls.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -128,17 +128,33 @@ func (i *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingReq
...
+	tok := xsrftoken.Generate(it.SecretAppKey, cookieID.Value(), actionID)
 	r.SetContext(context.WithValue(r.Context(), tokenCtxKey{}, tok))
 	return safehttp.NotWritten()
 }
 
-// Commit is a no-op, required to satisfy the safehttp.Interceptor interface.
-func (i *Interceptor) Commit(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest, resp safehttp.Response, cfg interface{}) safehttp.Result {
+// Commit adds the XSRF token corresponding to the correct user to the
+// safehttp.TemplateResponse to be subsequently injected in HTML forms as a
+// hidden form field.
```
kele, line 143:
```suggestion
// Commit adds the XSRF token to the safehttp.TemplateResponse.FuncMap with key "XSRFToken".
// The token corresponds to the user information found in the request.
```

safehttp/response.go
```diff
@@ -38,12 +38,13 @@ type Template interface {
 	Execute(wr io.Writer, data interface{}) error
 }
 
-// TemplateResponse bundles a Template with its data to be passed together to the
-// commit phase. This will be passed when the commit phase is initiated from
-// ResponseWriter.WriteTemplate.
+// TemplateResponse bundles a Template with its data and names to function
+// mappings to be passed together to the commit phase. A TemplateResponse will
+// be created when the commit phase is initiated from ResponseWriter.WriteTemplate.
```
kele, line 43:
```suggestion
// TemplateResponse bundles a Template with its data and names to function
// mappings to be passed together to the commit phase.
```

kele: I left out information that doesn't seem to benefit the user (what are they supposed to do about it?).

kele (changes requested):
In the future: please split the PR. CSP and XSRF could've been done in separate ones and it would've been easier to review.

kele:
Let's leave the `FuncMap` in this PR, but it feels really clunky to me and seems to be heavily tailored to `safehtml/template`, which has a very specific API (`map[string]interface{}` doesn't look like something we *really* want to encourage).

safehttp/response.go
maramihali: Right, that's implementation details.

safehttp/plugins/csp/csp_test.go
maramihali: That is right, other interceptors or handlers could change the `ResponseWriter`, but I implemented this test to verify that the Before and Commit phases work together. Is there a better way to do it?

maramihali:
That is true, it's very specific to `safehtml/template`. It's not very clear to me how general the solution should be since a user-provided implementation of injecting nonces into the template should also probably be discouraged, right? I opened #184 anyway.

kele:
Depends who do you have in mind when you say "user". Handler author? Yeah, they shouldn't inject them themselves. Dispatcher authors? 100% yes, they might have their own `safehtml/template` equivalent.

safehttp/plugins/csp/csp_test.go
kele: Well, but this just tests that a nonce found it's way to the context. It just feels like a quite incomplete unit test.

Wouldn't it make more sense to instead create a test that actually runs through a handler? This way you can assert that:
- the headers are set with a given nonce
- the nonce is available inside the handler through `csp.Nonce()` call (this is important for migration, when a service wants to adopt the plugin to generate nonces and headers, but doesn't have automated nonce-injection implemented yet)
- the same nonce is passed in `FuncMap["CSPNonce"]` in the `TemplateResponse` and can be used by the dispatcher

kele: Similarily for the XSRF token.

maramihali: I see, yeah, that might be better. Which headers dym?

kele: CSP. :)

maramihali: In order to test that functionality we need a `ServeMux` on which to install the plugins so I think it's better if I extend the integration test.

kele: Fair. I'd remove this test then, as it doesn't really add much value if we have these unit tests for individual functions and we have an integration test for a real scenario (i.e. used with mux).

tests/integration/mux/mux_test.go
```diff
@@ -181,3 +187,44 @@ func TestMuxDefaultDispatcherUnsafeResponses(t *testing.T) {
...
+		}
+		t := safetemplate.Must(safetemplate.
+			New("name").Funcs(fns).
+			Parse(`
+			<script nonce="{{CSPNonce}}" type="application/javascript">
+			alert("script")
+			</script>
+			<form>
+			<input type="hidden" name="token" value="{{XSRFToken}}">{{.}}
+			</form>
```
kele, line 209:
```suggestion
			<script nonce="{{CSPNonce}}" type="application/javascript">
				alert("script")
			</script>
			<form>
				<input type="hidden" name="token" value="{{XSRFToken}}">{{.}}
			</form>
```

safehttp/plugins/xsrf/xsrf_test.go
```diff
@@ -309,9 +309,125 @@ func TestMissingCookiePostRequest(t *testing.T) {
...
+	i.Commit(rec.ResponseWriter, req, tr, nil)
+
+	tok, ok := tr.FuncMap["XSRFToken"]
+	if !ok {
+		t.Fatal(`tr.FuncMap["XSRFToken"] not found`)
+	}
+
+	fn, ok := tok.(func() string)
+	if !ok {
+		t.Errorf(`tr.FuncMap["XSRFToken"]: got %T, want "func() string"`, fn)
```
kele, line 335: Please apply my comment about `Fatal` in all the tests you just added.

kele (changes requested):

safehttp/default_dispatcher.go
```diff
@@ -60,17 +60,29 @@ func (DefaultDispatcher) WriteJSON(rw http.ResponseWriter, resp JSONResponse) er
...
-// ExecuteTemplate applies a parsed template to the provided data object if the
-// template is a safe HTML template, writing the output to the  http.
-// ResponseWriter.
+// ExecuteTemplate applies the parsed template to the provided data object, if
+// the template is a safe HTML template, writing the output to the  http.
+// ResponseWriter. If the funcMap is non-nil, its elements override the
+// existing names to functions mappings in the template. An attempt to define a
+// new name to function mapping that is not already in the template will result
+// in a panic. The template, data object and funcMap are contained in the
+// TemplateResponse.
```
kele, line 69:
```suggestion
// in a panic.
```

kele: One can just look at the type definition.

tests/integration/xsrf/xsrf_test.go
```diff
@@ -0,0 +1,66 @@
...
+	it := xsrf.Interceptor{SecretAppKey: "testSecretAppKey"}
+	mux.Install(&it)
+
+	var token string
+	var err error
+	handler := safehttp.HandlerFunc(func(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+		fns := map[string]interface{}{
+			"XSRFToken": func() string { return "WrongToken" },
+		}
+		token, err = xsrf.Token(r)
```
kele, line 38:
```suggestion
		// These are not used in the handler, just recorder to test whether
		// the handler can retrieve the XSRF token (e.g. for when the xsrf plugin
		// is installed, but auto-injection isn't yet supported and has to be done
		// manually by the handler).
		token, err = xsrf.Token(r)
```

tests/integration/xsrf/xsrf_test.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		t.Fatalf("xsrf.Token: got error %v", err)
+	}
+
+	if got, want := rr.Status(), safehttp.StatusOK; got != want {
+		t.Errorf("rr.Status() got: %v want: %v", got, want)
+	}
+
+	wantBody := `<form><input type="hidden" name="token" value="` +
+		token + `">Content</form>`
```
kele, line 61: Want to also check what the token is?

tests/integration/csp/csp_test.go
```diff
@@ -0,0 +1,77 @@
...
+	it := csp.Default("")
+	mux.Install(&it)
+
+	var nonce string
+	var err error
+	handler := safehttp.HandlerFunc(func(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+		fns := map[string]interface{}{
+			"CSPNonce": func() string { return "WrongNonce" },
+		}
+		nonce, err = csp.Nonce(r.Context())
```
kele, line 39: Please add a similar comment as I did for the XSRF test.

tests/integration/csp/csp_test.go
```diff
@@ -0,0 +1,77 @@
...
+		"Content-Type": {"text/html; charset=utf-8"},
+		"Content-Security-Policy": {
+			"object-src 'none'; script-src 'unsafe-inline' 'nonce-" + nonce + "' 'strict-dynamic' https: http:; base-uri 'none'",
+			"frame-ancestors 'self'"},
+	}
+	if diff := cmp.Diff(wantHeaders, map[string][]string(rr.Header())); diff != "" {
+		t.Errorf("rr.Header(): mismatch (-want +got):\n%s", diff)
+	}
+
+	wantBody := `<script nonce="` + nonce +
```
kele, line 71: Could you also assert what the nonce is?

kele (approved):
Good work!

PR #183, opened 2020-09-24
Rename the package name of the initial form parameters and query parameters test

maramihali:
Renamed the initial form parameters and query parameters test packages so they build when using Google internal tooling conventions. Otherwise, the build rules generated by running `glaze` will result in errors. This changes do not affect building the packages in the open-source.


kele (changes requested):
Please add a PR description why you're making this change.

kele (approved):

PR #185, opened 2020-09-29
Host checks are now inside a hostcheck plugin.

### LINKED ISSUE #168: Consider making host checks in a plugin rather than in the core mux implementation
Currently, host checking is done here: https://github.com/google/go-safeweb/blob/0c930b079f42206df056e531d88cf9009caa6f97/safehttp/mux.go#L173

It might make sense to do it in an interceptor. Especially since we might need to provide convenient ways of configuring the check (e.g. for local development against `localhost`).

kele:
Host checks were performed inside the mux to protect from DNS rebinding
and HTTP request smuggling attacks. There is no real reason why these
should be inside the mux implementation instead of a plugin.

Fixes #168.

### DIFF
--- a/safehttp/plugins/hostcheck/hostcheck_test.go
+++ b/safehttp/plugins/hostcheck/hostcheck_test.go
@@ -1,0 +1,52 @@
+package hostcheck_test
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/google/go-safeweb/safehttp"
+	"github.com/google/go-safeweb/safehttp/plugins/hostcheck"
+	"github.com/google/go-safeweb/safehttp/safehttptest"
+	"github.com/google/safehtml"
+)
+
+func TestInterceptor(t *testing.T) {
+	var test = []struct {
+		name       string
+		req        *http.Request
+		wantStatus safehttp.StatusCode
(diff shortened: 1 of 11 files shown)


safehttp/plugins/hostcheck/hostcheck_test.go
```diff
@@ -0,0 +1,66 @@
...
+			h := safehttp.HandlerFunc(func(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+				return w.Write(safehtml.HTMLEscaped("<h1>Hello World!</h1>"))
+			})
+			mux.Handle("/", safehttp.MethodGet, h)
+
+			b := &strings.Builder{}
+			rw := safehttptest.NewTestResponseWriter(b)
+			mux.ServeHTTP(rw, tt.req)
+
+			if rw.Status() != tt.wantStatus {
```
maramihali, line 61: nit: also check the headers and response body are the expected ones

maramihali (approved):

safehttp/plugins/hostcheck/hostcheck_test.go
kele: I deliberately removed these. I'm not convinced that this should be tested here.

Issue #117, opened 2020-08-25
Make sure ServeMux is used in the right order


empijei:
The `ServeMux` construction goes through phases:

1. Plugin registration
2. Handlers registration and configuration
3. Execution

Nothing that is done in one of the phases should be done in one of the others.

I think we should enforce this somehow, by either providing builders for the phases or panicking if something is done out of order at startup, and then ignoring what happens at runtime (e.g. registering a plugin or a handler after a request has been served will log an error and do nothing).

bsiegert:
In Goa, we have state variables that we set and check with atomics and return an error if in the wrong phase. For instance, if you want to add a module to a running server, it will be rejected this way.

empijei:
> return an error if in the wrong phase

Wouldn't it be better to try and avoid this at compile time or startup time?

empijei:
My proposal is an API that builds the mux in stages.

For example (this is just on the top of my head, don't take it as a proposal):
```go
myHandler := safehttp.ServeMuxBuilder(dispatcher, domains)
  .InstallInterceptor(...)
  .InstallInterceptor(...)
  .Mux()
  .Handle(...)
  .Handle(...)
  .Handler()
```

Where:
* `ServeMuxBuilder` returns something that only has the `InstallInterceptor` and `Mux` methods.
* `Mux` returns something that only has the `Handle` and `Handler` methods
* `Handler` returns a `safehttp.Handler`

empijei:
@kele WDYT?

kele:
I like it.

empijei:
Alternatively we could just error or make out-of-order operations do nothing and log the error.

We could also panic.

I would just like our code to not have to care about things happening in an unreasonable sequence.

maramihali:
Should this be v0?

kele:
No. Focus on existing V0 tasks.

kele:
Isn't this fixed by the `ServeMuxBuilder` from #191?

Issue #88, opened 2020-08-14
Conformance checks during mux startup


kele:
Create the ability of adding conformance checks that run after the mux has been set up. These would be able to:
- list all the plugins registered for each handler, together with the method, path and all other information
- ask plugins for their configuration

Conformance checks would need to cooperate with plugins, i.e. there is going to be no complex protocol for information exchange between a plugin and the conformance check. A simple `func (p Plugin) Config() interface{}` would be enough, as the conformance check needs to have full trust in the plugins anyway.

PR #191, opened 2020-10-06
Introducing ServeMuxBuilder

kele:
Related issues: #117, #88.

This is the first draft of a `ServeMuxBuilder`. Right now it's just a thin wrapper that contains virtually no logic.

In the nearest future, `ServeMux` will be simplified and some of the logic will be moved into `ServeMuxBuilder` instead.

### DIFF
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -164,8 +164,8 @@ type handlerRegistration struct {
 	cfgs    []InterceptorConfig
 }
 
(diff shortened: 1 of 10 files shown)


maramihali (approved):
Unrelated question but why rename to `standardHandler`?

kele:
> Unrelated question but why rename to `standardHandler`?

My reasoning here was to indicate that this will actually be registered on the `standard` mux.

PR #192, opened 2020-10-07
Allow the user to set a custom 1xx or 2xx status code in the safehttp.ResponseWriter

### LINKED ISSUE #151: Add support to set 2XX statuses that are not 200 and 204

maramihali:
Fixes #151

Previously, there was no way of setting a status code other than `safehttp.NoContent` as the write methods in the `safehttp.ResponseWriter` would  always set the status code to `safehttp.StatusOK`. Implemented initial functionality and tests to make this possible.

### DIFF
--- a/safehttp/response_writer_test.go
+++ b/safehttp/response_writer_test.go
@@ -143,19 +143,10 @@ func TestResponseWriterStatusCode(t *testing.T) {
 			name: "Custom status code before calling write",
 			want: safehttp.StatusCreated,
 			write: func(w *safehttp.ResponseWriter) {
-				w.Code = safehttp.StatusCreated
+				w.SetStatusCode(safehttp.StatusCreated)
 				w.Write(safehtml.HTMLEscaped("<h1>Escaped, so not really a heading</h1>"))
 			},
 		},
-		{
-			name: "Custom status code after calling write",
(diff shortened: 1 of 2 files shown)


safehttp/response_writer_test.go
```diff
@@ -125,3 +125,103 @@ func TestResponseWriterWriteTwicePanic(t *testing.T) {
...
+				w.Write(safehtml.HTMLEscaped("<h1>Escaped, so not really a heading</h1>"))
+			},
+		},
+		{
+			name: "Custom status code after calling write",
+			want: safehttp.StatusOK,
+			write: func(w *safehttp.ResponseWriter) {
+				w.Write(safehtml.HTMLEscaped("<h1>Escaped, so not really a heading</h1>"))
+				w.Code = safehttp.StatusCreated
+			},
```
kele, line 156: Why is this considered good?

kele:
I don't think we want to set the status code by just assigning it to a `ResponseWriter` field.

safehttp/response_writer_test.go
maramihali: it's true it shouldn't be allowed and should probably trigger an error

maramihali:
Closing as this was an experimental PR

PR #195, opened 2020-10-09
Implemented XSRF protection for Angular

### LINKED ISSUE #194: Extend the XSRF plugin to provide protection for Angular XHR
See: https://docs.angularjs.org/api/ng/service/$http#cross-site-request-forgery-xsrf-protection for more details on how this can be acheived.

maramihali:
Fixes #194 

Extended the XSRF plugin so it can be configured to provide XSRF protection for Angular. This is achieved by generating a cryptographically random token and assigning it to the `XSRF-TOKEN` cookie, set in the response to the first `GET` request. On all subsequent requests, the value of this cookie is expected to match the value of a custom request header, `X-XSRF-TOKEN`.


bsiegert (approved):

maramihali:
Closed as this will be then as a follow-up PR for #199.

Issue #174, opened 2020-09-17
htmlinject: Export constants for noncing functions names


empijei:
When using htmlinject it is required to call `Funcs` on the template with the names used by htmlinject (e.g. CSPNonce and XSRFToken), but htmlinject currently does not export those constants, making it brittle.

Issue #176, opened 2020-09-18
htmlinject: Add nonces to style tags


empijei:
We currently only nonce scripts and links.

PR #197, opened 2020-10-09
htmlinject: added rule for styles and exported useful constants.

empijei:
It was previously hard and brittle to use htmlinject as some constants were not exported (the ones to use as map keys in funcmap).

Moreover we were not noncing syles because our CSP plugin does not set the style-src directive, but users might want to use a different, stricter, CSP implementation.


empijei:
This fixes #174 and #176.

PR #198, opened 2020-10-09
htmlinject: provide helpers to load safehtml.Templates

### LINKED ISSUE #173: htmlinject: Provide a helper to parse templates
In order to use htmlinject the user needs to:

* Create a new empty template
* Call `Funcs` on it to make it so the functions `Parse` expects are defined (e.g. passing no-op functions that have the right signature and name for csrftoken and xsrftoken generation)
* Load the template source
* Use htmlinject to rewrite the template
* Pass the result to `Parse` on the aforementioned template

The code would more or less look like this:
```go
func parsetpl(path string) (*template.Template, error) {
	f, err := os.Open(path)
	if err != nil {
		return nil, err
	}
	noop := func() string { panic("this function should never be called") }
	t := template.New("").Funcs(map[string]interface{}{"CSPNonce": noop, "XSRFToken": noop})
	transf, err := htmlinject.Transform(f, htmlinject.CSPNoncesDefault, htmlinject.XSRFTokensDefault)
	if err != nil {
		return nil, err
	}
	return t.Parse(transf)
}
```

It would be nice if htmlinject provided helpers to perform all of those operations with a single call, and maybe even an helper to do so while loading templates from disk (like the standard html/template package does).

empijei:
It was previously impossible to use htmlinject together with safehtml without an uncheckedconversion.

Depends on PR #197 
Fixes #173

### DIFF
--- a/safehttp/plugins/htmlinject/htmlinject.go
+++ b/safehttp/plugins/htmlinject/htmlinject.go
@@ -19,8 +19,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
+	"path/filepath"
 	"strings"
 
+	"github.com/google/safehtml/template"
+	"github.com/google/safehtml/template/uncheckedconversions"
 	"golang.org/x/net/html"
 )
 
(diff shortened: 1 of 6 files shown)


safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -106,6 +121,85 @@ func Transform(src io.Reader, cfg ...Config) (string, error) {
 	return rw.out.String(), nil
 }
 
+// LoadTrustedTemplate processes the given TrustedTemplate with the specified default configurations and
+// adds it to the given template.
+// If the given template is nil a new one is created.
+func LoadTrustedTemplate(tpl *template.Template, csp, xsrf bool, src template.TrustedTemplate) (*template.Template, error) {
```
kele, line 127: Let's get these all into a separate package, WDYT?

kele, line 127: These two boolean args next to each other might hurt readability and be error prone.

If you end up opting for a separate package, may I suggest instead:

```
package safehtmlinject

// wrapping so this package controls which ones are safe
type Config struct {
  cfg htmlinject.Config
}

func CSP() Config { return Config{htmlinject.CSPNoncesDefault} }

func XSRF() Config { return XSRF{htmlinject.XSRFTokensDefault } }

func LoadTrustedTemplate(tpl *template.Template, src template.TrustedTemplate, cfg ...Config) (*template.Template, error) { ... }
```

kele (changes requested):

safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -106,6 +121,85 @@ func Transform(src io.Reader, cfg ...Config) (string, error) {
...
+}
+
+// LoadGlob matches the behavior of safehtml.ParseGlob but runs a transformation on every loaded template.
+func LoadGlob(tpl *template.Template, csp, xsrf bool, pattern template.TrustedSource) (*template.Template, error) {
+	filenames, err := filepath.Glob(pattern.String())
+	if err != nil {
+		return nil, err
+	}
+	if len(filenames) == 0 {
+		return nil, fmt.Errorf("htmlinject: pattern matches no files: %#q", pattern.String())
```
kele, line 161: The `htmlinject:` part is known to the caller, so it should probably be omitted here.

kele: Other places too.

kele (changes requested):

safehttp/plugins/htmlinject/htmlinject.go
empijei: What is the value in adding fragmentation to two things that would almost always be used together? Also, note that this package has an extremely small API, I don't think this is bloated. Moreover this package was created to do exactly this job, I am not in favor of creating another to do that job.

empijei: Done.

kele: My feeling was that you aim to create an `htmlinject` package that isn't supposed to work with `safehtml/template` only.

safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -106,6 +121,93 @@ func Transform(src io.Reader, cfg ...Config) (string, error) {
...
+	DisableCsp bool
+	// DisableXsrf disables XSRF token injection
+	DisableXsrf bool
+}
+
+// LoadTrustedTemplate processes the given TrustedTemplate with the specified default configurations and
+// adds it to the given template.
+// If the given template is nil a new one is created.
+func LoadTrustedTemplate(tpl *template.Template, lcfg LoadConfig, src template.TrustedTemplate) (*template.Template, error) {
+	// Using bools to select configs prevents users to accidentally specify arbitrary and potentially unsafe rules.
```
kele, line 136: This comment is no longer relevant, right?

safehttp/plugins/htmlinject/htmlinject.go
```diff
@@ -106,6 +121,93 @@ func Transform(src io.Reader, cfg ...Config) (string, error) {
...
+			tpl = template.New(name)
+		}
+		if name == tpl.Name() {
+			t = tpl
+		} else {
+			t = tpl.New(name)
+		}
+		// We are loading a file from a TrustedSource, so this conversion is safe.
+		tts := uncheckedconversions.TrustedTemplateFromStringKnownToSatisfyTypeContract(string(b))
+		_, err = LoadTrustedTemplate(t, lcfg, tts)
```
maramihali, line 203: very optional: wouldn't the code be easier to follow if `LoadTrustedTemplate` would be placed after this function?

safehttp/plugins/htmlinject/htmlinject.go
empijei: Transform is everything you need for html/template.

This patch is just to make it *also* usable with safehtml/template.

empijei: If you think it would be useful I can duplicate the helpers to also support html/template.

safehttp/plugins/htmlinject/htmlinject.go
empijei: I see what you mean, so I rearranged the functions to be used only *after* they have been defined (to read the code top-down).

safehttp/plugins/htmlinject/htmlinject.go
kele: I just don't see a good reason to put it here as it seems like a different level of abstraction (i.e. `htmlinject.Transform` and `LoadTrustedTemplate`).

If you have a reason to not do the split (I don't see harm in doing it) then that's okay. LGTM overall as I don't see it as blocking.

kele (approved):

safehttp/plugins/htmlinject/htmlinject.go
empijei: My reason is just to not bloat our directory structure, and these functions seemed related enough to go in the same package.

We can always do the split in a follow up PR, let's discuss this in person.

PR #199, opened 2020-10-12
Refactored the XSRF plugin to a more general structure

maramihali:


safehttp/plugins/xsrf/xsrf.go
```diff
@@ -22,142 +22,234 @@ import (
...
 
-// Interceptor implements XSRF protection.
 type Interceptor struct {
-	// SecretAppKey uniquely identifies each registered service and should have
-	// high entropy as it is used for generating the XSRF token.
-	SecretAppKey string
+	secretAppKey string
+	g            Generator
+	c            Checker
+	i            Injector
```
kele, line 46: `Checker` should check whether the request follows XSRF protection protocol (i.e. for state-changing it should require an XSRF token; the `Checker` is responsible for taking the request and determining whether it is fine or not). If not, I'd make the `Checker` return that information as opposed to directly writing the response. The XSRF plugin itself should do it.

`Generator` should generate the XSRF tokens (or whatever needs to be in a cookie). It's not clear to me what it should return, but I don't think it should do anything with the `ResponseWriter`.

`Injector` should get whatever `Generator` returned and apply it to the safe response. Due to the fact that cookies are tied to `ResponseWriter` and not to the safe response itself, it needs to accept a `ResponseWriter` alongside the safe response.

cc: @empijei to double-check we have a shared vision.

kele (changes requested):
High-level comments, didn't look at readability at all.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -23,141 +23,249 @@ import (
...
-// interceptor checks for the presence of the XSRF token in the request body
-// (expected to have been injected) and validates it.
 func (it *Interceptor) Before(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest, _ safehttp.InterceptorConfig) safehttp.Result {
-	needsValidation := !statePreservingMethods[r.Method()]
-	cookieID, err := r.Cookie(cookieIDKey)
+	code := it.c.Check(r)
+	if code != safehttp.StatusOK {
+		return w.WriteError(code)
+	}
+	return safehttp.NotWritten()
```
kele, line 86: :+1:

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -23,141 +23,249 @@ import (
...
+}
+
+func (it *Interceptor) Commit(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest, resp safehttp.Response, _ safehttp.InterceptorConfig) safehttp.Result {
+	data, err := it.g.Generate(r)
```
kele, line 90: `data` is too broad. Please find a better name.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -23,141 +23,249 @@ import (
...
+	if err != nil {
+		buf := make([]byte, 20)
+		if _, err := rand.Read(buf); err != nil {
+			return nil, fmt.Errorf("crypto/rand.Read: %v", err)
 		}
+		c = safehttp.NewCookie(g.cookieIDKey, base64.StdEncoding.EncodeToString(buf))
+		c.SetSameSite(safehttp.SameSiteStrictMode)
+		data.setCookie = true
 	}
+	data.cookieID = c
```
kele, line 137: Creating cookies should be done in `Injector`, not `Generator`.

For instance, how weird would it be to set cookie strictness in `Generator`?

kele (changes requested):
Let's just another way:

Angular should have a dedicated `Checker` and `Injector`, **not** `Generator`.
Tink should have a dedicated `Generator`, not `Checker` and not `Injector`.

safehttp/plugins/xsrf/xsrf.go
empijei: > Checker should check whether the request follows XSRF protection protocol (i.e. for state-changing it should require an XSRF token; the Checker is responsible for taking the request and determining whether it is fine or not). If not, I'd make the Checker return that information as opposed to directly writing the response. The XSRF plugin itself should do it.

I agree

> Generator should generate the XSRF tokens (or whatever needs to be in a cookie). It's not clear to me what it should return, but I don't think it should do anything with the ResponseWriter.

We cannot generalize this: some defences will be a cookie, some others will be a token for the template, some will be headers. How would you do it? And what would you inject, where and how?

maramihali: The `Injector` should probably take care of ensuring all protection is there (i.e. it should set cookies and/or inject the template with the token, based on the case) so "inject the protection". Currently, generating the token is also done in the `Injector`. It is rather hard to abstract away the `Generator` (the thing that should generate the token) from the `Injector`, as the info it needs depends on whether the plugin provides protection for Angular or for the "default" case as well as what crypto stuff it uses to generate the token. Hence, a `Generate` function would take different arguments based on the case. One alternative I could think of is the following:
```
type Generator interface {
Generate(r *safehttp.IncomingRequest, data ...interface{}) (string, error)
}
```
Then, in the "default case":
- we pass the `secretAppKey` and `cookieID` in the `Inject` function alongside the request and `Generate` generates the token either using Tink or `xsrftoken`

and in the Angular case:
- we could pass a `secretAppKey with the request and the token to be set in the cookie could be generated using that key or simply using `crypto/rand` (afaiu, pre-login protection is not an issue in Angular, right?)

kele: As discussed: focus on the interfaces that `xsrf` plugin needs first.

One good exercise you can do: write comments for your interface methods. It will make it clearer to you how to divide the responsibilities.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -18,146 +18,229 @@
...
 
-type tokenCtxKey struct{}
-
-// Token extracts the XSRF token from the incoming request. If it is not
-// present, it returns a non-nil error.
-func Token(r *safehttp.IncomingRequest) (string, error) {
-	tok := r.Context().Value(tokenCtxKey{})
-	if tok == nil {
-		return "", errors.New("xsrf token not found")
+func New(key string, c Checker, i Injector) Interceptor {
```
empijei, line 42: Nit: I would name the parameter secretAppKey.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -18,146 +18,229 @@
...
 	safehttp.MethodHead:    true,
 	safehttp.MethodOptions: true,
 }
 
-// Interceptor implements XSRF protection.
 type Interceptor struct {
-	// SecretAppKey uniquely identifies each registered service and should have
-	// high entropy as it is used for generating the XSRF token.
-	SecretAppKey string
+	secretAppKey string
```
empijei, line 35: Nit: maybe key will suffice?

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -18,146 +18,229 @@
...
-		// not the case, a server misconfiguration occured.
-		return w.WriteError(safehttp.StatusInternalServerError)
+		return nil
 	}
 
 	// TODO(maramihali@): Change the key when function names are exported by
 	// htmlinject
 	// TODO: what should happen if the XSRFToken key is not present in the
 	// tr.FuncMap?
 	tmplResp.FuncMap["XSRFToken"] = func() string { return tok }
```
empijei, line 192: This constant is now exported by htmlinject, you might want to use that.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -18,146 +18,229 @@
...
+	Retrieve(r *safehttp.IncomingRequest) (token, userID, actionID string, code safehttp.StatusCode)
+	// Validate validates the token contained in the request, optionally using
+	// the user session and action the user is taking.
+	Validate(token string, userID string, actionID string) safehttp.StatusCode
+}
+
+type Injector interface {
+	// Inject adds the protection necessary to the safehttp.Response and
+	// safehttp.ResponseWriter so that subsequent request can be deemed safe.
+	Inject(resp safehttp.Response, w *safehttp.ResponseWriter, r *safehttp.IncomingRequest) error
```
empijei, line 114: Nite: for sake of coherence let's preserve the order of the Commit phase: w, r, resp

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -18,146 +18,229 @@
...
+	}
+	tok, userID, actionID, code := it.c.Retrieve(r)
+	if code != safehttp.StatusOK {
+		return w.WriteError(code)
+	}
+	code = it.c.Validate(tok, userID, actionID)
+	if code != safehttp.StatusOK {
+		return w.WriteError(code)
+	}
+	return safehttp.NotWritten()
```
empijei, line 89: @kele I think we might have overdone the abstraction bit a tad too much.

All of this CL and all of this work is just to allow us to not repeat these 12 lines?

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -18,146 +18,229 @@
...
+type angularChecker struct {
+	tokenCookieName string
+	tokenHeaderName string
+}
+
+func (c angularChecker) Retrieve(r *safehttp.IncomingRequest) (token, userID, actionID string, code safehttp.StatusCode) {
+	cookie, err := r.Cookie(c.tokenCookieName)
+	if err != nil {
+		code = safehttp.StatusForbidden
+		return
```
empijei, line 205: Try to avoid using naked returns, especially if you have patterns like 

```
assignment to return value
naked return
```

This line would be more readable like 
```return "","","", safehttp.StatusForbidden```
since it would highlight what you are doing.

I understand it is harder to write code like this, but I think the readability improvement is worth.

Same for other naked returns.

File: safehttp/plugins/xsrf/xsrf.go [hunk 11]
```diff
@@ -18,146 +18,229 @@
...
+	if token == "" || token != cookie.Value() {
+		code = safehttp.StatusUnauthorized
+		return
+	}
+	return
+}
+
+func (c angularChecker) Validate(_, _, _ string) safehttp.StatusCode {
+	return safehttp.StatusOK
+}
```
empijei, line 217: I know we already discussed this, but maybe it is worth adding a comment on to the why we do not need to re-validate this.

safehttp/plugins/xsrf/xsrf.go
```diff
@@ -18,146 +18,229 @@
...
+	tok := make([]byte, 20)
+	if _, err := rand.Read(tok); err != nil {
+		return fmt.Errorf("crypto/rand.Read: %v", err)
+	}
+	c = safehttp.NewCookie(i.tokenCookieName, base64.StdEncoding.EncodeToString(tok))
+
+	c.SetSameSite(safehttp.SameSiteStrictMode)
+	c.SetPath("/")
+	// Set the duration of the token cookie to 24 hours.
+	c.SetMaxAge(86400)
```
empijei, line 237: We should be coherent and expire the other XSRF tokens in the same amount of time.

I think 24 hours is a bit tight, what about setting this and the other to a week?

safehttp/plugins/xsrf/xsrf.go
kele: > @kele I think we might have overdone the abstraction bit a tad too much.

Hard for me to say until I see the end result. Right now it is still super weird that `Retrieve` and `Validate` belong to the same interface. `Retrieve` can be front-end specific, whereas `Validate` can be crypto-specific. And if we think the second argument is not possible enough, then `Retrieve` and `Validate` should simply be one call: `Check()` and then it all gets simplified.

kele:
Angular has it's own structure now. Let's revisit once we decide to add Tink support.

PR #202, opened 2020-10-12
Added the OnError method to the safehttp.Interceptor interface

### LINKED ISSUE #201: Extend the Interceptor interface with an OnError method
This will run when the `safehttp.ResponseWriter.WriteError` method is called, before the error response is actually written, similarly to how `Commit` is run in the other methods for writing to the `ResponseWriter`.

maramihali:
Fixes #201 

The `OnError` method of all interceptors associated with a handler will be called in `safehttp.ResponseWriter.WriteError` before the error response is actually written.

### DIFF
--- a/safehttp/interceptor.go
+++ b/safehttp/interceptor.go
@@ -30,6 +30,11 @@ type Interceptor interface {
 	// is written to the ResponseWriter, then the Commit phases from the
 	// remaining interceptors won't execute.
 	Commit(w *ResponseWriter, r *IncomingRequest, resp Response, cfg InterceptorConfig) Result
+
+	// OnError runs when ResponseWriter.WriteError is called, before the
+	// actual error response is written. An attempt to write to the
+	// ResponseWriter in this phase will result in an irrecoverable error.
+	OnError(w *ResponseWriter, r *IncomingRequest, resp Response, cfg InterceptorConfig) Result
 }
 
 // InterceptorConfig is a configuration of an interceptor.
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -295,7 +295,7 @@ func (h handler) commitPhase(w *ResponseWriter, resp Response) {
 // ResponseWriter.
 //
 // TODO: BIG WARNING, if an interceptor attempts to write to the ResponseWriter
-// in the OnError phase then this method will recurse.
+// in the OnError phase will result in an irrecoverable error.
 func (h handler) errorPhase(w *ResponseWriter, resp Response) {
 	for i := len(h.interceps) - 1; i >= 0; i-- {
 		it := h.interceps[i]
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -289,3 +289,16 @@ func (h handler) commitPhase(w *ResponseWriter, resp Response) {
 		}
 	}
 }
+
+// errrorPhase calls the OnError phases of all the interceptors associated with
+// a handler. This stage runs before an error response is written to the
+// ResponseWriter.
+//
+// TODO: BIG WARNING, if an interceptor attempts to write to the ResponseWriter
+// in the OnError phase then this method will recurse.
+func (h handler) errorPhase(w *ResponseWriter, resp Response) {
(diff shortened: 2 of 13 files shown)


safehttp/interceptor.go
```diff
@@ -30,6 +30,10 @@ type Interceptor interface {
 	// is written to the ResponseWriter, then the Commit phases from the
 	// remaining interceptors won't execute.
 	Commit(w *ResponseWriter, r *IncomingRequest, resp Response, cfg InterceptorConfig) Result
+
+	// OnError runs when ResponseWriter.WriteError is called, before the
+	// actual error response is written.
```
kele, line 35: There are some extra expectations here, right? I.e. any error inside `OnError` is irrecoverable and results in a 500 (best case) or it being completely ignored.

safehttp/mux.go
```diff
@@ -289,3 +289,16 @@ func (h handler) commitPhase(w *ResponseWriter, resp Response) {
 		}
 	}
 }
+
+// errrorPhase calls the OnError phases of all the interceptors associated with
+// a handler. This stage runs before an error response is written to the
+// ResponseWriter.
+//
+// TODO: BIG WARNING, if an interceptor attempts to write to the ResponseWriter
+// in the OnError phase then this method will recurse.
```
kele, line 298: Is this actually true?

kele (changes requested):

safehttp/interceptor.go
maramihali: Yep, I thought a bit more about this and I think, in this PR, an attempt to write to the `ResponseWriter` in `OnError` should just trigger a panic.

safehttp/mux.go
maramihali: I changed it to panic if that happens.

safehttp/mux_test.go
```diff
@@ -597,3 +659,41 @@ func TestMuxHandlerReturnsNotWritten(t *testing.T) {
...
+	registeredHandler := safehttp.HandlerFunc(func(w *safehttp.ResponseWriter, r *safehttp.IncomingRequest) safehttp.Result {
+		return w.Write(safehtml.HTMLEscaped("<h1>Hello World!</h1>"))
+	})
+	mb.Handle("/bar", safehttp.MethodGet, registeredHandler)
+
+	defer func() {
+		if r := recover(); r == nil {
+			t.Errorf(`mb.Mux().ServeHTTP(rw, req): expected panic`)
+		}
+	}()
```
empijei, line 696:
```suggestion
	defer func() {
		if r := recover(); r != nil {
			return
		}
		t.Errorf(`mb.Mux().ServeHTTP(rw, req): expected panic`)
	}()
```

SuperNit: Reading a recover with a `== nil` condition might be surprising for some users.

safehttp/interceptor.go
```diff
@@ -30,6 +30,11 @@ type Interceptor interface {
 	// is written to the ResponseWriter, then the Commit phases from the
 	// remaining interceptors won't execute.
 	Commit(w *ResponseWriter, r *IncomingRequest, resp Response, cfg InterceptorConfig) Result
+
+	// OnError runs when ResponseWriter.WriteError is called, before the
+	// actual error response is written. An attempt to write to the
+	// ResponseWriter in this phase will result in an irrecoverable error.
+	OnError(w *ResponseWriter, r *IncomingRequest, resp Response, cfg InterceptorConfig) Result
```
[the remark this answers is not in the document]
kele: @empijei Interesting, I haven't considered this. Would you consider this an extra feature we can revisit?

kele (approved):

safehttp/interceptor.go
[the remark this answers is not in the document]
maramihali: Added a TODO

maramihali:
Rebased on top of master, ready for merge.

safehttp/interceptor.go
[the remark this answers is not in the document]
empijei: We should revisit this in the future. Definitely not a blocker.

PR #204, opened 2020-10-13
coop: internal refactor

empijei:
As discussed with @kele this provides better separation between data and code.

### DIFF
--- a/safehttp/plugins/coop/coop.go
+++ b/safehttp/plugins/coop/coop.go
@@ -49,18 +49,26 @@ func (p Policy) String() string {
 	return string(p.Mode) + `; report-to "` + p.ReportingGroup + `"`
 }
 
-// NewInterceptor constructs an interceptor that applies the given policies.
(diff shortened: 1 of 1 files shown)


safehttp/plugins/coop/coop.go
```diff
@@ -49,18 +49,26 @@ func (p Policy) String() string {
...
-			enf = append(enf, p.String())
+			s.enf = append(s.enf, p.String())
 		}
 	}
-	return Interceptor{rep: rep, enf: enf}
+	return s
+}
+
+// NewInterceptor constructs an interceptor that applies the given policies.
+func NewInterceptor(policies ...Policy) Interceptor {
```
maramihali, line 70: Why is `New` as function name not enough?

empijei: This package exports other types than just Interceptor, so I just added the extra bit to make sure ppl did not confuse this for an overrider or smth else.

PR #205, opened 2020-10-13
ResponseWriter and Dispatcher now have a single Write method

### LINKED ISSUE #150: Consider removing WriteTemplate and WriteJSON, implement helpers to just rely on Write
### Rationale

Currently we have a quite low-level API (`Write(Response)`) and two very high-level one, which are tailored to a specific function.

My proposal is to just stick to the general purpose API (Write) which is the one responsible for the wiring with the Dispatcher.

This will keep the interface for Dispatcher and ResponseWriter simpler, and will make it clear that the responsibility for wrapping and handling different types is on the Dispatcher implementers.

This will also remove the duplicated code that we have that asks for content-type in every call to WriteTemplate, WriteJSON etc.

### Template

The specific type we require is a `Template` interface but we actually ask more of it: for example when we will need to nonce-inject we will need Clone and Funcs methods, and we cannot express the Clonable interface, so we will be asking users to implement something, but will **implicitly ask for more**.

Moreover this API does not work if we have to call ExecuteTemplate instead, which will require an additional string parameter.

To then render templates we would offer helpers that would look like these: `w.Write(mytpl.WithData(data))` `w.Write(mytpl.NamedWithData("foo", data))`. If we don't want to change the packages that export the templates we can also put those helpers and wrappers in safehttp directly (which we are already doing anyways, even if in another form). With the extra added benefit that the signature will match the exact types we expect.

### JSON

Currently WriteJSON just wraps things in a JsonResponse. We could just create an helper for that and pass JSONResponse to Write. Example usage would look like `w.Write(safehttp.JsonResponse{data})` instead of `w.WriteJSON(data)` so it will be slightly more verbose but I think the overall API will benefit from the simplification.

### An optional extra step
We could even make the so-created `Response` implementations have an (unexported) method that takes care of properly setting the content-type and one for serializing the body, which would allow us to actually limit the `safehttp.Response` implementers to be checked at compile-time.

At that point the dispatcher responsibility would just be to check that something extraneous to safehttp is an acceptably safe response when put in a generic wrapper.

Overall we would reduce the typeswitching a lot and make the API cleaner.

maramihali:
Fixes #150 

Unified the functionality of `WriteJSON`, `ExecuteTemplate` and `Write` inside the new `safehttp.DefaultDispatcher.Write` method, also responsible for setting the `Content-Type` header and writing the status code.

### DIFF
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -26,6 +26,25 @@ import (
 // DefaultDispatcher is responsible for writing safe responses.
 type DefaultDispatcher struct{}
 
+// ContentType returns the Content-Type of a response if it's deemed safe and
+// an error otherwise.
+func (DefaultDispatcher) ContentType(resp Response) (string, error) {
+	switch x := resp.(type) {
+	case safehtml.HTML:
+		return "text/html; charset=utf-8", nil
+	case TemplateResponse:
+		_, ok := (*x.Template).(*template.Template)
+		if !ok {
+			return "", fmt.Errorf("%T is not a safe response type, a Content-Type cannot be provided", resp)
+		}
+		return "text/html; charset=utf-8", nil
+	case JSONResponse:
+		return "application/json; charset=utf-8", nil
+	default:
+		return "", fmt.Errorf("%T is not a safe response type, a Content-Type cannot be provided", resp)
+	}
+}
+
 // Write writes the response to the http.ResponseWriter if it's deemed safe.
 // A safe response is either safe HTML, a JSON object or safe HTML template. It
 // will  return a non-nil error if the response is deemed unsafe or if writing
@@ -33,8 +52,6 @@ type DefaultDispatcher struct{}
 func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
 	switch x := resp.(type) {
 	case JSONResponse:
-		rw.Header().Set("Content-Type", "application/json; charset=utf-8")
-		rw.WriteHeader(int(StatusOK))
 		io.WriteString(rw, ")]}',\n") // Break parsing of JavaScript in order to prevent XSSI.
 		return json.NewEncoder(rw).Encode(x.Data)
 	case TemplateResponse:
@@ -42,8 +59,6 @@ func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
 		if !ok {
 			return fmt.Errorf("%T is not a safe template and it cannot be parsed and written", t)
 		}
-		rw.Header().Set("Content-Type", "text/html; charset=utf-8")
-		rw.WriteHeader(int(StatusOK))
 		if len(x.FuncMap) == 0 {
 			return t.Execute(rw, x.Data)
 		}
@@ -52,14 +67,10 @@ func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
 			return err
 		}
 		return cloned.Funcs(x.FuncMap).Execute(rw, x.Data)
-	default:
-		r, ok := x.(safehtml.HTML)
-		if !ok {
-			return fmt.Errorf("%T is not a safe response type and it cannot be written", resp)
-		}
-		rw.Header().Set("Content-Type", "text/html; charset=utf-8")
-		rw.WriteHeader(int(StatusOK))
-		_, err := io.WriteString(rw, r.String())
+	case safehtml.HTML:
+		_, err := io.WriteString(rw, x.String())
 		return err
+	default:
+		return fmt.Errorf("%T is not a safe response type and it cannot be written", resp)
 	}
 }
--- a/safehttp/default_dispatcher.go
+++ b/safehttp/default_dispatcher.go
@@ -26,62 +26,40 @@ import (
 // DefaultDispatcher is responsible for writing safe responses.
 type DefaultDispatcher struct{}
 
-// ContentType returns the Content-Type of a safe response if it's of a
-// safe-type and an error otherwise.
-func (DefaultDispatcher) ContentType(resp Response) (string, error) {
-	switch resp.(type) {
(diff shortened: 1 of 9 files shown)


safehttp/default_dispatcher_test.go
```diff
@@ -51,11 +51,7 @@ func TestDefaultDispatcherValidResponse(t *testing.T) {
 						Parse("<h1>{{ . }}</h1>")))
 				var data interface{}
 				data = "This is an actual heading, though."
-				resp := safehttp.TemplateResponse{
-					Template: &t,
-					Data:     &data,
-				}
-				return d.ExecuteTemplate(w, resp)
+				return d.Write(w, safehttp.NewTemplateResponse(t, data, nil))
```
empijei, line 54: `NewTemplateResponse` will be used *a lot* and it seems quite a long name to use.

What do you think about just `Tpl` or `TplResponse` or `Template` or other ways to shorten it?

safehttp/response.go
```diff
@@ -49,3 +51,24 @@ type TemplateResponse struct {
...
+// NewTemplateResponse creates a TemplateResponse from a Template, its data and
+// the name to function mappings. These will be pre-processed and then written
+// to the http.ResponseWriter, if deemed safe. If the funcMap is non-nil, its
+// elements will override the existing name to function mappings in the
+// template. An attempt to define a new name to function mapping that is not
+// already in the template will result in a panic.
+func NewTemplateResponse(t Template, data interface{}, funcMap map[string]interface{}) TemplateResponse {
+	return TemplateResponse{
+		Template: &t,
+		Data:     &data,
```
empijei, line 71: Why are you taking a reference here?

safehttp/default_dispatcher_test.go
maramihali: Is `TemplateResp` and `JSONResp` okay?

safehttp/response.go
maramihali: Right, that is not actually needed. When we designed the commit phase we've assumed the token and nonce injection is going to be done directly by the XSRF and CSP plugin which is not actually the case.

safehttp/default_dispatcher_test.go
kele: Wasn't the idea to do:

```
func handler(w safehttp.ResponseWriter, req IncomingRequest) {
  // ...
  safehttp.ExecuteTemplate(w, tpl, data)
}
```

safehttp/default_dispatcher.go
```diff
@@ -39,49 +45,32 @@ func (DefaultDispatcher) ContentType(resp Response) (string, error) {
 	}
 }
 
-// Write writes the response to the Response Writer if it's safe HTML. It
-// returns a non-nil error if the response is not safe HTML or if writing the
-// response fails.
+// Write writes the response to the http.ResponseWriter if it's deemed safe.
+// A safe response is either safe HTML, a JSON object or safe HTML template. It
+// will  return a non-nil error if the response is deemed unsafe or if writing
+// operation fails.
```
kele, line 51:
```suggestion
// Write writes the response to the http.ResponseWriter if it's deemed safe.
// It returns a non-nil error if the response is deemed unsafe or if writing
// operation fails.
```

safehttp/response.go
```diff
@@ -41,11 +43,32 @@ type Template interface {
...
+// elements will override the existing name to function mappings in the
+// template. An attempt to define a new name to function mapping that is not
+// already in the template will result in a panic.
+func TemplateResp(t Template, data interface{}, funcMap map[string]interface{}) TemplateResponse {
+	return TemplateResponse{
+		Template: t,
+		Data:     data,
+		FuncMap:  funcMap,
+	}
+}
```
kele, line 74: Let's leave it as it is for now, but I'm not a fan of `FooResp`, especially since it clashes with `FooResponse` and whoever uses autocompletion will get confused.

I suppose we should revisit the idea from: https://github.com/google/go-safeweb/pull/205#discussion_r504652946.

cc @empijei

safehttp/response_writer_test.go
```diff
@@ -67,52 +68,68 @@ func TestResponseWriterWriteTwicePanic(t *testing.T) {
...
+				w.Redirect(ir, "/asdf", safehttp.StatusMovedPermanently)
 			},
 		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			w := safehttp.NewResponseWriter(safehttp.DefaultDispatcher{}, safehttptest.NewTestResponseWriter(&strings.Builder{}), nil)
+			defer func() {
+				if r := recover(); r == nil {
```
kele, line 98: @empijei are you okay with this pattern? I've seen some discussions on another PR.

I don't have strong opinions here.

kele (approved):

safehttp/response_writer_test.go
maramihali: Thanks for pointing this out. :) Got used to checking whether `recover()` returns `nil` instead of the other way around.

safehttp/response.go
maramihali: I think it's not a big change to add the two high-level functions.

safehttp/response_writer.go
```diff
@@ -219,36 +163,21 @@ func (w *ResponseWriter) SetCookie(c *Cookie) error {
...
+	// Dispatcher.Write. Writing should not proceed if ContentType returns an
+	// error.
 	ContentType(resp Response) (string, error)
+
+	// Write writes a Response to the underlying http.ResponseWriter.
+	//
+	// It should return an error if the writing operation fails or if the
+	// provided Response should not be written to the http.ResponseWriter
+	// because it's unsafe.
+	Write(rw http.ResponseWriter, resp Response) error
```
kele, line 182: Note to other reviewers: this has been just rearranged.

kele (approved):

PR #208, opened 2020-10-15
Added a WriteCode method to the safehttp.ResponseWriter

### LINKED ISSUE #151: Add support to set 2XX statuses that are not 200 and 204

maramihali:
Fixes #151 and depends on PR #205 

Allow the user to set a custom status code when writing the response (1XX-2XX) as the `safehttp.ResponseWriter.Write` method will always set the status code to `safehttp.StatusOK`.


maramihali:
Up to date, can be merged.

safehttp/response_writer.go
```diff
@@ -66,8 +66,10 @@ func NotWritten() Result {
...
 
-// Write dispatches the response to the Dispatcher. This will be written to the
-// underlying http.ResponseWriter if the Dispatcher decides it's safe to do so.
+// Write dispatches the response to the Dispatcher after setting the
+// Content-Type header and the status code to safehttp.StatusOK. This is
+// written to the underlying http.ResponseWriter if the Dispatcher decides it's
+// safe to do so.
 //
 // TODO: replace panics with proper error handling when writing the response fails.
 func (w *ResponseWriter) Write(resp Response) Result {
```
kele, line 75: `Write` can be now reimplemented as `WriteCode(resp, StatusOK)`, right?

kele (changes requested):

PR #209, opened 2020-10-19
Allow the user to set a custom status code in the safehttp.ResponseWriter

### LINKED ISSUE #151: Add support to set 2XX statuses that are not 200 and 204

maramihali:
Fixes #151

Previously, there was no way of setting a status code other than `safehttp.NoContent` as the write method in the `safehttp.ResponseWriter` would always set the status code to `safehttp.StatusOK`. Implemented initial functionality and tests to make this possible.

### DIFF
--- a/safehttp/response_writer.go
+++ b/safehttp/response_writer.go
@@ -167,6 +167,9 @@ func (w *ResponseWriter) SetCookie(c *Cookie) error {
 // invalid status code is passed (i.e. not in the range 1XX-5XX) or if the
 // response has already been written.
 //
+// If SetCode was called before NoContent, Redirect or WriteError, the status
+// code set by the latter will be the actual response status.
+//
 // TODO(empijei@, kele@, maramihali@): decide what should be done if the
 // code passed is either 3XX (redirect) or 4XX-5XX (client/server error).
 func (w *ResponseWriter) SetCode(code StatusCode) {
--- a/safehttp/response_writer.go
+++ b/safehttp/response_writer.go
@@ -30,6 +30,7 @@ type ResponseWriter struct {
 
 	// Having this field unexported is essential for security. Otherwise one can
 	// easily overwrite the struct bypassing all our safety guarantees.
+	code    StatusCode
 	header  Header
 	handler handler
 	req     *IncomingRequest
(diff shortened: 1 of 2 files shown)


kele (commented):
What should happen if someone calls `SetCode` and then `NoContent`, `Redirect` or `WriteError`?

maramihali:
Currently, whatever `ResponseWriter.code` was set to by calling `SetCode` would be ignored and the response status will be set properly. That situation would be a bug, but should we actually error/panic in that case? It might be a bit extreme to panic, I don't know whether returning a 500 would be better than ignoring. :-?

kele:
It's not clear that it would always be a bug.

Imagine you call `SetCode()` at the beginning (because you know your handler returns this code), but you encounter an error and want to gracefully recover with `WriteError()`. I don't see a more ergonomic way to do this than what I described, so that shouldn't be treated as a bug.

maramihali:
Right now it is not treated as a bug and the behavior is the one described by you (btw I should write tests to exemplify this, thanks for pointing it out).

kele (approved):

safehttp/response_writer.go
```diff
@@ -157,6 +163,25 @@ func (w *ResponseWriter) SetCookie(c *Cookie) error {
...
+//
+// If SetCode was called before NoContent, Redirect or WriteError, the status
+// code set by the latter will be the actual response status.
+//
+// TODO(empijei@, kele@, maramihali@): decide what should be done if the
+// code passed is either 3XX (redirect) or 4XX-5XX (client/server error).
+func (w *ResponseWriter) SetCode(code StatusCode) {
+	if w.written {
+		panic("ResponseWriter was already written to")
+	}
```
empijei, line 178: I would not panic here. I would actually do nothing in this case. This is not an error worth panicking for, and if we have already indeed written something it's not like we can change the status to a 500, so I would just do nothing in this case and document that if we have already written calling SetCode does nothing.

This would likely either be a non-issue or a small bug, definitely hard to find this as an exploitable vulnerability, so I would not invest too much in making this hard to get wrong.

empijei (changes requested):

empijei (approved):

PR #214, opened 2020-10-21
safesql: fix doc

empijei:
safesql docs didn't display correctly and were incomplete.

### DIFF
--- a/safesql/safesql.go
+++ b/safesql/safesql.go
@@ -16,8 +16,11 @@
 // possible to the original one.
 // The concept of this package is to provide "safe by construction" SQL strings so that code that would accidentally introduce
 // SQL injection vulnerabilities does not compile.
+// If uncheckedconversions and legacyconversions are not used and the sql package is forbidden this package guarantees that only compile-time constants will be
+// interpreted as SQL, thus preventing attacker-controlled strings to be accidentally executed.
 //
-// Examples
+// Migration Examples
+//
 // Code like the following is trivial to migrate from sql to safesql:
 // 	db.Query("SELECT ...", args...)
 // The only change required would be to promote the string literal to a trusted string:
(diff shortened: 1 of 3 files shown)


safesql/sqlwrap.go
```diff
@@ -21,139 +21,276 @@ import (
 	"time"
 )
 
+// Drivers is a tyny wrapper for https://pkg.go.dev/sql#Drivers
```
maramihali, line 24: tiny (here and below)

safesql/sqlwrap.go
```diff
@@ -21,139 +21,276 @@ import (
...
 	return db.db.QueryRowContext(ctx, query.s, args)
 }
 
-// Go 1.15 func (db DB) SetConnMaxIdleTime(d time.Duration) { db.db.SetConnMaxIdleTime(d) }
-func (db DB) SetConnMaxLifetime(d time.Duration) { db.db.SetConnMaxLifetime(d) }
-func (db DB) SetMaxIdleConns(n int)              { db.db.SetMaxIdleConns(n) }
-func (db DB) SetMaxOpenConns(n int)              { db.db.SetMaxOpenConns(n) }
-func (db DB) Stats() DBStats                     { return db.db.Stats() }
+// Only available in Go 1.15+
+// func (db DB) SetConnMaxIdleTime(d time.Duration) { db.db.SetConnMaxIdleTime(d) }
```
maramihali, line 215: why is this commented out?

maramihali (changes requested):

safesql/sqlwrap.go
empijei: Done.

safesql/sqlwrap.go
empijei: Because otherwise this would only compile with super-recent versions of Go.

I moved it to a file that only gets compiled if go 1.15+ is used.

safesql/safesql.go
```diff
@@ -27,12 +30,18 @@
...
+// Note on API documentation.
 //
 // For documentation on methods and types that wrap the standard ones please refer to the stdlib package doc instead, as
 // all the types exported by this package are tiny wrappers around the standard ones and thus follow their behavior.
 // The only relevant difference is that functions accept TrustedSQLString instances instead of plain "strings" and that some
 // dangerous methods have been removed.
+//
+// Explainer
+//
+// This package wraps the sql package and all methods that would normally take a `string` take a TrustedSQLString instead.
```
kele, line 42: I have often got readability comments about not using backticks as they don't render in any special way in GoDoc.

kele (approved):

empijei:
Fixes #158

PR #218, opened 2020-11-12
ResponseWriter is now an interface.

kele:
This is needed for further refactorings of the ResponseWriter where we plan to handle errors due to double-writing, panicking and so on better.


bsiegert (approved):
Can you please comment on why you are making the change?

kele:
@bsiegert Done!

PR #220, opened 2020-11-13
Consolidated request handling and the ResponseWriter implementation to one object

kele:
The state needed to end-to-end handle a single request has been
scattered among many objects with unclear responsibilities. This is an
attempt to simplify the implementation in order to reason about
recursive calls (e.g. done by interceptors writing) and error handling.

Summary of changes:
- the `responseWriter` has been removed and most of it's responsibilities
have been moved to a new struct - `task` (the name is not ideal, TODO);
task implements `safehttp.ResponseWriter`
- the `safehttp/mux.go:handler` type has been removed in favor of
`HandlerConfig`, task and a `processRequest` function; `HandlerConfig` is just
data (no functions) and `processRequest` is responsible for bootstrapping
the task
- the `task` (a replacement for the `responseWriter`) does not hold a reference to
the `Dispatcher` anymore; previously both the `handler` (now removed) and the
`responseWriter` needed a reference to the `Dispatcher`


safehttp/task.go
```diff
@@ -0,0 +1,256 @@
...
+// If the ResponseWriter has already been written to, then this method will panic.
+func (t *Task) NoContent() Result {
+	if t.written {
+		panic("ResponseWriter was already written to")
+	}
+	t.written = true
+	t.commitPhase(NoContentResponse{})
+	t.rw.WriteHeader(int(StatusNoContent))
+	return Result{}
+}
```
empijei, line 124: Question for another PR: should we maybe make `NoContent` a type to pass to `Write` instead?

safehttp/task.go
```diff
@@ -0,0 +1,256 @@
...
+type Task struct {
+	rw  http.ResponseWriter
+	req *IncomingRequest
+
+	cfg HandlerConfig
+
+	code   StatusCode
+	header Header
+
+	written      bool
```
empijei, line 30: We are not distinguishing between actual writes and write calls. This means that when we start writing an error we don't know if the response serialization was already started (and thus writing the error is useless).

safehttp/task.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+}
+
+type HandlerConfig struct {
+	Handler      Handler
+	Dispatcher   Dispatcher
+	Interceptors []ConfiguredInterceptor
+}
+
+func ProcessRequest(cfg HandlerConfig, rw http.ResponseWriter, req *http.Request) {
```
empijei, line 55: Does this need to be exported?

safehttp/task.go
```diff
@@ -0,0 +1,256 @@
...
+// WriteError writes an error response (400-599) according to the provided
+// status code.
+//
+// If the ResponseWriter has already been written to, then this method will panic.
+func (t *Task) WriteError(code StatusCode) Result {
+	// TODO: accept custom error responses that need to go through the dispatcher.
+	if t.writtenError {
+		panic("ResponseWriter.WriteError called twice")
+	}
+	t.written = true
```
empijei, line 135: See comment I left on the `Task` type: at this point we should know if we already started writing or not.

Depending on the answer we might have to skip the `http.Error` call on line 139.

safehttp/task.go
```diff
@@ -0,0 +1,256 @@
...
+	http.Error(t.rw, http.StatusText(int(resp.Code)), int(resp.Code))
+	return Result{}
+}
+
+// Redirect responds with a redirect to the given url, using code as the status code.
+//
+// If the ResponseWriter has already been written to, then this method will panic.
+func (t *Task) Redirect(r *IncomingRequest, url string, code StatusCode) Result {
+	if code < 300 || code >= 400 {
+		panic("wrong method called")
```
empijei, line 148:
```suggestion
		panic(fmt.Sprintf("wrong method called: redirect with status %d", code))
```

safehttp/task.go
```diff
@@ -0,0 +1,256 @@
...
+// on the response. Headers must be set before writing a
+// response (e.g. Write, WriteTemplate).
+func (t *Task) Header() Header {
+	return t.header
+}
+
+// SetCookie adds a Set-Cookie header to the provided ResponseWriter's headers.
+// The provided cookie must have a valid Name, otherwise an error will be
+// returned.
+func (t *Task) SetCookie(c *Cookie) error {
```
empijei, line 168: Maybe we should change this to `AddCookie`?

safehttp/task.go
```diff
@@ -0,0 +1,256 @@
...
+		panic("ResponseWriter was already written to")
+	}
+	t.written = true
+	http.Redirect(t.rw, r.req, url, int(code))
+	return Result{}
+}
+
+// Header returns the collection of headers that will be set
+// on the response. Headers must be set before writing a
+// response (e.g. Write, WriteTemplate).
```
empijei, line 160:
```suggestion
// response.
```

safehttp/task.go
```diff
@@ -0,0 +1,256 @@
...
+
+// commitPhase calls the Commit phases of all the interceptors. This stage will
+// run before a response is written to the ResponseWriter. If a response is
+// written to the ResponseWriter in a Commit phase then the Commit phases of the
+// remaining interceptors won't execute.
+//
+// TODO: BIG WARNING, if ResponseWriter.Write and ResponseWriter.WriteTemplate
+// are called in Commit then this will recurse. CommitResponseWriter was an
+// attempt to prevent this by not giving access to Write and WriteTemplate in
+// the Commit phase.
```
empijei, line 199:
```suggestion
```

safehttp/task.go
```diff
@@ -0,0 +1,256 @@
...
+
+// errorPhase calls the OnError phases of all the interceptors associated with
+// a handler. This stage runs before an error response is written to the
+// ResponseWriter.
+//
+// TODO: BIG WARNING, if an interceptor attempts to write to the ResponseWriter
+// in the OnError phase will result in an irrecoverable error.
+func (t *Task) errorPhase(resp Response) {
+	for i := len(t.cfg.Interceptors) - 1; i >= 0; i-- {
+		t.cfg.Interceptors[i].OnError(t, t.req, resp)
```
empijei, line 214: Should we maybe pass a smaller interface to OnError that doesn't have Write* methods so we make sure Write is not called in this step?

safehttp/task.go
```diff
@@ -0,0 +1,256 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package safehttp
+
+import (
+	"net/http"
+)
+
+type Task struct {
```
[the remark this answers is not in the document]
kele: TBH I don't expect `Task` to be a public API at all.

safehttp/mux.go
```diff
@@ -190,14 +149,44 @@ func (s *ServeMuxConfig) Intercept(i Interceptor) {
...
+	registerHandlers(m, handlers)
+	return &ServeMux{mux: m}
+}
+
+func configureInterceptors(interceptors []Interceptor, cfgs []InterceptorConfig) []ConfiguredInterceptor {
+	var its []ConfiguredInterceptor
+	for _, it := range interceptors {
+		var cfg InterceptorConfig
+		for _, c := range cfgs {
+			if c.Match(it) {
```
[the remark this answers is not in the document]
kele: Yeah, I copied that from the old code cuz I didn't want to change *too much*. Gonna leave a TODO.

safehttp/task.go
kele: Yup, I'm gonna probably add a `dispatched` boolean here. I'll leave a TODO to not cram it into this PR.

safehttp/task.go
kele: Not in this form most likely. We'd rather have a "safe handler factory" in the future.

safehttp/task.go
kele: I considered this, but you'd need the `Dispatcher` to handle it (maybe that's the way to go?).

I would also prefer `Skip` instead of using `NotWritten`.

safehttp/task.go
kele: Probably, I'll leave a TODO.

safehttp/task.go
kele: Yeah, I haven't changed the comments yet at all.

safehttp/mux.go
```diff
@@ -190,14 +149,44 @@ func (s *ServeMuxConfig) Intercept(i Interceptor) {
...
 	}
+	// pattern -> method -> HandlerConfig
+	handlers := map[string]map[string]HandlerConfig{}
 	for _, hr := range s.handlers {
-		m.handle(hr.pattern, hr.method, hr.handler, hr.cfgs...)
+		if handlers[hr.pattern] == nil {
+			handlers[hr.pattern] = map[string]HandlerConfig{}
+		}
+		if _, ok := handlers[hr.pattern][hr.method]; ok {
+			panic(fmt.Sprintf("double registration of (pattern = %q, method = %q)", hr.pattern, hr.method))
```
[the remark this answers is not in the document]
kele: We likely should. Leaving a TODO to not cram too much in a single PR.

safehttp/task.go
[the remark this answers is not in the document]
kele: Not a public API anymore (for now).

We'll need to find a way to create a reasonable fake `ResponseWriter` implementation for testing. The main question is: do we want to provide just a `ResponseWriter` or something that really emulates our framework.

safehttp/task.go
kele: Left a TODO.

safehttp/task.go
kele: I'm not sure. It's not obvious to me what `OnError` should and shouldn't be able to do. How much should it be able to modify the response?

safehttp/task_test.go
```diff
@@ -26,9 +27,16 @@ import (
 	"github.com/google/safehtml"
 )
 
-func TestResponseWriterSetCookie(t *testing.T) {
+func stubResponseWriter(dispatcher safehttp.Dispatcher, rw http.ResponseWriter) safehttp.ResponseWriter {
```
kele, line 30: The `safehttp.NewResponseWriter` API used to return an incomplete `ResponseWriter` (the `ResponseWriter` needs to call interceptors associated with the current handler - this was missing among other things).

We currently settle for a (still broken) implementation, but the next refactoring PR will address the testing issues we have here.

safehttp/task.go
empijei: +1 on Skip and on asking the Dispatcher to handle a NoContent type.

safehttp/task.go
empijei: sometimes we call it after the response has been partially written, there isn't much it can modify at that point, is there?

safehttp/task.go
```diff
@@ -0,0 +1,264 @@
...
+
+// HandlerConfig is the safe HTTP handler configuration, including the
+// dispatcher and interceptors.
+type HandlerConfig struct {
+	Handler      Handler
+	Dispatcher   Dispatcher
+	Interceptors []ConfiguredInterceptor
+}
+
+func processRequest(cfg HandlerConfig, rw http.ResponseWriter, req *http.Request) {
```
empijei, line 65: personal taste: I'd make processRequest a method of HandlerConfig.

safehttp/task.go
```diff
@@ -0,0 +1,264 @@
...
+// limitations under the License.
+
+package safehttp
+
+import (
+	"fmt"
+	"net/http"
+)
+
+// TODO(kele): come up with a better name than this.
```
empijei, line 22: Maybe we can call this a "flight"

empijei (approved):

safehttp/task.go
kele: The current code is broken, sure. But I'd like to somehow gracefully handle panics occurred in the handler. In this case, our framework should (maybe) call `OnError` of all the interceptors, in which case maybe they should be able to do writes?

safehttp/task.go
kele: I did intend to keep `HandlerConfig` to be just data. Maybe I should just make it a value receiver method, though...

safehttp/task.go
kele: Done.

Issue #164, opened 2020-09-14
Fix interaction problems between passing a response to the safehttp.ResponseWriter writing functions that triggers an error and trying to call the markWritten function.


maramihali:
For an `safehttp.ResponseWriter w` with `safehttp.Dispatcher d`, calling `w.markWritten()` sets a boolean `w.written` that ensures a panic is triggered whenever two writing attempts are made to `w`. 

In every write function in the `safehttp.ResponseWriter` (`Write`, `WriteJSON`, `WriteTemplate`, `WriteError`, etc), `w.markWritten()` is called before attempting to retrieve the content type and dispatching the response to the `safehttp.Dispatcher`. 

This is problematic because passing an unsafe response to a writing function of `w` will trigger the following behaviour. First, `w.markWritten()` is called which will set `w.written` to `true`. Then if either `w.d.ContentType(...)` or `w.d.WriteSomething(...)` returns an error, we will panic and that will be recovered by the `ServeMux`. The` ServeMux` will then attempt to call `w.WriteError(..)` with a 500, which attempts to call `w.markWritten()` again, which will panic because `w.written` was already set to true.

This behaviour was signaled by the `TestMuxDefaultDispatcherUnsafeResponses` in integration/mux/mux_test.go that currently needs to be skipped.

We need to design a better approach to handling errors in the `safehttp.ResponseWriter` and write integration tests that thoroughly test the interaction between the `ServeMux`, `ResponseWriter` and `Dispatcher`.

kele:
Fixed by all of the above.

Issue #79, opened 2020-08-14
Add a way to uniformely and safely handle errors


empijei:
The plugin should provide infrastructure for uniform error handling (e.g. to prevent accidental leaks or XSS from error responses).

Error responses should not surface internal details or code errors to the end users.

PR #222, opened 2020-11-30
Interceptors' `Commit` and `OnError` methods can no longer affect the flow by calling write operations

kele:
`Commit` and `OnError` interceptor methods can no longer invoke write calls on a `ResponseWriter`. Instead, they receive a `ResponseHeadersWriter` that allows them to modify the HTTP response headers without affecting the control flow. Similarly, they can change the `Response` that's passed to them.

The only way to affect the control flow is for the interceptor to panic, in which case it is treated as an `InternalServerError`.

This introduces a conceptual change in how the framework operates.

1. Once the server has committed to a response (either via `Write` or `WriteError`), no more write operations are allowed. This simplifies further error handling mechanics. Call cycle detection is not needed anymore (e.g. `Write` -> `WriteError` cannot be introduced by user's code).
2. `Commit` and `OnError` can still modify the response, but has to do so by modifying the headers using `ResponseHeadersWriter` and altering the received `Response` object.

#164, #79

### DIFF
--- a/safehttp/response_writer.go
+++ b/safehttp/response_writer.go
@@ -21,6 +21,8 @@ package safehttp
 //
 // A ResponseWriter may not be used after the Handler.ServeHTTP method has returned.
 type ResponseWriter interface {
+	ResponseHeadersWriter
+
 	// Write writes a safe response.
 	Write(resp Response) Result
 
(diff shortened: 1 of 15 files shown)


kele:
Follow-up PR: #223.

safehttp/response_writer.go
```diff
@@ -38,7 +40,12 @@ type ResponseWriter interface {
 	//
 	// If the ResponseWriter has already been written to, then this method panics.
 	Redirect(r *IncomingRequest, url string, code StatusCode) Result
+}
 
+// ResponseHeadersWriter is used to alter the HTTP response headers.
+//
+// A ResponseHeadersWriter may not be used after the Handler.ServeHTTP method has returned.
+type ResponseHeadersWriter interface {
```
empijei, line 48: I would document that if somethings has already been written this entire type methods set becomes a no-op.

empijei (approved):

safehttp/response_writer.go
kele: Good idea!

kele: Actually, I'm going to think a bit about documenting it better with regards to lifetime. I think that nothing should hold a reference to `ResponseHeadersWriter` after the `Commit` or `OnError` function has been exited.

PR #224, opened 2020-11-30
`flight` now knows whether a response has been dispatched (possibly written to `net/http.ResponseWriter`)

kele:
`flight` doesn't distinguish anymore between `written` and `writtenError`.

Since `Write` -> `WriteError` calls cannot be introduced by users, we no
longer need to distinguish between them (currently only the framework
can call `WriteError` when we're still in a `Write`, but that's going to
change soon as it can only be triggered by a `panic`).

Added a `dispatched` field in order to distinguish between a
write-operation being triggered by the user, and an actual write to the
`net/http.ResponseWriter`. If the latter hasn't happened yet (i.e. we
haven't asked the Go standard library to start writing the response), we
can still handle errors gracefully. After we've written any bytes to the
`net/http.ResponseWriter` (e.g. `200 OK`), there is nothing we can do in
case of an error. This is because the reponse status might have been
already sent over the wire, and the error happened just now.

#164, #79

### DIFF
--- a/safehttp/flight.go
+++ b/safehttp/flight.go
@@ -29,10 +29,7 @@ type flight struct {
 	code   StatusCode
 	header Header
 
-	// TODO(kele): we need to distinguish between calling Write and actually
-	// writing to the net/http.ResponseWriter.
-	written      bool
-	writtenError bool
+	written bool
 }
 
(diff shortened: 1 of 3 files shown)


kele:
Recreating it with new commit message.

Issue #179, opened 2020-09-21
Passing an error message to ResponseWriter.WriteError


maramihali:
Currently, `ResponseWriter.WriteError` supports only reporting the status code not also an error message. We should discuss whether we want to support this.

kele:
We should allow passing safe responses there.

PR #231, opened 2020-12-22
Custom error responses

kele:
Adding custom error responses to the framework.

1. `ErrorResponse` is now an interface type. `StatusCode` implements it.
2. The `Dispatcher` now implement an `Error` method, accepting an `ErrorResponse` interface object.
3. Error responses written through `safehttp.ResponseWriter.WriteError` now go through the `Commit` phase. This means that `Interceptor.OnError` is not needed anymore.
4. The `ServeMuxConfig.Dispatcher` is now a public field (discovered that it's needed when writing the tests).


kele:
Fixes #79.
Fixes #179.

empijei (approved):
Nice! I didn't expect this to come out so clean considering how much is already there, good one!

PR #236, opened 2021-01-08
[Early review] ServeFiles response handler for serving files.

kele:
Please perform an early review (i.e. whether the high level idea is acceptable). If this looks good, I'll create a new PR with proper polish and tests.

### DIFF
--- a/safehttp/file.go
+++ b/safehttp/file.go
@@ -1,0 +1,56 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package safehttp
+
+import (
+	"mime"
+	"net/http"
+	"path/filepath"
+)
+
+// FileResponse is a file response. It is a wrapper around the http.FileServer.
+type FileResponse struct {
+	Root    http.FileSystem
+	Request *IncomingRequest
+}
+
+// ContentType returns the Content-Type based solely on the extension of the
+// filepath. If none is found, the Content-Type is set to
+// "application/octet-stream; charset=binary".
+func (fs *FileResponse) ContentType() string {
+	ct := mime.TypeByExtension(filepath.Ext(fs.Request.URL.Path()))
+	if ct == "" {
+		return "application/octet-stream; charset=binary"
+	}
+	return ct
+}
+
+// Writes writes the response. Write does not override the Content-Type when
+// successfuly serving files. It does override it for redirects and errors.
+//
+// To safely set the Content-Type based on the filepath extension, set the
+// header to the result of ContentType, before calling Write.
+func (fs *FileResponse) Write(rw http.ResponseWriter) {
+	// Note: FileServer does not override the Content-Type when successfuly
+	// serving files. It does override it for redirects and errors.
+	http.FileServer(fs.Root).ServeHTTP(rw, fs.Request.req)
+}
+
+// ServeFiles is a wrapper around the http.FileServer.
+func ServeFiles(root http.FileSystem) Handler {
+	return HandlerFunc(func(rw ResponseWriter, req *IncomingRequest) Result {
+		return rw.Write(FileResponse{root, req})
+	})
+}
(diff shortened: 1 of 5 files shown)


safehttp/file.go
```diff
@@ -0,0 +1,56 @@
...
+
+package safehttp
+
+import (
+	"mime"
+	"net/http"
+	"path/filepath"
+)
+
+// FileResponse is a file response. It is a wrapper around the http.FileServer.
```
empijei, line 23: It's not?

safehttp/file.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Writes writes the response. Write does not override the Content-Type when
+// successfuly serving files. It does override it for redirects and errors.
+//
+// To safely set the Content-Type based on the filepath extension, set the
+// header to the result of ContentType, before calling Write.
+func (fs *FileResponse) Write(rw http.ResponseWriter) {
+	// Note: FileServer does not override the Content-Type when successfuly
+	// serving files. It does override it for redirects and errors.
+	http.FileServer(fs.Root).ServeHTTP(rw, fs.Request.req)
```
empijei, line 48: Do we need to construct the server on every request?

safehttp/file.go
```diff
@@ -0,0 +1,56 @@
...
+		return "application/octet-stream; charset=binary"
+	}
+	return ct
+}
+
+// Writes writes the response. Write does not override the Content-Type when
+// successfuly serving files. It does override it for redirects and errors.
+//
+// To safely set the Content-Type based on the filepath extension, set the
+// header to the result of ContentType, before calling Write.
```
empijei, line 44:
```suggestion
// header to the result of ContentType before calling Write.
```

safehttp/file.go
empijei, line 48: This will set the content-type to `text/plain` and use the stdlib default 404 instead of the framework error page.

I think we should use `fs.Open` ourselves and figure out if this is an error or not if we want to have a consistent error experience.

I think the ServeFiles func should check for this and either call Error or call Write.

safehttp/file.go
```diff
@@ -0,0 +1,56 @@
...
+func (fs *FileResponse) Write(rw http.ResponseWriter) {
+	// Note: FileServer does not override the Content-Type when successfuly
+	// serving files. It does override it for redirects and errors.
+	http.FileServer(fs.Root).ServeHTTP(rw, fs.Request.req)
+}
+
+// ServeFiles is a wrapper around the http.FileServer.
+func ServeFiles(root http.FileSystem) Handler {
+	return HandlerFunc(func(rw ResponseWriter, req *IncomingRequest) Result {
```
empijei, line 53: You can construct the FileServer here and closure it in the handler.

safehttp/file.go
```diff
@@ -0,0 +1,56 @@
...
+// To safely set the Content-Type based on the filepath extension, set the
+// header to the result of ContentType, before calling Write.
+func (fs *FileResponse) Write(rw http.ResponseWriter) {
+	// Note: FileServer does not override the Content-Type when successfuly
+	// serving files. It does override it for redirects and errors.
+	http.FileServer(fs.Root).ServeHTTP(rw, fs.Request.req)
+}
+
+// ServeFiles is a wrapper around the http.FileServer.
+func ServeFiles(root http.FileSystem) Handler {
```
empijei, line 52: Just for consistency with the stdlib: what about calling this FileServer?

safehttp/file.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	return ct
+}
+
+// Writes writes the response. Write does not override the Content-Type when
+// successfuly serving files. It does override it for redirects and errors.
+//
+// To safely set the Content-Type based on the filepath extension, set the
+// header to the result of ContentType, before calling Write.
+func (fs *FileResponse) Write(rw http.ResponseWriter) {
```
empijei, line 45: JSONResponse returns an error, this does not, is it intended?

I think it would be nice to have a "serializable" interface (unexported or exported) to treat all safe responses the same way.

Also, I would call these "Serialize", not "Write". In the stdlib the "Write" method is defined on stuff you can **write to**. These instead write to the argument, so it feels weird. (Yeah, I know WriteHeader does exactly this but bear with me pls).

safehttp/status.go
```diff
@@ -96,6 +96,12 @@ func (c StatusCode) Code() StatusCode {
...
 }
 
+// String returns the status text.
 func (c StatusCode) String() string {
 	return http.StatusText(int(c))
 }
+
+// Write writes the error response.
+func (c StatusCode) Write(rw http.ResponseWriter) {
+	http.Error(rw, c.String(), int(c.Code()))
```
empijei, line 106: Why is this always an error? What if you try to serialize a 202?

safehttp/file.go
empijei: Also, the error reason will be leaked to the client: https://golang.org/src/net/http/fs.go?s=20871:20911#L561

kele:
I currently don't have time to wrap this up. Feel free to fork if desired.

kele:
Closing in favor of https://github.com/google/go-safeweb/pull/262.

PR #241, opened 2021-02-01
Example echo server

### LINKED ISSUE #216: Provide examples on how to use and adopt the framework

kele:
Fixes #216.

### DIFF
--- a/examples/echo/echo.go
+++ b/examples/echo/echo.go
@@ -1,3 +1,17 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


mikue1 (approved):
LGTM. This was very useful to get started.

nit: maybe document the easter_egg parameter in the endpoints list?

I was poking a bit whether I could break the security guarantees, but gave up :)
Initially I thought that defining some script tag inside the template should work for an "XSS", but the CSP seems to block that. As a follow-up I would like to see (or do myself) an example for how users are supposed to write script tags to the output in an "intended"/secure way using safehttp.

mikue1:
Not sure how this works on GitHub, but this should probably mention #216.

kele:
Actually, I want the easter egg to be a real easter egg and stay hidden. :p

With regards to scripts: yes, we should have more complex examples including scripts and state changing requests. Not this PR, though.

PR #243, opened 2021-02-02
Example of mixing net/http and safehttp handlers by nesting muxes.

mikue:


mikue1:
I wanted to see an example where safehttp and net/http handlers are used in the same server.

Maybe there is a better way to achieve this, let me know :)

kele:
I think this example clearly shows that we need a `ServeMux.Handler(...)` method that returns a safe `net/http.Handler`.

kele:
Take a look: #245.

mikue:
Added another PR #248 for easy comparability.

kele:
Closing in favor of https://github.com/google/go-safeweb/pull/248.

PR #247, opened 2021-02-08
Added a `defaults` package.

empijei:
For now it contains just a factory for safe ServeMuxConfigs.

### DIFF
--- a/safehttp/defaults/defaults.go
+++ b/safehttp/defaults/defaults.go
@@ -1,0 +1,49 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Pacakge defaults provides ready to use, safe, pre-configured instances of safehttp types.
+package defaults
+
+import (
+	"github.com/google/go-safeweb/safehttp"
+	"github.com/google/go-safeweb/safehttp/plugins/coop"
+	"github.com/google/go-safeweb/safehttp/plugins/cors"
+	"github.com/google/go-safeweb/safehttp/plugins/csp"
+	"github.com/google/go-safeweb/safehttp/plugins/fetchmetadata"
+	"github.com/google/go-safeweb/safehttp/plugins/hostcheck"
+	"github.com/google/go-safeweb/safehttp/plugins/hsts"
+	"github.com/google/go-safeweb/safehttp/plugins/staticheaders"
+	"github.com/google/go-safeweb/safehttp/plugins/xsrf/xsrfhtml"
+)
+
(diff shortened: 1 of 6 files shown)


safehttp/defaults/defaults.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/google/go-safeweb/safehttp/plugins/csp"
+	"github.com/google/go-safeweb/safehttp/plugins/fetchmetadata"
+	"github.com/google/go-safeweb/safehttp/plugins/hostcheck"
+	"github.com/google/go-safeweb/safehttp/plugins/hsts"
+	"github.com/google/go-safeweb/safehttp/plugins/staticheaders"
+	"github.com/google/go-safeweb/safehttp/plugins/xsrf/xsrfhtml"
+)
+
+// ServeMuxConfig creates a safe and ready to use ServeMuxConfig with all necessary security interceptors installed.
+func ServeMuxConfig(xsrfKey string, hosts, allowedCORSOrigins []string) *safehttp.ServeMuxConfig {
```
kele, line 31: Maybe this could benefit from go/gotip/episodes/41?

kele (approved):

safehttp/defaults/defaults.go
empijei: I thought about this but solution 2 doesn't really fit, and if you adopt solution 1 and you add one more parameter every user of this function will keep working without realizing the API has changed. With that solution there is no way to make parameters mandatory.

kele: Yup, you're right.

In this case, it seems to be more natural to me to have them in the `hosts, xsrf, cors` order. Alternatively, just `hosts, xsrf` and simply not support CORS in the default mux. WDYT?

safehttp/defaults/defaults_test.go
```diff
@@ -0,0 +1,58 @@
...
+			form, err := r.URL.Query()
+			if err != nil {
+				t.Errorf("Cannot parse GET form: %v", err)
+			}
+			b := form.Bool("test", false)
+			if !b {
+				t.Error("test parameter, got false, want true")
+			}
+			w.Write(safehtml.HTMLEscaped(resp))
+			return safehttp.Result{}
```
kele, line 41:
```suggestion
			return w.Write(safehtml.HTMLEscaped(resp))
```

mikue:
Since kele requested review;
A safe-by-default `ServeMuxConfig` should definitely live somewhere. I'm not super convinced of using a "defaults" package vs. e.g. having a `safehttp.NewServeMuxConfig`. Do you have ideas on what else would live in the defaults package eventually?

I also wonder if we should do more checks for the input to this function since it will be the main entry point for end-users. I suppose usage of this function will be in scope for security review anyway though, so maybe we don't need to be so strict (e.g. checking for an empty xsrf key).

mikue (approved):

kele:
> A safe-by-default `ServeMuxConfig` should definitely live somewhere. I'm not super convinced of using a "defaults" package vs. e.g. having a `safehttp.NewServeMuxConfig`. Do you have ideas on what else would live in the defaults package eventually?

Not possible. Dependency cycles.

> I also wonder if we should do more checks for the input to this function since it will be the main entry point for end-users. I suppose usage of this function will be in scope for security review anyway though, so maybe we don't need to be so strict (e.g. checking for an empty xsrf key).

+1 on this

safehttp/defaults/defaults.go
empijei: Agreed!

kele:
Ready to merge. Go ahead @empijei

PR #248, opened 2021-02-10
Example of mixing net/http and safehttp handlers using http.Handler.

### LINKED ISSUE #245: Allow using ServeMux as a factory for net/http.Handlers
Currently to use the `safehttp` package handlers together with a `net/http.ServeMux` (e.g. during migration) one needs to pass the whole `ServeMux` as a `net/http.Handler` (possible because `safehttp.ServeMux` satisfies the interface).

Instead, it'd make sense to add a `func (m *ServeMux) Handler(path string) http.Handler` method, mostly for clarity of intent and easy discoverability.

Note: additional benefit of doing this is avoiding double-routing that otherwise happens when `ServeMux` is used as a handler for `net/http.ServeMux`.

mikue:
Add ServeMux.Handler to extract http.Handlers from a ServeMux.

Fixes #245.

### DIFF
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"net/url"
 )
 
 // The HTTP request methods defined by RFC.
@@ -217,4 +218,18 @@ func (s *ServeMuxConfig) Clone() *ServeMuxConfig {
 	copy(c.handlers, s.handlers)
 	copy(c.interceptors, s.interceptors)
 	return c
+}
+
+// Handler returns the handler to use for the given method and path.
+// TODO: add a host parameter?
+func (m *ServeMux) Handler(method, path string) http.Handler {
+	r := &http.Request{
+		Method: method,
+		Host:   "",
+		URL: &url.URL{
+			Path: path,
+		},
+	}
+	handler, _ := m.mux.Handler(r)
+	return handler
 }
--- a/examples/httpcompose/httpcompose.go
+++ b/examples/httpcompose/httpcompose.go
@@ -1,0 +1,81 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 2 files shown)


mikue:
PTAL. I'm not super happy with this (especially the meh wrapping around net/http mux.Handler func), but didn't clearly see a better way. We always need to call Mux() before any safehttp handlers can be used, right?

Better ideas/approaches very welcome.

safehttp/mux.go
```diff
@@ -204,3 +205,17 @@ func (s *ServeMuxConfig) Clone() *ServeMuxConfig {
...
+// Handler returns the handler to use for the given method and path.
+// TODO: add a host parameter?
+func (m *ServeMux) Handler(method, path string) http.Handler {
+	r := &http.Request{
+		Method: method,
+		Host:   "",
+		URL: &url.URL{
+			Path: path,
+		},
+	}
```
kele, line 218: Yeah, this is a bit hacky.

My recommendation:

```
func (m *ServeMux) RawHandler(path string) http.Handler {
  return m.handlers[path]
}
```

You'd need to add the `handlers` field to `ServeMux`, which shouldn't be a problem.

kele (changes requested):
The implementation of the `Handler` can indeed be less hacky, but I like how much nicer this API is to use. :)

safehttp/mux.go
empijei: There is one problem with that:

Requesting a handler for "/foo/bar" will not return the handler for "/foo/", but just the exact one if one is registered. Otherwise it will return the 404 handler.

So what about using the ServeMux as a factory that takes a safe handler and returns a standard handler wrapped with all the security stuff? We can even rename the `Handler` method to something more clearly stating this is to be used just for transition.

kele: Oh, good catch!

Maybe `RegisteredHandler` would be enough here? We can godoc that this is supposed to be used for transition purposes only too.

empijei: Not a big fan of the name TBH, what about WrappedHandlerDuringTransition or something similarly communicative?

kele:
Closing in favor of #287.

PR #250, opened 2021-02-12
Add a Server type

empijei:
This allows to construct a ready-to use and safe server from a Mux and an address

### DIFF
--- a/safehttp/server.go
+++ b/safehttp/server.go
@@ -17,90 +17,173 @@ package safehttp
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"net"
 	"net/http"
 	"time"
 )
 
 // Server is a safe wrapper for a standard HTTP server.
+// The zero value is safe and ready to use and will apply safe defaults on serving.
+// Changing any of the fields after the server has been started is a no-op.
 type Server struct {
-	srv *http.Server
+	// Addr optionally specifies the TCP address for the server to listen on,
+	// in the form "host:port". If empty, ":http" (port 80) is used.
+	// The service names are defined in RFC 6335 and assigned by IANA.
+	// See net.Dial for details of the address format.
+	Addr string
+
+	// Mux is the ServeMux to use for the current server. A nil Mux is invalid.
+	Mux *ServeMux
+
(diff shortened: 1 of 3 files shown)


safehttp/server.go
```diff
@@ -0,0 +1,106 @@
...
+	"crypto/tls"
+	"net"
+	"net/http"
+	"time"
+)
+
+// Server is a safe wrapper for a standard HTTP server.
+type Server struct {
+	srv *http.Server
+}
```
kele, line 28: The `http.Server` zero value is valid. Is there a way we could preserve this in our package?

safehttp/server.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	if s.MaxHeaderBytes != 0 {
+		srv.MaxHeaderBytes = 10 * 1024
+	}
+	if s.TLSConfig != nil {
+		cfg := s.TLSConfig.Clone()
+		cfg.MinVersion = tls.VersionTLS12
+		cfg.PreferServerCipherSuites = true
+		srv.TLSConfig = cfg
+	}
```
kele, line 118: Please document these overrides.

safehttp/server.go
```diff
@@ -0,0 +1,189 @@
...
+	if s.WriteTimeout != 0 {
+		srv.WriteTimeout = s.WriteTimeout
+	}
+	if s.IdleTimeout != 0 {
+		srv.IdleTimeout = s.IdleTimeout
+	}
+	if s.MaxHeaderBytes != 0 {
+		srv.MaxHeaderBytes = 10 * 1024
+	}
```
kele, line 112: Why is this set this way?

kele (changes requested):

safehttp/server.go
empijei: The default is a very very high and unreasonable value. 10K should be enough as a default. According to [this](http://dev.chromium.org/spdy/spdy-whitepaper) the average is about 2K, so I decided to leave a bit more room for it.

kele: But, if I set `s.MaxHeaderBytes` to `123`, why does it get overridden to `10*1024`?

kele (approved):

Issue #77, opened 2020-08-14
Add framing protection


empijei:
Framing protection is done in two ways:

X-Frame-Options header, which should by default set to "sameorigin"
A frame-ancestors CSP directive set to "self". This part should somehow interact with the CSP plugin to set the additional directive.

Configuration should only be possible via some gated API that relaxes the framing protection on some endpoints. Note that on those endpoints XFO will be set to "allow" and CSP frame-ancestors to a specified list of allowed origins.

empijei:
Update: We currently have half of this implemented in CSP, but we lack the XFO part and the orchestration between the two.

empijei:
I think the best way to do this is to just add a fallback to XFO to the CSP plugin framing policy.

The idea would be to set a XFO header that matches the CSP behavior: `DENY` if the frame ancestors is `none`, `SAMEORIGIN` if self, and not set it otherwise.

PR #253, opened 2021-03-12
Adds support for generating X-Frame-Options headers based on framing policy.

mtrzos:
First step to allow customising the framing policy - described in Issue #77.

- [ ] Tests pass

### DIFF
--- a/safehttp/plugins/csp/csp.go
+++ b/safehttp/plugins/csp/csp.go
@@ -19,6 +19,8 @@
 //  - A strict nonce based CSP
 //  - A framing policy which sets frame-ancestors to 'self'
 //  - A Trusted Types policy which makes usage of dangerous web API functions secure by default
+//
+// In case a framing policy is present, a relevant X-Frame-Options will be added as a fallback option.
 package csp
 
 import (
@@ -167,6 +169,12 @@ func (f FramingPolicy) Serialize(nonce string) string {
 	return b.String()
 }
 
+// toXFO returns a value of the X-Frame-Options header that reflects settings of the policy.
+func (f FramingPolicy) toXFO() string {
+	// TODO: allow relaxation on specific endpoints according to #77.
+	return "SAMEORIGIN"
+}
+
 // TrustedTypesPolicy policy can be used to create a new CSP which makes
 // dangerous web API functions secure by default.
 //
@@ -219,8 +227,8 @@ func Default(reportURI string) Interceptor {
 	}
 }
 
-// Before claims and sets the Content-Security-Policy header and the
-// Content-Security-Policy-Report-Only header.
+// Before claims and sets the Content-Security-Policy and Content-Security-Policy-Report-Only.
+// It also sets X-Frame-Options header as a fallback option for cases when CSP is not supported.
 func (it Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest, _ safehttp.InterceptorConfig) safehttp.Result {
 	nonce := generateNonce()
 	r.SetContext(context.WithValue(r.Context(), ctxKey{}, nonce))
@@ -233,15 +241,32 @@ func (it Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequ
 	for _, p := range it.ReportOnly {
 		reportCSPs = append(reportCSPs, p.Serialize(nonce))
 	}
+	// X-Frame-Options works only in enforce mode.
+	var XFO = []string{getXFO(&it.Enforce)}
 
 	h := w.Header()
 	setCSP := h.Claim("Content-Security-Policy")
 	setCSPReportOnly := h.Claim("Content-Security-Policy-Report-Only")
+	setXFO := h.Claim("X-Frame-Options")
 
 	setCSP(CSPs)
 	setCSPReportOnly(reportCSPs)
+	setXFO(XFO)
 
 	return safehttp.NotWritten()
+}
+
+// getXFO returns X-Frame-Options header value that reflects configuration of the FramingPolicy.
+// If no FramingPolicy exists, defaults to DENY.
+func getXFO(p *[]Policy) string {
+	for _, p := range *p {
+		f, ok := p.(FramingPolicy)
+		if !ok {
+			continue
+		}
+		return f.toXFO()
+	}
+	return "DENY"
 }
 
 // Commit adds the nonce to the safehttp.TemplateResponse which is going to be
--- a/safehttp/plugins/csp/csp.go
+++ b/safehttp/plugins/csp/csp.go
@@ -225,7 +225,7 @@ type Interceptor struct {
 var _ safehttp.Interceptor = Interceptor{}
 
 // Default creates a new CSP interceptor with a strict nonce-based policy, a framing policy
-// and a TrustedType policy. All in enforcement mode.
+// and a TrustedTypes policy. All in enforcement mode.
 func Default(reportURI string) Interceptor {
 	return Interceptor{
 		Enforce: []Policy{
--- a/safehttp/plugins/csp/csp.go
+++ b/safehttp/plugins/csp/csp.go
@@ -167,7 +167,7 @@ func (f FramingPolicy) Serialize(nonce string) string {
 	return b.String()
 }
 
-// TrustedTypesPolicy policy can be used to create a new CSP policy which makes
+// TrustedTypesPolicy policy can be used to create a new CSP which makes
 // dangerous web API functions secure by default.
 //
 // See https://web.dev/trusted-types for more info.
--- a/safehttp/plugins/csp/csp.go
+++ b/safehttp/plugins/csp/csp.go
@@ -15,9 +15,10 @@
 // Package csp provides a safehttp.Interceptor which applies Content-Security Policies
 // to responses.
 //
-// Two default policies are provided:
+// Three default policies are provided:
 //  - A strict nonce based CSP
 //  - A framing policy which sets frame-ancestors to 'self'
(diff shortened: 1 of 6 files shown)


safehttp/plugins/csp/csp.go
```diff
@@ -219,8 +227,8 @@ func Default(reportURI string) Interceptor {
 	}
 }
 
-// Before claims and sets the Content-Security-Policy header and the
-// Content-Security-Policy-Report-Only header.
+// Before claims and sets the Content-Security-Policy and Content-Security-Policy-Report-Only.
+// It also sets X-Frame-Options header as a fallback option for cases when CSP is not supported.
```
empijei, line 231:
```suggestion
// It also sets X-Frame-Options header as a fallback option for user agents that don't support CSP frame-ancestors.
```

safehttp/plugins/csp/csp.go
```diff
@@ -233,17 +241,34 @@ func (it Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequ
...
 	setCSP(CSPs)
 	setCSPReportOnly(reportCSPs)
+	setXFO(XFO)
 
 	return safehttp.NotWritten()
 }
 
+// getXFO returns X-Frame-Options header value that reflects configuration of the FramingPolicy.
+// If no FramingPolicy exists, defaults to SAMEORIGIN.
+func getXFO(p *[]Policy) string {
```
empijei, line 261: Is it just me or this will always return "SAMEORIGIN"?

safehttp/plugins/csp/csp.go
```diff
@@ -233,17 +241,34 @@ func (it Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequ
...
+	var XFO = []string{getXFO(&it.Enforce)}
 
 	h := w.Header()
 	setCSP := h.Claim("Content-Security-Policy")
 	setCSPReportOnly := h.Claim("Content-Security-Policy-Report-Only")
+	setXFO := h.Claim("X-Frame-Options")
 
 	setCSP(CSPs)
 	setCSPReportOnly(reportCSPs)
+	setXFO(XFO)
```
empijei, line 254: Won't this always set "SAMEORIGIN"?

Shouldn't we *not* set XFO if a framing policy is not set?

safehttp/plugins/csp/csp_test.go
```diff
@@ -126,11 +126,13 @@ func TestBefore(t *testing.T) {
 		wantEnforcePolicy    []string
 		wantReportOnlyPolicy []string
 		wantNonce            string
+		wantXFO              []string
```
empijei, line 129: Can you please add a test to check that we don't set XFO if no framing policy is specified and if the framing policy is RO?

safehttp/plugins/csp/csp.go
```diff
@@ -167,6 +169,12 @@ func (f FramingPolicy) Serialize(nonce string) string {
 	return b.String()
 }
 
+// toXFO returns a value of the X-Frame-Options header that reflects settings of the policy.
+func (f FramingPolicy) toXFO() string {
```
kele, line 173: nit: I'd name it "XFrameOptions".

The "to" or "get" prefixes in Go are quite uncommon.

safehttp/plugins/csp/csp.go
kele, line 261: Slices are passed by reference, no need for it to be a pointer.

safehttp/plugins/csp/csp.go
```diff
@@ -233,17 +241,34 @@ func (it Interceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequ
...
 	setCSPReportOnly(reportCSPs)
+	setXFO(XFO)
 
 	return safehttp.NotWritten()
 }
 
+// getXFO returns X-Frame-Options header value that reflects configuration of the FramingPolicy.
+// If no FramingPolicy exists, defaults to SAMEORIGIN.
+func getXFO(p *[]Policy) string {
+	for _, p := range *p {
```
kele, line 262: Don't shadow the variable.

safehttp/plugins/csp/csp.go
kele, line 261: It returns the first XFrameOptions, right? Consider naming it something along the lines of `firstXFrameOptions`

"get" is usually treated as a filler word in Go function names as it doesn't add anything to the meaning of it. It's almost always better to drop it altogether or replace it with a more specific word (e.g. `fetch` if it involves an RPC).

safehttp/plugins/csp/csp_test.go
kele, line 129: Please add a test for multiple framing policies provided. I'd suspect that you need to alter what your PR does here to cover this case.

kele (changes requested):

safehttp/plugins/csp/csp_test.go
empijei: We will probably have to look up the strictest one if we have more than one setting, which we currently don't have, so I guess it is fine for now.

kele: WDYM?

What if you create an interceptor with multiple framing policies?

kele (changes requested):

safehttp/plugins/csp/csp_test.go
empijei: image
Why would anyone do that? And even if they did... all policies would have the same strictness value.

kele: Alright, I think I understand now what this PR is trying to do. See my PR-level comment.

kele (changes requested):
It took me a while to understand what this PR is trying to do...

My current understanding is that: you try to add a `X-Frame-Options = SAMEORIGIN header` if the `csp` plugin has `FramingPolicy` used.

A few problems:
1. `FramingPolicy` should probably be called `FramingPolicySelf` instead, otherwise it's confusing.
2. I'm not sure that the csp plugin should own the XFO header. 
3. Any implementation that sets XFO based on CSP config should not rely on the current limitation of the csp plugin that allows setting only a specific `self` framing policy. I can easily see adding support for framing directives to the csp plugin in the nearest future, and it'll likely break any fragile integration with XFO.

My recommendation: either write an XFO implementation that is future-proof, or start with adding framing controls support to the csp plugin first and then add support for XFO.

mtrzos:
> It took me a while to understand what this PR is trying to do...
> 
> My current understanding is that: you try to add a `X-Frame-Options = SAMEORIGIN header` if the `csp` plugin has `FramingPolicy` used.
That is correct.
> A few problems:
> 
> 1. `FramingPolicy` should probably be called `FramingPolicySelf` instead, otherwise it's confusing.
I'm not sure what you mean by that. Could you explain?
> 2. I'm not sure that the csp plugin should own the XFO header.
Are there any downsides?
> 3. Any implementation that sets XFO based on CSP config should not rely on the current limitation of the csp plugin that allows setting only a specific `self` framing policy. I can easily see adding support for framing directives to the csp plugin in the nearest future, and it'll likely break any fragile integration with XFO.
Adding support for other framing directives and integration with XFO would be the next step.
> My recommendation: either write an XFO implementation that is future-proof, or start with adding framing controls support to the csp plugin first and then add support for XFO.
In that case I think I'll keep working on this PR and will add full support at once.

kele:
> > It took me a while to understand what this PR is trying to do...
> > My current understanding is that: you try to add a `X-Frame-Options = SAMEORIGIN header` if the `csp` plugin has `FramingPolicy` used.
> 
> That is correct.
> 
> > A few problems:
> > 
> > 1. `FramingPolicy` should probably be called `FramingPolicySelf` instead, otherwise it's confusing.
> 
> I'm not sure what you mean by that. Could you explain?

CSP can have meaningful framing policies that actually allow iframing from different origins. Currently the comment says that `SAMEORIGIN` is the only thing and it seems that Rob's comments regarding handling multiples policies hinge on that. If we want to keep that contract for this specific struct, it should be renamed to `FramingPolicySelf` to be absolutely clear that it's just `SAMEORIGIN`. Or maybe `FramingPolicySameOrigin`.

If we want to keep `FramingPolicy` and extend it's capabilities in the nearest future, we need to future-proof the XFO code to handle that (e.g. gracefully handle multiple `FramingPolicy` objects).

> > 1. I'm not sure that the csp plugin should own the XFO header.
> 
> Are there any downsides?

Yes. The Content Security Policy plugin should own the CSP headers only, conceptually. XFO is a separate thing from CSP and it's really surprising to see it here. @empijei can fill you in on the plugin orchestration idea probably.

> > 1. Any implementation that sets XFO based on CSP config should not rely on the current limitation of the csp plugin that allows setting only a specific `self` framing policy. I can easily see adding support for framing directives to the csp plugin in the nearest future, and it'll likely break any fragile integration with XFO.
> 
> Adding support for other framing directives and integration with XFO would be the next step.

:+1:

> 
> > My recommendation: either write an XFO implementation that is future-proof, or start with adding framing controls support to the csp plugin first and then add support for XFO.
> 
> In that case I think I'll keep working on this PR and will add full support at once.

Thank you! Please double check with @empijei first though before you put significant work into this.

safehttp/plugins/csp/csp.go
mtrzos: Done.

safehttp/plugins/csp/csp.go
mtrzos: Done.

mtrzos:
I'm abandoning this PR and will continue working on frame ancestors extension.

PR #257, opened 2021-03-31
Extend FramingPolicy to include allowed sources.

mtrzos:
This is the first step to add a full support for framing protection (described in #77)

> It's a good idea to open an issue first for discussion.

- [ ] Tests pass

### DIFF
--- a/safehttp/plugins/csp/csp.go
+++ b/safehttp/plugins/csp/csp.go
@@ -186,8 +186,8 @@ func report(reportURI string) string {
 	if reportURI != "" {
 		b.WriteString("report-uri ")
 		b.WriteString(reportURI)
+		b.WriteString("; ")
 	}
-	b.WriteString("; ")
 
 	return b.String()
 }
--- a/safehttp/plugins/csp/csp.go
+++ b/safehttp/plugins/csp/csp.go
@@ -147,6 +147,9 @@ func (s StrictPolicy) Serialize(nonce string) string {
 //
 // TODO: allow relaxation on specific endpoints according to #77.
 type FramingPolicy struct {
+	// Specifies valid parents that may embed a page using: <frame>, <iframe>, <object>, <embed>, or <applet>.
+	// More details: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors#sources
+	Sources []string
 	// ReportURI controls the report-uri directive. If ReportUri is empty, no report-uri
 	// directive will be set.
 	ReportURI string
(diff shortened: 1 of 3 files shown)


safehttp/plugins/csp/csp.go
```diff
@@ -159,6 +162,15 @@ func (f FramingPolicy) Serialize(nonce string) string {
 	var b strings.Builder
 	b.WriteString("frame-ancestors 'self'")
```
kele, line 163: At this point I'd carve out the `frame-ancestors` and `report-uri` directives to separate functions and join them with `; `.

e.g.:

```
ancestors := frameAncestors(f.Sources)
rep := report(f.ReportURI)

return strings.Join([]string{ancestors, rep}, "; ")
```

Alternatively:

```
var b strings.Builder
f.writeAncestors(&b)
f.writeReportURI(&b)
return b.String()
```

(the `writeFoo` methods always append `; `)

kele (changes requested):

safehttp/plugins/csp/csp.go
mtrzos: I chose the first option. It adds a semicolon to the default policy but it's still valid.

safehttp/plugins/csp/csp.go
```diff
@@ -159,9 +162,30 @@ func (f FramingPolicy) Serialize(nonce string) string {
...
 
-	if f.ReportURI != "" {
-		b.WriteString("; report-uri ")
-		b.WriteString(f.ReportURI)
+	ancestors := frameAncestors(f.Sources)
+	rep := report(f.ReportURI)
+	b.WriteString(strings.Join([]string{ancestors, rep}, "; "))
```
kele, line 167: I'm a bit nitpicky here, but please bear with me.

As it is, you have four lines:
1. Use `WriteString` to write a part of the `frame-ancestors` directive.
2. Create a string that contains a part of the `frame-ancestors` directive.
3. Create a string that contains the full `report-uri` directive.
4. Use `WriteString` to write the two parts above, joined by `"; "`.

Wouldn't it be cleaner to merge 1 and 2 inside `frameAncestors`?

This way the code would end up being:
```
var b strings.Builder
b.WriteString(frameAncestors(f.Sources)) // writes at least "frame-ancestors 'self'; " (note the space)
b.WriteString(report(f.ReportURI)) // writes a "; " delimited directive or an empty string
return strings.TrimSpace(b.String())
```

kele (changes requested):

safehttp/plugins/csp/csp.go
mtrzos: That's true. It's simpler and avoids extra unneeded semicolon in the case report-uri is not present.

safehttp/plugins/csp/csp.go
```diff
@@ -157,11 +160,31 @@ type FramingPolicy struct {
...
+	}
+
+	return b.String()
+}
+
+func report(reportURI string) string {
+	var b strings.Builder
+
+	if reportURI != "" {
 		b.WriteString("; report-uri ")
```
kele, line 186: This is brittle.

There is no reason for `Serialize` to not write `report()` first and then `frameAncestors`, but putting the semicolon at the beginning breaks this.

kele (approved):
One last nit. Otherwise LGTM.

safehttp/plugins/csp/csp.go
mtrzos: Ok, I made it order agnostic.

kele:
Thank you! :)

Issue #233, opened 2021-01-05
Add path traversal attack tests


kele:
https://owasp.org/www-community/attacks/Path_Traversal

Check the behavior of the Go's standard library file server functions, like:
- [`http.FileServer`](https://godoc.org/net/http#FileServer)
- [`http.ServeFile`](https://godoc.org/net/http#ServeFile)

empijei:
I would also add that the `Filename` field of the multipart `FileHeader` type *might* be a relative directory path or a fully qualified path unlike the name seems to suggest.

This means that users might inadvertently allow attackers to upload file outside of their intended directory if no path stripping is performed.

PR #262, opened 2021-04-13
FileServer implementation.

### LINKED ISSUE #187: Simple file server (no dispatcher, no interceptors)
Applications frequently serve static files as a response (`.js`, `.css`, images, etc) and, for these, developers  implement unexported function to detect and set the `Content-Type`, based on the files' extension, omitting the `charset` parameter often. We might want to do this properly in the `safehttp.DefaultDispatcher`.

kele:
In our threat model we assume that the directory passed to the
`FileServer` contains exclusively fully-trusted content.

- a `FileServerResponse` is generated when the underlying `http.FileServer`
calls `WriteHeader(200)`; the response is written as any other safe
response (e.g. goes through interceptors)
- the `DefaultDispatcher` sets the Content-Type according to what
`FileServerResponse` sets, but doesn't actually write the content due to
the nature of `http.FileServer` (many direct calls to
`http.ResponseWriter.Write([]byte)`)
- all errors coming from `http.FileServer` are converted to
`safehttp.StatusNotFound` in order to prevent information leaks

Run `go mod tidy` too.

Fixes #187.


kele:
@googlebot I fixed it.

safehttp/fileserver.go
```diff
@@ -0,0 +1,127 @@
...
+
+import (
+	"errors"
+	"net/http"
+)
+
+// FileServer returns a handler that serves HTTP requests with the contents of
+// the file system rooted at root.
+func FileServer(root string) Handler {
+	fileServer := http.FileServer(http.Dir(root))
```
mtrzos, line 25: Just out of curiosity - the default http.FileServer is safe from path traversal?

safehttp/fileserver.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return HandlerFunc(func(rw ResponseWriter, req *IncomingRequest) Result {
+		fileServerRW := &fileServerResponseWriter{flight: rw.(*flight), header: http.Header{}}
+		fileServer.ServeHTTP(fileServerRW, req.req)
+		return fileServerRW.result
+	})
+}
+
+type fileServerResponseWriter struct {
+	flight *flight
```
mtrzos, line 35: does that mean response in-flight? Is this some broader known convention in Go?

kele: It's the part of the framework that needs to be run per response in-flight, yes. It's not a broadly known convention in Go.

safehttp/fileserver.go
kele: AFAIK yes, but we want extra tests to confirm: https://github.com/google/go-safeweb/issues/233.

safehttp/fileserver.go
```diff
@@ -0,0 +1,127 @@
...
+	return fileServerRW.header
+}
+
+func (fileServerRW *fileServerResponseWriter) Write(b []byte) (int, error) {
+	if !fileServerRW.committed {
+		fileServerRW.WriteHeader(int(StatusOK))
+	}
+
+	if fileServerRW.errored {
+		// Let the framework handle the error
```
empijei, line 62: Ultranit
```suggestion
		// Let the framework handle the error.
```

safehttp/fileserver.go
```diff
@@ -0,0 +1,127 @@
...
+		return 0, errors.New("discarded")
+	}
+	return fileServerRW.flight.rw.Write(b)
+}
+
+func (fileServerRW *fileServerResponseWriter) WriteHeader(statusCode int) {
```
empijei, line 68: fileServerRW is a bit long as a name for a receiver. Consider, maybe, fsrw?

safehttp/fileserver.go
```diff
@@ -0,0 +1,127 @@
...
+	// private, to not allow modifications
+	contentType string
+}
+
+// ContentType is the Content-Type of the response.
+func (resp FileServerResponse) ContentType() string {
+	return resp.contentType
+}
+
+func contentType(h http.Header) string {
```
empijei, line 121: This is a very generic name for something that is visible by the entire package.

Maybe consider defining this inside the only function that uses it or giving it a more specific name?

mtrzos (approved):

safehttp/fileserver.go
kele: Removed altogether.

PR #267, opened 2021-04-13
Sanitise file names in multipart forms.

### LINKED ISSUE #264: Make multipart forms resilient to path traversal attacks
See https://tools.ietf.org/html/rfc7578:

> If a "filename" parameter is supplied, the requirements of
> Section 2.3 of [RFC2183] for the "receiving MUA" (i.e., the receiving
> Mail User Agent) apply to receivers of multipart/form-data as well:
> do not use the file name blindly, check and possibly change to match
> local file system conventions if applicable, and do not use directory
> path information that may be present.

mtrzos:
Fixes #264

### DIFF
--- a/safehttp/form.go
+++ b/safehttp/form.go
@@ -17,6 +17,7 @@ package safehttp
 import (
 	"fmt"
 	"mime/multipart"
+	"path/filepath"
 	"strconv"
 )
 
@@ -212,7 +213,17 @@ func (f *MultipartForm) File(param string) []*multipart.FileHeader {
 	if !ok {
 		return nil
 	}
+	sanitise(fh)
 	return fh
+}
+
+// sanitise removes trailing path separators from all file names.
+// This is to ensure that uploaded files are not stored outside of the
+// designated directory.
+func sanitise(fhs []*multipart.FileHeader) {
+	for _, fh := range fhs {
+		fh.Filename = filepath.Base(fh.Filename)
+	}
 }
 
 // RemoveFiles removes any temporary files associated with a Form and returns
--- a/safehttp/form.go
+++ b/safehttp/form.go
@@ -205,6 +205,28 @@ type MultipartForm struct {
 	mf *multipart.Form
 }
 
+// newMultipartForm constructs a new MultipartForm with sanitized values.
+func newMulipartForm(mf *multipart.Form) *MultipartForm {
+	return &MultipartForm{
+		Form: Form{
+			values: mf.Value,
+		},
+		mf: sanitizeFilenames(mf),
+	}
+}
+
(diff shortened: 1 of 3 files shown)


safehttp/form.go
```diff
@@ -212,9 +213,19 @@ func (f *MultipartForm) File(param string) []*multipart.FileHeader {
...
 		return nil
 	}
+	sanitise(fh)
 	return fh
 }
 
+// sanitise removes trailing path separators from all file names.
+// This is to ensure that uploaded files are not stored outside of the
+// designated directory.
+func sanitise(fhs []*multipart.FileHeader) {
```
kele, line 223: The name is a bit too generic to be in the safehttp package.

safehttp/form.go
```diff
@@ -212,9 +213,19 @@ func (f *MultipartForm) File(param string) []*multipart.FileHeader {
 	if !ok {
 		return nil
 	}
+	sanitise(fh)
```
kele, line 216: Shouldn't this be done once when the form is parsed and returned by IncomingRequest?

kele (changes requested):

safehttp/form.go
mtrzos: Channged to sanitiseFilenames

safehttp/form.go
mtrzos: moved sanitization to constructor

safehttp/form_test.go
```diff
@@ -546,8 +544,25 @@ func TestMultipartFormValidFileAndVals(t *testing.T) {
...
+	fh := &multipart.FileHeader{Filename: "../tmp/myfile.txt"}
+	mf := &multipart.Form{
+		File: map[string][]*multipart.FileHeader{
+			"foo": {fh},
+		}}
+	f := newMulipartForm(mf)
+
+	filename := f.File("foo")[0].Filename
+	if want, got := "myfile.txt", filename; want != got {
+		t.Errorf(`f.File("foo").Filename: got %s, want %s`, got, want)
```
kele, line 557: Use %q for strings.

safehttp/form_test.go
```diff
@@ -546,8 +544,25 @@ func TestMultipartFormValidFileAndVals(t *testing.T) {
...
+	}
+	if err := f.Err(); err != nil {
+		t.Errorf(`f.Err(): got err %v`, err)
```
kele, line 560: , want nil

kele (approved):

safehttp/form_test.go
mtrzos: Done

safehttp/form_test.go
mtrzos: Done

PR #271, opened 2021-04-14
Add support for go1.16 embed.FS.

empijei:
Note that I bumped the go.mod file to use go 1.16 but made sure this
still works with go1.15.

This module will now conditionally build with different features
depending on the compiler version used. Expressing the 1.16 version in
go.mod only means this module *may* use newer features, but it can still
be backward compatible.

This is gonna be useful for any users that want to bundle static resources
in their web servers and doesn't break users on previous versions of Go.

### DIFF
--- a/safehttp/fileserver_1_16_test.go
+++ b/safehttp/fileserver_1_16_test.go
@@ -1,0 +1,80 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


safehttp/fileserver_1_16_test.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	mb := &safehttp.ServeMuxConfig{}
+	mb.Handle("/", safehttp.MethodGet, safehttp.FileServerEmbed(testEmbeddedFS))
+	m := mb.Mux()
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var b strings.Builder
+			rr := safehttptest.NewTestResponseWriter(&b)
```
kele, line 74: Rebase and use httptest.ResponseRecorder.

kele (approved):

safehttp/fileserver_1_16_test.go
empijei: Done

PR #274, opened 2021-04-14
Speed-up requesttesting test harness

### LINKED ISSUE #59: The test harness for testing net/http parsing is very slow
Running the tests in `tests/headers`, `tests/formparams` and `tests/queryparams` currently takes 30+ seconds. This is **very** slow. They all use the testing harness in `internal/requesttesting/test_harness.go`. @empijei suspects that this might be because something is not closed correctly which causes the test harness to have to wait for a timeout to finish on each test.

This needs to be investigated further. The cause of the slowdown needs to identified and fixed.

kele:
Fixes #59.

For some reason `server.Shutdown()` waits for quite a bit (~0.5sec) even though everything is closed. Using a forceful `server.Close()` instead.

### DIFF
--- a/internal/requesttesting/test_harness.go
+++ b/internal/requesttesting/test_harness.go
@@ -33,7 +33,9 @@ type AssertHandler struct {
 
 func (h *AssertHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	h.callback(r)
-	io.WriteString(w, "Hello world!")
+	if _, err := io.WriteString(w, "Hello world!"); err != nil {
+		panic(err)
+	}
 }
 
 // FakeListener creates a custom listener that avoids opening a socket in order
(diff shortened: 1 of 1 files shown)


internal/requesttesting/test_harness.go
```diff
@@ -129,18 +138,5 @@ func MakeRequest(ctx context.Context, req []byte, callback func(*http.Request))
...
-	// goroutines to shut everything down. When server.Shutdown is called it
-	// checks if there are any idle, non-closed connections left and waits for
-	// them to close. It performs this check every 0.5 s. But our connection
-	// above takes a lot less time than 0.5 s to close. Therefore, if we just
-	// give up execution for a moment to let the other goroutines close
-	// everything before calling server.Shutdown, we don't get punished by
-	// having to wait 0.5s every time we try to shut down.
-	runtime.Gosched()
-	return resp[:n], server.Shutdown(ctx)
+	return resp[:n], server.Close() /* Forceful shutdown. We don't want to delay anything. */
```
empijei, line 141: Does this work? Like... does this not cause issues?

kele: Seems to work.

`MakeRequest` is fully self-contained, so even if something stays alive for a bit, it shouldn't be an issue in testing.

empijei (approved):

Issue #230, opened 2020-12-22
When mux responds with Method Not Allowed, make the response go through WriteError


kele:
https://github.com/google/go-safeweb/blob/407fc4e34c05ad9fea347a5388fb855045791551/safehttp/mux.go#L78

Quick fix: create a dummy handler with `w.WriteError(safehttp.StatusMethodNotAllowed)`.

kele:
Fixed by #278.

PR #277, opened 2021-04-15
Simplified the Dispatcher

### LINKED ISSUE #239: flight.WriteError calling http.ResponseWriter.WriteHeader causes DefaultDispatcher to superfluously call WriteHeader the second time
A log is printed by the `net/http` package about calling `WriteHeader` twice:

```
2021/01/29 10:59:36 http: superfluous response.WriteHeader call from github.com/google/go-safeweb/safehttp.DefaultDispatcher.Error (default_dispatcher.go:92)
```

This is becase `flight.WriteError` calls `WriteHeader`, and then the `DefaultDispatcher` calls `http.Error` which also calls `WriteHeader`.

**Should `flight.WriteError` call `WriteHeader`, or should this be the responsibility of the `Dispatcher`?**

kele:
Simplified the Dispatcher.

New Dispatcher responsibilities:
- setting the `Content-Type` response header
- setting the HTTP response status code if different than `200 OK`
- `Dispatcher.Error()` now needs to accept all implementations of `ErrorResponse`: this unblocks #230

Other changes:
- `flight` no longer sets the `Content-Type` nor calls `http.ResponseWriter.WriteHeader` in `flight.Write` or `flight.WriteError`
- removed `ResponseWriter.SetCode`

 TODO: unify response handling for `flight.NoContent` and `flight.Redirect`.

Fixes #239.


safehttp/default_dispatcher.go
```diff
@@ -62,13 +41,15 @@ func (DefaultDispatcher) ContentType(resp Response) (string, error) {
 func (DefaultDispatcher) Write(rw http.ResponseWriter, resp Response) error {
```
empijei, line 41: Please add something to the doc that says it sets CT. I know it's also in the interface definition but I'd like to make sure this is smth that can't just be overlooked by readers.

empijei (approved):

safehttp/default_dispatcher.go
kele: Done.

PR #279, opened 2021-04-15
Refactor handler generation in mux.go and expose handler factory.

mikue:
Refactor handler generation in mux.go

Besides other things, this enables to expose a public http.Handler factory API, which allows to use net/http and (converted) safehttp handlers in a http mux more easily.

This is the 3rd iteration attempting to address #245 (last one was #248 which has some relevant review comments).

I still consider this a bit drafty, but it seems to work (example works like before, and tests pass :)).

### DIFF
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -72,37 +72,36 @@ type ServeMux struct {
 }
 
 func registerHandlers(mux *http.ServeMux, handlers map[string]map[string]*handlerConfig) {
-	for pattern, methodToHandlerConfigMap := range handlers {
-		httpHandler := generateHttpHandlerFunc(methodToHandlerConfigMap)
-		mux.HandleFunc(pattern, httpHandler)
+	for pattern, methodToHandlerConfig := range handlers {
+		httpHandler := stdHandler(methodToHandlerConfig)
+		mux.Handle(pattern, httpHandler)
 	}
 }
 
-// generateHttpHandlerFunc converts a handlerConfig to a classic http.HandlerFunc.
-// We need to have this map because http.Handle handles requests for any HTTP request method per default,
+// stdHandler create a http.Handler using a handlerConfig.
+// The handlerConfig map is needed because http.Handle handles requests for any HTTP request method,
 // which we want to avoid in safehttp.
-func generateHttpHandlerFunc(methodToHandlerConfig map[string]*handlerConfig) http.HandlerFunc {
-	return func(w http.ResponseWriter, r *http.Request) {
+func stdHandler(methodToHandlerConfig map[string]*handlerConfig) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		cfg, ok := methodToHandlerConfig[r.Method]
 		if !ok {
 			http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
 			return
 		}
 		processRequest(cfg, w, r)
-	}
+	})
 }
 
-// HttpHandlerForTransition generates a classic http.Handler using its ServeMuxConfig
-// (i.e. already installed Interceptors etc).
+// StdHandler creates an http.Handler using the ServeMuxConfig
+// (i.e. using already installed Interceptors etc).
 // This is intended to be used during transition from using the http package to safehttp.
-func (s *ServeMuxConfig) HttpHandlerForTransition(method string, handler Handler, cfgs ...InterceptorConfig) http.Handler {
+func (s *ServeMuxConfig) StdHandler(method string, handler Handler, cfgs ...InterceptorConfig) http.Handler {
 	cfg := s.handlerRegistrationToHandlerConfig(&handlerRegistration{
-		pattern: "",
 		method:  method,
 		handler: handler,
 		cfgs:    cfgs,
 	})
-	return generateHttpHandlerFunc(map[string]*handlerConfig{
+	return stdHandler(map[string]*handlerConfig{
 		method: cfg,
 	})
 }
--- a/safehttp/mux.go
+++ b/safehttp/mux.go
@@ -71,18 +71,40 @@ type ServeMux struct {
 	mux *http.ServeMux
 }
 
-func registerHandlers(mux *http.ServeMux, handlers map[string]map[string]handlerConfig) {
-	for pattern, handlersPerMethod := range handlers {
-		handlersPerMethod := handlersPerMethod
-		mux.HandleFunc(pattern, func(w http.ResponseWriter, r *http.Request) {
-			cfg, ok := handlersPerMethod[r.Method]
-			if !ok {
-				http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
-				return
-			}
-			processRequest(cfg, w, r)
-		})
+func registerHandlers(mux *http.ServeMux, handlers map[string]map[string]*handlerConfig) {
+	for pattern, methodToHandlerConfigMap := range handlers {
+		httpHandler := generateHttpHandlerFunc(methodToHandlerConfigMap)
+		mux.HandleFunc(pattern, httpHandler)
 	}
+}
+
+// generateHttpHandlerFunc converts a handlerConfig to a classic http.HandlerFunc.
+// We need to have this map because http.Handle handles requests for any HTTP request method per default,
+// which we want to avoid in safehttp.
+func generateHttpHandlerFunc(methodToHandlerConfig map[string]*handlerConfig) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		cfg, ok := methodToHandlerConfig[r.Method]
+		if !ok {
+			http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
+			return
+		}
+		processRequest(cfg, w, r)
+	}
+}
+
+// HttpHandlerForTransition generates a classic http.Handler using its ServeMuxConfig
+// (i.e. already installed Interceptors etc).
+// This is intended to be used during transition from using the http package to safehttp.
+func (s *ServeMuxConfig) HttpHandlerForTransition(method string, handler Handler, cfgs ...InterceptorConfig) http.Handler {
+	cfg := s.handlerRegistrationToHandlerConfig(&handlerRegistration{
+		pattern: "",
+		method:  method,
+		handler: handler,
+		cfgs:    cfgs,
+	})
+	return generateHttpHandlerFunc(map[string]*handlerConfig{
+		method: cfg,
+	})
 }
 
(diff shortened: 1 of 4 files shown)


safehttp/mux.go
```diff
@@ -71,20 +71,42 @@ type ServeMux struct {
...
-			cfg, ok := handlersPerMethod[r.Method]
-			if !ok {
-				http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
-				return
-			}
-			processRequest(cfg, w, r)
-		})
+func registerHandlers(mux *http.ServeMux, handlers map[string]map[string]*handlerConfig) {
+	for pattern, methodToHandlerConfigMap := range handlers {
+		httpHandler := generateHttpHandlerFunc(methodToHandlerConfigMap)
```
kele, line 76: Please revert this rename: go/go-style/decisions#variable-names.

> Omit types and type-like words from most variable names.

safehttp/mux.go
```diff
@@ -71,20 +71,42 @@ type ServeMux struct {
...
+		mux.HandleFunc(pattern, httpHandler)
+	}
+}
+
+// generateHttpHandlerFunc converts a handlerConfig to a classic http.HandlerFunc.
+// We need to have this map because http.Handle handles requests for any HTTP request method per default,
+// which we want to avoid in safehttp.
+func generateHttpHandlerFunc(methodToHandlerConfig map[string]*handlerConfig) http.HandlerFunc {
```
kele, line 84: "generate" is a filler that has little meaning here.

Maybe `stdHandler` is a better name here? Also, can it return a `http.Handler` instead of `http.HandlerFunc`?

safehttp/mux.go
```diff
@@ -71,20 +71,42 @@ type ServeMux struct {
...
+		cfg, ok := methodToHandlerConfig[r.Method]
+		if !ok {
+			http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
+			return
+		}
+		processRequest(cfg, w, r)
 	}
 }
 
+// HttpHandlerForTransition generates a classic http.Handler using its ServeMuxConfig
```
kele, line 95: `StandardHandler` or `StdHandler` instead?

kele, line 95: `// Foo creates an http.Handler...`

"Generate" is not really the right word here.

safehttp/mux.go
```diff
@@ -71,20 +71,42 @@ type ServeMux struct {
...
+// (i.e. already installed Interceptors etc).
+// This is intended to be used during transition from using the http package to safehttp.
+func (s *ServeMuxConfig) HttpHandlerForTransition(method string, handler Handler, cfgs ...InterceptorConfig) http.Handler {
```
kele, line 98: As discussed: it'd make more sense to have a separate `HandlerFactory` or similar object which the `ServeMuxConfig` uses under the hood (or can be constructed from).

It feels weird to have to create a `ServeMuxConfig` and not use it to create a `Mux`.

safehttp/mux.go
```diff
@@ -167,27 +195,26 @@ func (s *ServeMuxConfig) Mux() *ServeMux {
...
-			panic(fmt.Sprintf("double registration of (pattern = %q, method = %q)", hr.pattern, hr.method))
+			handlers[hr.pattern] = make(map[string]*handlerConfig)
 		}
-		handlers[hr.pattern][hr.method] =
-			handlerConfig{
-				Dispatcher:   s.dispatcher,
-				Handler:      hr.handler,
-				Interceptors: configureInterceptors(s.interceptors, hr.cfgs),
-			}
+		handlers[hr.pattern][hr.method] = s.handlerRegistrationToHandlerConfig(&hr)
```
kele, line 203: Isn't `&hr` a bug here? Aren't you taking a pointer to the same local variable?

safehttp/mux.go
```diff
@@ -71,20 +71,42 @@ type ServeMux struct {
...
+		}
+		processRequest(cfg, w, r)
 	}
 }
 
+// HttpHandlerForTransition generates a classic http.Handler using its ServeMuxConfig
+// (i.e. already installed Interceptors etc).
+// This is intended to be used during transition from using the http package to safehttp.
+func (s *ServeMuxConfig) HttpHandlerForTransition(method string, handler Handler, cfgs ...InterceptorConfig) http.Handler {
+	cfg := s.handlerRegistrationToHandlerConfig(&handlerRegistration{
```
kele, line 99: Wouldn't it be easier to just create `handlerConfig` here?

safehttp/mux.go
```diff
@@ -71,20 +71,42 @@ type ServeMux struct {
...
+			http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
+			return
+		}
+		processRequest(cfg, w, r)
 	}
 }
 
+// HttpHandlerForTransition generates a classic http.Handler using its ServeMuxConfig
+// (i.e. already installed Interceptors etc).
+// This is intended to be used during transition from using the http package to safehttp.
```
kele, line 97: How are you going to support the case of two handlers registered for the same path but different methods?

kele (changes requested):

safehttp/mux.go
mikue: Is this comment related to `methodToHandlerConfigMap`? Happy to drop "Map", but when I was refactoring, I had to pause every time I came across `handlersPerMethod` to understand what is going on. What makes things more difficult here is that it is part of `map[string]map[string]handlerConfig` and one needs to know which of these strings refers to what.
If you aren't familiar with http's handler registration (one handler takes care of all methods by default; safehttp avoids that), it isn't obvious why we need a "pattern -> method -> config" map at all.

Would you be okay with `handlerConfigPerMethod` (and prefer it over `methodToHandlerConfig`)?

kele: Yup.

safehttp/mux.go
mikue: Done.

safehttp/mux.go
mikue: My thinking was that a user of this framework would look for something that gives a net.http response, and by having http in the name, autocomplete would make it more discoverable.

The verbose name comes from https://github.com/google/go-safeweb/pull/248#discussion_r595045280

Anyway, going with Std everywhere for this iteration.

mikue: Done.

safehttp/mux.go
mikue: Good question, I was hoping for a test to catch that, but I will take another look.

safehttp/mux.go
mikue: Very good catch... I'll think about it.

mikue: Seems like we will need to expose the private stdHandler here and and make users pass a method -> handler map or so.
Or live with the fact that we are doing things different from `http` and ask users of the transition API to write method-aware handlers (like they have to do in standard http) -- this seems very bad, as long as we want to stick with a per-method handler registration API in safehttp.

mikue: PTAL, I pushed a new version that should work (in an arguably ugly way). I'm not convinced yet that letting users handle this (like in example above) isn't the best way.

safehttp/mux.go
mikue: Agreed, will work on this once the other pieces in this PR look okay.

safehttp/mux.go
```diff
@@ -71,18 +71,48 @@ type ServeMux struct {
...
+type HandlerRegistration struct {
+	Method  string
+	Handler Handler
+	Cfgs    []InterceptorConfig
+}
+
+// StdHandler creates an http.Handler using the ServeMuxConfig
+// (i.e. using already installed Interceptors etc).
+// This is intended to be used during transition from using the http package to safehttp.
+func (s *ServeMuxConfig) StdHandler(hr []*HandlerRegistration) http.Handler {
```
kele, line 105: `hrs ...*HandlerRegistration`

kele: Of course, the `ServeMuxConfig` should not have the `StdHandler` method and it should belong to a different type instead. But my understanding is that you'll proceed with this once you get an LGTM on the idea. :)

safehttp/mux.go
kele: In this case it isn't because you are not storing `&hr` anywhere, but it is definitely error-prone.

safehttp/mux.go
```diff
@@ -167,27 +203,26 @@ func (s *ServeMuxConfig) Mux() *ServeMux {
...
-				Interceptors: configureInterceptors(s.interceptors, hr.cfgs),
-			}
+		handlers[hr.pattern][hr.method] = s.handlerRegistrationToHandlerConfig(&hr)
 	}
 	m := http.NewServeMux()
 	registerHandlers(m, handlers)
 	return &ServeMux{mux: m}
 }
 
+func (s *ServeMuxConfig) handlerRegistrationToHandlerConfig(hr *handlerRegistration) *handlerConfig {
```
kele, line 218: nit: `configureHandler`?

safehttp/mux.go
```diff
@@ -71,18 +71,48 @@ type ServeMux struct {
...
+			http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
+			return
+		}
+		processRequest(cfg, w, r)
+	})
+}
+
+// TODO rename...
+type HandlerRegistration struct {
+	Method  string
```
kele, line 97: I'd treat `Method == ""` as "all methods" and do a sanity check inside the `StandardHandler` call to make sure methods don't overlap, the length of the `hr` slice is 1 if `Method == ""` is used and so on.

kele (changes requested):

kele:
I'd like to discuss what should let us get the `http.Handler`s out of a `safehttp.Handler`. Currently it's either `ServeMuxConfig`, `ServeMux` or a new `HandlerFactory`.

I think there are some pros and cons there, and although `HandlerFactory` is the most detached (and therefore appealing), I'm wondering whether `ServeMux.Handler` isn't the best approach here. Especially since it'd mimic https://pkg.go.dev/net/http#ServeMux.Handler (and we can even use it underneath, simplifying the whole PR).

safehttp/mux.go
mikue: That's the plan :)

safehttp/mux.go
empijei: I'd honestly have the `ServeMuxConfig` being used to instantiate the temporary factory for transition.

You will **eventually** use it to construct the Mux, but not during the transition phase. I think it is the right way to create a user journey towards the full framework adoption.

mikue: Yeah, I also think that whatever we use here should be the thing that users use for both muxes and handlers (transition), which makes it a pretty important decision

kele:
I'll try to wrap this up over the next few days.

kele:
Closing in favor of #287.

PR #281, opened 2021-04-16
Complete overhaul of the safehttptest package

### LINKED ISSUE #152: Status code in safehttptest.ResponseRecorder should be 0 by default instead of 200 so that interceptors that don't set it can be tested properly.
Currently the `ResponseRecorder` in `safehttptest` sets the status code to 200 if no status code has been set yet. This should be changed to 0 so that interceptors that don't set a status code and leave it up to the handler can be tested properly.

kele:
Complete overhaul of the safehttptest package.

- introduced FakeResponseWriter and FakeDispatcher for testing purposes
- removed safehttp.ResponseRecorder
- removed the dreaded safehttp.DeprecatedNewResponseWriter (!)
- removed change detectors in Interceptor tests (e.g. when headers were
checked for ones that are introduced directly by the Dispatcher
implementation, not the Interceptor itself)

Fixes #152 (obsolete).


safehttp/plugins/cors/cors_test.go
```diff
@@ -283,24 +271,18 @@ func TestRequestDisallowedContentTypes(t *testing.T) {
...
 
-			if want := safehttp.StatusUnsupportedMediaType; rr.Status() != want {
-				t.Errorf("rr.Status() got: %v want: %v", rr.Status(), want)
-			}
-			wantHeaders := map[string][]string{
-				"Content-Type":           {"text/plain; charset=utf-8"},
-				"X-Content-Type-Options": {"nosniff"},
+			if want := safehttp.StatusUnsupportedMediaType; rr.Code != int(want) {
+				t.Errorf("rr.Code got: %v want: %v", rr.Code, want)
 			}
```
kele, line 281: I've dropped the headers/body checks in many cases where it felt like a change detector.

Most importantly, headers and bodies in these tests were coming from the `Dispatcher` implementation, not the `Interceptor` that was under test, so they don't really help testing the interceptor's logic.

safehttp/plugins/hsts/hsts_test.go
```diff
@@ -24,120 +24,122 @@ import (
 	"github.com/google/go-safeweb/safehttp/safehttptest"
 )
 
-func TestHSTS(t *testing.T) {
+func TestHSTSReject(t *testing.T) {
```
kele, line 27: I've split these to not test for the `Body` always because it's up to the `Dispatcher`, not really the `Interceptor`.

safehttp/safehttptest/recorder.go
```diff
@@ -21,34 +21,96 @@ import (
...
-
-// NewResponseRecorder creates a ResponseRecorder.
-func NewResponseRecorder() *ResponseRecorder {
-	rw := httptest.NewRecorder()
-	return &ResponseRecorder{
-		rw:             rw,
-		ResponseWriter: safehttp.DeprecatedNewResponseWriter(rw, nil),
+// FakeResponseWriter creates a fake safehttp.ResponseWriter implementation.
+//
+// It does no error checking nor runs interceptors. Only the
```
mtrzos, line 26: Only the...  - is this intended?

mtrzos (approved):

safehttp/safehttptest/recorder.go
kele: Good catch, fixed!

Issue #270, opened 2021-04-14
A linter for disallowing imports and function calls


kele:
This is a subset of #116.

## Requirements

*   banning imports (with an allowlist for exceptions)
    *   e.g. don't allow calling `github.com/google/go-safeweb/safesql/legacyconversions`. Note that safesql must be allowed to import “sql”.
*   banning function calls (with an allowlist for exceptions)
    *   e.g. don't allow calling `safehttp.NewServeMuxConfig`
*   allowlists live in a configuration file (JSON)

## Suggested Usage

```
# sql.json
{
  "imports": {
    "github.com/google/go-safeweb/safesql/legacyconversions": {
      "message": "To use legacyconversions contact security@mycompany.org",
      "exemptions": [
        {
          "justification": "...",
          "allowed": "mycompany.com/my/project/vetted",
        },
        {
          "justification": "...",
          "allowed": "mycompany.com/my/subdirs/vetted/...",
        },
      ],
    }
  },
}
```

```
# mux.json
{
  "calls": {
    "github.com/google/go-safeweb/safehttp.NewServeMuxConfig": [
      "message": "Use mycompany.com/my/project/vetted.NewServeMuxConfig instead.",
      "exemptions": [
        {
          "justification": "...",
          "allowed": "mycompany.com/my/project/vetted",
        },
        {
          "justification": "...",
          "allowed": "mycompany.com/my/subdirs/vetted/...",
        },
      ],
    }
  },
}
```

(those two files could be merged, there is no requirement to have one directive per file)

```
$ bancheck -c sql.json -c mux.json ./…
```

## Useful resources

The Go analysis framework: https://pkg.go.dev/golang.org/x/tools/go/analysis. A talk about it: https://www.youtube.com/watch?v=pOl82CwQp50

The tool should likely live under `cmd/bancheck`.

PR #283, opened 2021-04-20
Add representation of a bancheck config.

mtrzos:
First step in implementing a full linter (#270). Adds representation of configuration file contents with banned APIs.


cmd/bancheck/config/config.go
```diff
@@ -0,0 +1,112 @@
...
+type BannedApis map[string][]BannedApi
+
+type Exemption struct {
+	Justification string `json:"justification"`
+	AllowedDir    string `json:"allowedDir"`
+}
+
+// ReadBannedImports reads banned imports from all config files
+// and concatenates them into one object.
+func ReadBannedImports(files []string) (BannedApis, error) {
```
mtrzos, line 39: I decided to split Read() into ReadBannedImports() and ReadBannedFunctions(). I think that makes more sense because the client will only be interested in one or the other, not both.

kele: > the client will only be interested in one or the other, not both.

Why? Who is the client in your mind?

mtrzos: The [analyzers](https://pkg.go.dev/golang.org/x/tools/go/analysis#hdr-Analyzer), one analyzer will check for banned imports and the other for banned function calls.

kele: I think this is one of the benefits of splitting config-reading logic from actual usage of these -- that you *don't* need to split reading "different parts of a config".

I'd leave it as it was before and do a single `ReadConfig` func and apply all my comments from the other PR.

As it is right now:
- this is basically a copy paste, including all the boilerplate related to reading a file and so on
- you read a file multiple times; not a big issue per se (we don't need to worry about performance)

cmd/bancheck/config/config.go
```diff
@@ -0,0 +1,112 @@
...
+
+		for _, fn := range config.Functions {
+			fns[fn.Name] = append(fns[fn.Name], fn)
+		}
+	}
+
+	return fns, nil
+}
+
+// config represents contents of a configuration file passed to the linter.
```
kele, line 75: I am being really nitpicky here, but I think it'll benefit your comments in the long term.

The fact that it is a file is irrelevant (and may change in the future).

cmd/bancheck/config/config.go
```diff
@@ -0,0 +1,112 @@
...
+	f, err := readFile(filename)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	return decodeCfg(f)
+}
+
+func readFile(filename string) (*os.File, error) {
```
kele, line 91: It doesn't read it, it just opens it.

cmd/bancheck/config/config.go
```diff
@@ -0,0 +1,112 @@
...
+	if info.IsDir() {
+		return nil, errors.New("file is a directory")
+	}
+
+	return os.Open(filename)
+}
+
+func decodeCfg(f *os.File) (*config, error) {
+	var cfg config
+	d := json.NewDecoder(f)
```
kele, line 105: nit: probably don't need the intermediate variable

cmd/bancheck/config/config.go
kele: BTW. We only require comments for public APIs (lowercase letter start). You're welcome to add comments to private symbols, but I think in this case it's pretty self explanatory.

kele (changes requested):

cmd/bancheck/config/config.go
mtrzos: Makes sense. I changed it back.

cmd/bancheck/config/config.go
mtrzos: Done

cmd/bancheck/config/config.go
mtrzos: Removed

cmd/bancheck/config/config.go
```diff
@@ -0,0 +1,112 @@
...
+
+type BannedApi struct {
+	Name       string      `json:"name"` // fully qualified identifier name
+	Msg        string      `json:"msg"`  // additional information e.g. rationale for banning
+	Exemptions []Exemption `json:"exemptions"`
+}
+
+// BannedApis is a map of identifier names to a list of all corresponding
+// BannedApi entries with additional information.
+type BannedApis map[string][]BannedApi
```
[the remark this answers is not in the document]
mtrzos: dropped the extra type.

cmd/bancheck/config/config.go
mtrzos: That means `BannedApi` and `Exemption` should also get comments?

kele: Yes. https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

cmd/bancheck/config/config.go
```diff
@@ -0,0 +1,87 @@
...
+	Msg        string      `json:"msg"`  // additional information e.g. rationale for banning
+	Exemptions []Exemption `json:"exemptions"`
+}
+
+type Exemption struct {
+	Justification string `json:"justification"`
+	AllowedDir    string `json:"allowedDir"`
+}
+
+// ReadConfigs reads banned Apis from all files.
```
kele, line 39: nit: APIs

cmd/bancheck/config/config.go
```diff
@@ -0,0 +1,87 @@
...
+	"os"
+)
+
+// Config represents a configuration passed to the linter.
+type Config struct {
+	Imports   []BannedApi `json:"imports"`
+	Functions []BannedApi `json:"functions"`
+}
+
+type BannedApi struct {
```
kele, line 28: nit: `BannedAPI` https://github.com/golang/go/wiki/CodeReviewComments#initialisms

cmd/bancheck/config/config_test.go
```diff
@@ -0,0 +1,285 @@
...
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"golang.org/x/tools/go/analysis/analysistest"
+)
+
+func TestReadConfigs(t *testing.T) {
```
kele, line 25: Add one tests that has both imports and functions in one file.

kele (approved):

cmd/bancheck/config/config_test.go
mtrzos: Done.

cmd/bancheck/config/config.go
mtrzos: Done

cmd/bancheck/config/config.go
mtrzos: Done

cmd/bancheck/config/config.go
mtrzos: Added comments

PR #284, opened 2021-05-05
Fix brittle test

dsnet:
It is considered poor practice to compare exact error strings,
which causes this test to break on Go1.17.
Fix the test to be more tolerant and work on both Go1.16 and Go1.17.
This test would othewise be broken by https://golang.org/cl/308952.


dsnet:
\cc @kele @empijei

empijei (approved):

bsiegert (approved):

kele:
Thank you. :)

PR #287, opened 2021-05-06
A way to adopt Go Safe Web one handler at a time

### LINKED ISSUE #245: Allow using ServeMux as a factory for net/http.Handlers

kele:
Added a `RegisteredHandler` function that retrieves the registered HTTP handler.

This allows for using `safehttp.ServeMux` alongside `net/http.Handler`.

Fixes #245.


empijei (commented):
I like this. It's clear and straightforward enough :)

safehttp/doc.go
```diff
@@ -123,6 +123,25 @@
...
 // TODO
 //
+// Incremental Adoption
+//
+// In order to migrate your service using http.Handlers to the safehttp package,
+// we recommend you start doing that one endpoint at a time. Use the
+// ServeMux.RegisteredHandler method to do this. See:
+//
+//  safeMuxConfig := /* configured ServeMuxConfig, including interceptors */
+//  safeMuxConfig.Handle("/bar", safehttp.MethodGET, barSafeHandler)
```
mikue, line 133: Maybe add a line
`safeMuxConfig.Handle("/bar", safehttp.MethodPOST, barSafeHandler)`
to showcase method "overloading". It wasn't clear to me immediately that this would work as expected when we call `RegisteredHandler` later.

mikue:
I like this as well.

While we are back to calling `Mux()` which we wanted to try avoiding in #279, I didn't find a nice way to do so without having an ugly public API like in the example from #279.
I dislike the duplicate `Handle()` calls, but I think the trade-off is okay in the end, because otherwise the public API gets more convoluted, so LGTM.

safehttp/doc.go
kele: Done.

empijei (approved):

PR #301, opened 2021-05-21
fetchmetadata: Update to new API.

### LINKED ISSUE #300: fetchmetadata: switch to configs for exceptions
The fetchmetadata plugin dates back to when we didn't have configs.

This means that exceptions for the protection must currently be passed on construction instead of handler registration.

We should switch to the new mechanism.

empijei:
Fixes #300

### DIFF
--- a/safehttp/plugins/fetchmetadata/fetchmetadata.go
+++ b/safehttp/plugins/fetchmetadata/fetchmetadata.go
@@ -23,7 +23,7 @@ import (
 
 // RequestLogger is used for logging Fetch Metadata policy violations;
 type RequestLogger interface {
-	Log(ir *safehttp.IncomingRequest)
+	Log(ir *safehttp.IncomingRequest, navigationIsolation bool)
 }
 
 var (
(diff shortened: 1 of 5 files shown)


safehttp/plugins/fetchmetadata/fetchmetadata.go
```diff
@@ -138,37 +126,56 @@ func (p *Plugin) SetEnforce() {
...
+	return w.WriteError(safehttp.StatusForbidden)
 }
 
 // Commit is a no-op, required to satisfy the safehttp.Interceptor interface.
-func (p *Plugin) Commit(w safehttp.ResponseHeadersWriter, r *safehttp.IncomingRequest, resp safehttp.Response, _ safehttp.InterceptorConfig) {
+func (p *Interceptor) Commit(w safehttp.ResponseHeadersWriter, r *safehttp.IncomingRequest, resp safehttp.Response, _ safehttp.InterceptorConfig) {
+}
+
+// Disable is a configuration that disables Fetch Metadata protection and switches
+// it to report-only.
```
kele, line 172:
```suggestion
// Disable disables Fetch Metadata protection and switches it to report-only.
```

kele (approved):

Issue #307, opened 2021-05-27
Support Interceptor Wrapping


empijei:
For various reasons (e.g. experiments) it might be useful to wrap interceptors.

The problem is that once an interceptor gets wrapped, it will not be matched to its configurations anymore.

I propose to add the following feature:

When a Mux is constructed we match configs with their interceptor, but we also check if the the interceptor has a

`Unwrap() Interceptor`

method on it.

If so, we recursively check configs against the inner interceptors.

kele:
Is my understanding correct that we don't need anything extra to support this case? Maybe just a GoDoc example?

PR #309, opened 2021-05-31
Add support for wrapped interceptor, detect config clashes.

### LINKED ISSUE #302: Make sure that no two given configs apply to the same interceptor
This is a TODO expressed here:
https://github.com/google/go-safeweb/blob/d29986caf7ba2bd243f3ca698413cff9fa91bc92/safehttp/mux.go#L225-L233

We should either panic there, or require configs to be somehow mergeable.

empijei:
Fixes #302, #307


kele (approved):
Could you add a comment to the methods that accept `InterceptorConfig` explaining that at most one config can match?

PR #321, opened 2021-06-14
Interceptors can now carry state throughout the request processing.

### LINKED ISSUE #314: SetContext violates the context.Context contract
In all APIs that rely on context there is no way for a callee to change the context of the caller.

Our IncomingRequest.SetContext method allows to do that. We should instead mimic the stdlib way and clone the request.

Alternatively we could explicitly pass context.Context as first parameter, but it would be quite a bit leap from the stlib api.

kele:
Introduces a `safehttp.FlightValues()` function that lets interceptors and handlers refer to a shared state through the `safehttp.IncomingRequest.Context()`.

See `safehttp/flightvalues_test.go` for usage.

Fixes #314.

### DIFF
--- a/safehttp/flightvalues_test.go
+++ b/safehttp/flightvalues_test.go
@@ -1,0 +1,89 @@
+package safehttp_test
+
+import (
+	"context"
+	"fmt"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-safeweb/safehttp"
+	"github.com/google/safehtml"
+)
+
+type safeHeadersInterceptor struct{}
+
+func (ip *safeHeadersInterceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest, cfg safehttp.InterceptorConfig) safehttp.Result {
+	// We claim the header here in order to protect it from being tampered. The
+	// only way to set it is through a helper method exposed by this package. It
+	// only allows for setting safe values.
+	h := w.Header().Claim("Super-Safe-Header")
+	safehttp.FlightValues(r.Context()).Put(safeHeaderKey{}, h)
+	return safehttp.NotWritten()
+}
+
+func (ip *safeHeadersInterceptor) Commit(w safehttp.ResponseHeadersWriter, r *safehttp.IncomingRequest, resp safehttp.Response, cfg safehttp.InterceptorConfig) {
+}
+
+func (ip *safeHeadersInterceptor) Match(_ safehttp.InterceptorConfig) bool {
+	// This interceptor does not offer any configuration options.
+	return false
+}
+
+type safeHeaderKey struct{}
+
+func SetHeaderSafely(ctx context.Context, level int) {
+	var value string
+	switch level {
+	case 0:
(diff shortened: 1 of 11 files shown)


safehttp/flightvalues_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/google/safehtml"
+)
+
+type safeHeadersInterceptor struct{}
+
+func (ip *safeHeadersInterceptor) Before(w safehttp.ResponseWriter, r *safehttp.IncomingRequest, cfg safehttp.InterceptorConfig) safehttp.Result {
+	// We claim the header here in order to protect it from being tampered. The
+	// only way to set it is through a helper method exposed by this package. It
+	// only allows for setting safe values.
+	h := w.Header().Claim("Super-Safe-Header")
```
stapelberg, line 34: h → setter

safehttp/flightvalues_test.go
```diff
@@ -0,0 +1,103 @@
...
+	case 0:
+		value = "Safe"
+	case 1:
+		value = "VerySafe"
+	case 2:
+		value = "VeryVerySafe"
+	default:
+		value = "Safe"
+	}
+	safehttp.FlightValues(ctx).Get(safeHeaderKey{}).(func([]string))([]string{value})
```
stapelberg, line 61: This can be broken apart to be more descriptive/readable:

setter := safehttp.FlightValues(ctx).Get(safeHeaderKey{}).(func([]string))
setter([]string{value})

safehttp/incoming_request.go
```diff
@@ -47,11 +50,15 @@ func NewIncomingRequest(req *http.Request) *IncomingRequest {
 	if req == nil {
 		return nil
 	}
+	req = req.WithContext(context.WithValue(req.Context(),
+		flightValuesCtxKey{}, &flightValues{m: map[interface{}]interface{}{}}))
```
stapelberg, line 54: make(map[interface{}]interface{}) might be a little easier to read here than a literal

stapelberg (approved):

safehttp/incoming_request.go
```diff
@@ -47,11 +47,15 @@ func NewIncomingRequest(req *http.Request) *IncomingRequest {
...
```
chressie, line 54: You store a pointer to flightValues, but the methods Put and Get are defined on the non-pointer type. Is that on purpose?

chressie (approved):
Nice approach! :)

safehttp/flightvalues_test.go
kele: Done in 0ce62ea.

safehttp/flightvalues_test.go
kele: Done in 0ce62ea.

safehttp/incoming_request.go
kele: Done in 0ce62ea.

safehttp/incoming_request.go
kele: Changed to a value. Thanks!

Done in 0ce62ea.

PR #326, opened 2021-07-09
Add a plaintext type that is safe by default

### LINKED ISSUE #246: Add a plaintext type that is safe by default
It is currently possible to use the default dispatcher to write JSON, templated HTML and static HTML.

It would be beneficial to add one more type to allow text/plain strings to be written.

kallydev:
Fixes #246. 

> It's a good idea to open an issue first for discussion.

This is the smallest test case.

```golang
func example(w safehttp.ResponseWriter, req *safehttp.IncomingRequest) safehttp.Result {
	return safehttp.WriteString(w, "hello world")
}
```

I prevented the code `w.Write("string")` from appearing by specifying a `StringResponse` instead of a `string` type, because I noticed that this is a very different style from the rest of the API.

- [x] Tests pass


kallydev:
@googlebot I signed it!

PR #334, opened 2022-01-17
all: update build tags

bsiegert:
Running gofmt adds these "new-style" build tags now. This is what
has caused CI to be red.

- [x] Tests pass


kele:
Thank you! :)

kele (approved):

Issue #330, opened 2021-10-26
gofmt GitHub action seems broken


kele:
On this PR: https://github.com/google/go-safeweb/runs/4009332391?check_suite_focus=true

I get:

```
Run [ "$(gofmt -l $(find . -name '*.go') 2>&1)" = "" ]
  [ "$(gofmt -l $(find . -name '*.go') 2>&1)" = "" ]
  shell: /bin/bash -e {0}
  env:
    GOROOT: /opt/hostedtoolcache/go/1.17.2/x64
Error: Process completed with exit code 
```

There is little debugging information here, too. I think this should be changed.

kele:
Fixed by #356

PR #348, opened 2022-10-11
Run gofmt from Go 1.19 over all files

bsiegert:
Comments are now Markdown-ish. This should fix the issue in #330.


bsiegert:
It looks like the copyyright linter now disagrees with `gofmt` :(

empijei (approved):

empijei:
Looks like one of the files has a problem with the license header, can you please take a look?

bsiegert:
@empijei This is ready to merge now!

PR #356, opened 2023-04-07
Make "copyright" CI action pass

bsiegert:
This is purely a whitespace change.
    
Part of the copyright header is a line of the form
`//<space><tab>URL`
Putting spaces before tabs is a mistake from a typographical point of view. gofmt seems to agree and occasionally removes the space when    rewriting a file. This makes it impossible to get the "gofmt" and "copyright" linters passing at the same time.

Let's fix this for good by removing the extra space across the codebase.


bsiegert:
I think this will fix #330 for good.

PR #362, opened 2024-07-26
fix: renames onshudown to onShutdown

Jkarage:
- [x] Tests pass

The doc was already referring to this function as onShutdown here https://github.com/google/go-safeweb/blob/master/safehttp/server.go#l=70


bsiegert (approved):
Thank you for the change! Please go through the CLA requirement so that this can be merged.

Jkarage:
Done Already @bsiegert

PR #369, opened 2026-03-29
collector: limit request body size to prevent OOM

johanr-8:
The collector handler reads the entire POST body using ioutil.ReadAll
with no size limit. An attacker can send an arbitrarily large body to
exhaust server memory and crash the server with a single unauthenticated
POST request.

Fix: wrap the body reader with io.LimitReader before reading,
capping input at 1MB which is well above any legitimate CSP or
Reporting API payload size.


bsiegert (approved):

safehttp/plugins/collector/collector.go
bsiegert, line 23: You need to add an import for "io". Ideally, run `goimports`.

PR #370, opened 2026-03-29
collector: fix partial report processing on mixed valid/invalid batch

johanr-8:
The handleReport function calls the user-supplied callback h() for each
valid report in a batch before checking whether any reports in the batch
were malformed. This means an attacker can send a mixed batch of valid
and invalid reports, causing valid reports to be processed and committed
while the server returns HTTP 400 Bad Request, as if the entire request
failed.

Fix: collect all valid reports into a slice first, validate the entire
batch, and only dispatch to the callback after confirming no errors
occurred.


safehttp/plugins/collector/collector.go
```diff
@@ -270,4 +276,4 @@ func cspViolationHandler(m json.RawMessage) (body interface{}, ok bool) {
 		LineNumber:        r.LineNumber,
 		ColumnNumber:      r.ColumnNumber,
 	}, true
-}
+}
```
bsiegert, line 279: nit: newline at end of file please

bsiegert (changes requested):

